Issue #10, opened 2014-10-14
parse function


kossnocorp:

LeshaKoss:
Closed in 947eae4aba78d22025bbd33ccd290c57eb804b73, 91a8c273e7ad4932c111798b0cd213bdc23e1fa2

PR #26, opened 2015-01-27
First implementation of parse() #10

LeshaKoss:


kossnocorp:
Please test against Sugar.js and Moment.js implementations using http://jsperf.com and share the test.

LeshaKoss:
http://jsperf.com/parse-iso-8601

kossnocorp:
A-we-some!

Issue #9, opened 2014-10-14
distanceOfTimeInWords function


kossnocorp:
http://api.rubyonrails.org/classes/ActionView/Helpers/DateHelper.html#method-i-distance_of_time_in_words

Issue #8, opened 2014-10-14
timeAgoInWords function


kossnocorp:
http://api.rubyonrails.org/classes/ActionView/Helpers/DateHelper.html#method-i-time_ago_in_words

PR #27, opened 2015-01-28
distanceOfTimeInWords #9 and timeAgoInWords #8

LeshaKoss:


src/distance_of_time_in_words.js
```diff
@@ -0,0 +1,207 @@
...
+    other: 'about %{count} years'
+  },
+  overXYears: {
+    one: 'over 1 year',
+    other: 'over %{count} years'
+  },
+  almostXYears: {
+    one: 'almost 1 year',
+    other: 'almost %{count} years'
+  }
```
kossnocorp: I think it makes sense to prefer `'almost a year'`-like forms over `'almost 1 year'`. For me it sounds better.

src/distance_of_time_in_words.js
```diff
@@ -0,0 +1,207 @@
...
+
+var isLeapYear = function(year) {
+  return new Date(year, 1, 29).getMonth() == 1;
+};
+
+var locale = {
+  lessThanXSeconds: {
+    one: 'less than a second',
+    other: 'less than %{count} seconds'
+  },
```
kossnocorp: Let's start instead start with more common form: `'a few seconds'` (next will be `'half a minute'`)

PR #30, opened 2015-01-30
getQuarter and setQuarter functions

LeshaKoss:


src/get_quarter.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Returns quarter of year of passed date.
+ * @param {date|string} dirtyDate
+ * @returns {number} (quarter)
```
kossnocorp: @LeshaKoss can you give add an example (`@example`)?

PR #39, opened 2015-02-06
Updated parse() behaviour

LeshaKoss:
`CHANGE BEHAVIOR: returns null when fails, returns milliseconds when plain time`

I. `returns null when fails` — very useful change: 
1)
```javascript
if (var date = parse(maybeDateString)) {
   // ...
}
```
2)
```javascript
var parse = require('date-fns/src/parse');

var arrayOfDates = arrayOfStringsWithDateSomewhere.reduce(function(array, string) {
  if (var date = parse(string)) {
    array.push(date);
  }
  return array;
}, []);
```

II. `returns milliseconds when plain time` — before, parse() returned today date with time set to parsed time. It made no sense, since we couldn't use that time any meaningful way.
```javascript
var date = parse('2014-07-02');
var time = parse('12:00');
var length = parse('02:00');
date.setTime(date.getTime() + time + length);
//=> Wed Jul 02 2014 14:00:00
```

### DIFF
--- a/src/__tests__/parse_test.js
+++ b/src/__tests__/parse_test.js
@@ -1,6 +1,13 @@
 var parse = require('../parse');
 
 describe('parse', function() {
+  describe('failure', function() {
+    it('returns null if there is no date in string', function() {
+      var result = parse('foobarTbaz');
+      expect(result).to.be.null;
+    });
+  });
+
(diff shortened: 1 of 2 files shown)


LeshaKoss:
@kossnocorp check this out please

PR #44, opened 2015-08-08
getQuarter and setQuarter functions, fix behaviour of setMonth

LeshaKoss:
For setMonth:
```javascript
setMonth(new Date(2015, 11 /* Dec */, 31), 1 /* Feb */);
// => 2015-02-28
```
Now it works the same as addMonths and subMonths, setting last day of shorter month if original date was last day of longer month


kossnocorp:
@LeshaKoss please rebase against master

PR #55, opened 2015-08-21
Timezone fixes

dolbyzerr:
- use `parse` instead of `Date` constructor to parse string passed to `format` function
 - `setISOWeek` will set hours and minutes to the starts of the day

### DIFF
--- a/src/format.js
+++ b/src/format.js
@@ -1,5 +1,6 @@
 var startOfDay = require('./start_of_day');
 var startOfYear = require('./start_of_year');
+var parse = require('./parse');
 
 var NUMBER_OF_MS_IN_DAY = 864e5;
 
@@ -10,7 +11,12 @@ var NUMBER_OF_MS_IN_DAY = 864e5;
  * @returns {string}
  */
 var format = function(date, format) {
-  date = date instanceof Date ? date : new Date(date);
+  var type = toString.call(date)
+  if (type == '[object String]') {
+    date = parse(date)
+  } else if(type == '[object Number]') {
+    date = new Date(date)
+  }
 
   if (!format) {
     format = 'YYYY-MM-DDTHH:mm:ss.SSSZ';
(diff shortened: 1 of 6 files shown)


radanskoric:
:checkered_flag:

src/format.js
```diff
@@ -10,7 +11,12 @@ var NUMBER_OF_MS_IN_DAY = 864e5;
  * @returns {string}
  */
 var format = function(date, format) {
-  date = date instanceof Date ? date : new Date(date);
+  var type = toString.call(date)
+  if (type == '[object String]') {
+    date = parse(date)
+  } else if(type == '[object Number]') {
+    date = new Date(date)
+  }
```
kossnocorp: Else?

src/format.js
```diff
@@ -10,7 +11,12 @@ var NUMBER_OF_MS_IN_DAY = 864e5;
  * @returns {string}
  */
 var format = function(date, format) {
-  date = date instanceof Date ? date : new Date(date);
+  var type = toString.call(date)
+  if (type == '[object String]') {
```
kossnocorp: Why not just `typeof`?

dolbyzerr: It's just a habit, it's kinda "safer", 
`typeof(new String('a')) //object`
`toString.call(new String('a')) //[object String]`

I agree that it's not an real life example, and `toString.call` looks uglier, so I will change

src/format.js
dolbyzerr: Else don't do anything with `date`

src/format.js
kossnocorp: I mean `typeof date == 'string'`. Is there is some problem with that?

src/format.js
kossnocorp: Makes sense

src/format.js
dolbyzerr: only problem with `typeof date == 'string'` is that it will be false, when date is created using `new String()` constructor, which is not real life usage.

kossnocorp: Yeah, without that code looks much more simpler.

PR #65, opened 2015-08-31
Add am/pm as long with a.m./p.m. to the format function

dolbyzerr:
In sake of consistency with `moment` format, let's leave `a` format option as "am/pm", and introduce a new format key "aa" which will return "a.m./p.m" with dots.


kossnocorp:
@LeshaKoss please include this in changelog

PR #69, opened 2015-09-09
Convert tests to power-assert

LeshaKoss:
and add es5-shim dependency to them thus making them work in IE8


twada:
Hi, thank you for using power-assert.
Is there any problem using power-assert on IE8?
Feel free to tell me if there is something I can help.

kossnocorp:
@twada actually IE8 was the main reason why we switched to power-assert (chai uses object getters). Thanks, it works great :+1:, I'm in love with it.

twada:
@kossnocorp I'm very glad to hear that! :smile:

PR #100, opened 2015-10-03
isValid function (closes #97)

### LINKED ISSUE #97: `isValid` function

LeshaKoss:


src/is_valid.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Is the given date valid?
```
kossnocorp: @LeshaKoss can you elaborate what is valid data and what is not?

PR #101, opened 2015-10-03
Throw invalid argument exception in eachDay and isWithinRange (closes #92)

### LINKED ISSUE #92: Add a check in `eachDay` function if end date is before start date
Throw exception in:
* eachDay
* isWithinRange

LeshaKoss:


src/is_within_range.js
```diff
@@ -20,12 +21,17 @@ var parse = require('./parse')
...
-  return(
-    time >= parse(dirtyStartDate).getTime()
-    && time <= parse(dirtyEndDate).getTime()
-  )
+  var time = parse(dirtyDate).getTime()
+  var startTime = parse(dirtyStartDate).getTime()
+  var endTime = parse(dirtyEndDate).getTime()
+
+  if (startTime > endTime) {
+    var err = new Error('Argument dirtyStartDate must be before dirtyEndDate')
```
kossnocorp: 1. Error message could be better. It should explain explicitly what "before" means.
2. In JSDoc comment arguments can be named in more friendly way (`date` instead of `dirtyDate`, `startRangeDate` or `leftRangeDate` instead of `dirtyStartDate` etc). Also it makes sense to change it everywhere in the library.

LeshaKoss: 1.
2. I opened an issue for friendlier JSDoc arguments: #104

PR #186, opened 2016-03-09
Fix `power-assert` IE8 bug

LeshaKoss:


kossnocorp:
Awesome! :+1: :rocket:

PR #194, opened 2016-04-27
Fix isValid

LeshaKoss:


kossnocorp:
Nice finding! Please rebase against the master and merge it.

PR #195, opened 2016-04-27
Remove validateDateArguments

LeshaKoss:
Reason: **overengineered**


LeshaKoss:
Note for future generations: to bring the function back, undo only 58f4883, but leave b5880a4

PR #205, opened 2016-05-19
Added flow types for every function

JohnyDays:
Added a .js.flow file for each function describing their inputs/outputs.
Added flow headers to every test file, so i could verify they were correct
All test files pass flow checks correctly!
This doesn't change behaviour so every test and build step passes / remains unchanged.

PS: If you're not familiar with flow, I can give some more context or examples of the typechecking

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,0 +1,2 @@
+[ignore]
+.*node_modules/config-chain/test/broken.json
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ script:
   - npm run test-tz
   - npm run test-tz-extended
   - npm run count-tests
+  - npm run flow-check
 
 # Deployment
 before_deploy:
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "babel-eslint": "^5.0.0",
     "babel-preset-es2015": "^6.6.0",
     "cloc": "^1.0.4",
+    "flow-bin": "^0.25.0",
     "fs-extra": "^0.26.4",
     "fs-promise": "^0.3.1",
     "glob-promise": "^1.0.4",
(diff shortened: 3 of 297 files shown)


kossnocorp:
@JohnyDays great stuff, thank you a lot! I'm familiar with flow but I need some time to play with PR. Also, it makes sense to add flow check to CI to make sure that it will not break in the future.

JohnyDays:
I'm not too familiar with travis, but I think that does the trick

kossnocorp:
@JohnyDays 👍 it does, but now I have problem with SauceLabs.

kossnocorp:
@JohnyDays it turned out that Travis [doesn't expose SauceLabs credentials for security reasons](https://docs.travis-ci.com/user/pull-requests/#Security-Restrictions-when-testing-Pull-Requests). I've [disabled cross-platform tests for external PRs](https://github.com/date-fns/date-fns/pull/206). To make it green, please rebase against master branch.

JohnyDays:
Done 👍

kossnocorp:
@JohnyDays thanks 🙏🏽!

JohnyDays:
Cool 🍒 
Does something have to be done to include these files in the next npm release? release_npm.sh seems to only exlclude test files so I'm guessing these will be there

kossnocorp:
@JohnyDays I will adjust this. 

I have an another problem: I just realied that `// @flow` must be at the first line, otherwise a file going to be ignored:

tmux___users_koss_src_kossnocorp_assets-webpack-plugin

vs

tmux___users_koss_src_kossnocorp_assets-webpack-plugin

I'm not enough experienced with Flow so I can't think of a suitable solution 😞

JohnyDays:
Oh, you're right. I'll submit another PR.

kossnocorp:
@JohnyDays 👍

PR #210, opened 2016-05-22
Fix bug in setDay

LeshaKoss:


kossnocorp:
It's not a bug, it's a feature 😆

PR #212, opened 2016-05-24
Add closestToIndex

dolbyzerr:
Similar to `closestTo` function, `closestToIndex` would return index of the closest date in array.

```
var dateToCompare = new Date(2015, 8, 6)
var datesArray = [new Date(2015, 0, 1), new Date(2016, 0, 1), new Date(2017, 0, 1)]
var result = closestToIndex(dateToCompare, datesArray)
// 1
```

### DIFF
--- a/src/closest_to_index/index.js
+++ b/src/closest_to_index/index.js
@@ -1,0 +1,45 @@
+var parse = require('../parse')
+
+/**
+ * @category Common Helpers
+ * @summary Return an index of closest date from array comparing to given date.
+ *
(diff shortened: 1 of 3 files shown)


LeshaKoss:
Thank you! The function was renamed to `closestIndexTo` and released with [1.3.0](https://github.com/date-fns/date-fns/releases/tag/v1.3.0). (diff: https://github.com/date-fns/date-fns/compare/v1.2.0...v1.3.0)

PR #217, opened 2016-06-07
Proofread the copy of README and Getting Started

sidonath:
I've added a missing coma to sync with https://github.com/date-fns/date-fns.org/pull/86 and also added a few missing articles.


kossnocorp:
❤️

kossnocorp:
Thank you a lot!

PR #225, opened 2016-09-02
Add leading zeroes for small years

just-boris:
Now `format` function for small values return year without leading zeroes.

This is not okay, beacause sometimes it produces wrong output

```js
format(new Date("0001-01-01"), "YYYY-MM-DD") // returns "1-01-01"
```

It causes also issues with `<input type="date">` which accepts only strings with year as 0001

### DIFF
--- a/src/format/index.js
+++ b/src/format/index.js
@@ -164,12 +164,12 @@ var formats = {
 
   // Year: 00, 01, ..., 99
   'YY': function () {
-    return String(this.getFullYear()).substr(2)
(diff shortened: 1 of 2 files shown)


LeshaKoss:
Oops, forgot to say, your fix was released with [1.4.0](https://github.com/date-fns/date-fns/releases/tag/v1.4.0). (diff: https://github.com/date-fns/date-fns/compare/v1.3.0...v1.4.0) Thank you for your contribution!

PR #235, opened 2016-10-15
Chinese Simplified locale

KingMario:


LeshaKoss:
Hello @KingMario! Thanks a lot for your contribution!

I believe that `chs` isn't really a correct language code for Simplified Chinese. [This link](http://stackoverflow.com/a/4894634) says that both `zh_cn` and `zh_hans` are used, later is more generic, but former is much more popular.

What code do you think is more appropriate and/or correct for your locale?

KingMario:
@LeshaKoss
Right. `chs` isn't a correct language code.
I prefer `zh_CN` because it is shorter. :)
So should I rename the folder and commit it again?

kossnocorp:
Yes, please name it `zh_cn` (downcased for consistency). Thank you a lot for you help, this is great piece of work!

KingMario:
My pleasure. ^_^

kossnocorp:
@KingMario thank you again, you rock! `zh_cn` was released as `1.6.0`.

PR #239, opened 2016-10-20
Experimental Moment API

kossnocorp:
Made it for lulz first but then realized that it's possible to achieve.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -2,7 +2,7 @@
   "name": "date-fns",
   "version": "1.6.0",
   "homepage": "https://github.com/date-fns/date-fns",
-  "authors": ["Sasha Koss <koss@nocorp.me>"],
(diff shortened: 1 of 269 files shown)


kossnocorp:
Please left a comment if you're interested.

RIanDeLaCruz:
@kossnocorp Was wondering what this is exactly doing? Creating an API that mirrors the API of moment?

kossnocorp:
@RIanDeLaCruz exactly

RIanDeLaCruz:
@kossnocorp I'm mildly interested. What do you think would be the benefit of doing this?

kossnocorp:
@RIanDeLaCruz I thought that it would encourage people to migrate from Moment and help to identify the subtle differences between similar functions.

Closing it as I doubt that I'll have enough time and focus for that.

PR #247, opened 2016-10-25
Add Travis status badge

p0deje:
Just what it says.


kossnocorp:
The badge is missing for a reason. The reason is that the badge doesn't give much value for the users:

- It doesn't represent the status of the latest package build.
- It's not possible (and not recommended) to use `master` as a dependency.
- We run an extended test suite on a push to `master` which occasionally fails due to flakiness. I won't make users think that red badge means broken package.

We used to have bunch of badges in date-fns README (a dozen, really) and one day I killed them all and happy about it.

Anyway, thank you for an attempt to get a Hacktoberfest shirt :trollface: 😉

p0deje:
> Anyway, thank you for an attempt to get a Hacktoberfest shirt

You got it, dude.

Issue #256, opened 2016-11-23
Write custom locale tests for `format`, `distanceInWords`, `distanceInWordsToNow` and `parse`


LeshaKoss:
I.e. to test if some locale can be passed into these functions, and also to test what if the provided locale does not provide helpers for this specific function

LeshaKoss:
Closed by #321

PR #259, opened 2016-11-25
parse → toDate & new parse function (closes #219)

### LINKED ISSUE #219: Add an ability to pass the format to parse function, so it would be possible to parse dates in any format
Thank you for the nice library!

I was going to use this library, but figured out that it is not possible to parse data by arbitary format. I'd like to do the following

``` js
var dateString = '10.02.1999'
var date = parse(dateString, 'DD.MM.YYYY')
```

[`format`](https://date-fns.org/docs/format) function is already has set of tokens, which can also be used for parsing.

[Moment also has the same feature](http://momentjs.com/docs/#/parsing/string-format/), so it is reasonable to have it there as well.

I can try to implement this and send a PR if you approve general idea of this feature.

LeshaKoss:
- [x] English
- [ ] all other locales


LeshaKoss:
@kossnocorp please review/tell your opinion

kossnocorp:
Go on!

kossnocorp:
How we're going to proceed with languages what are not here? Let's contact original authors and contributors (comments, reviews)?

LeshaKoss:
@kossnocorp we certainly should contact original authors, but before that I want to make helper functions to make it easier to create a locale. For example, `4 утра понедельника 5 января` should be parsed the same way as `4 утро понедельник 5 январь`

LeshaKoss:
@kossnocorp also I wanted to do #256 before that, to make it possible to use a locale when it doesn't support either `distantanceInWords` or `format` or `parse`

kossnocorp:
@LeshaKoss 👌

LeshaKoss:
#375

PR #282, opened 2016-12-06
Added initial flow support

amilajack:
I've only tested this locally. Would be great if someone else could confirm this before merging


kossnocorp:
Works at CI: https://travis-ci.org/date-fns/date-fns/builds/181538267#L231 👍 

I going to test, thanks!

amilajack:
Right! Fixed

kossnocorp:
Here too https://github.com/date-fns/date-fns/blob/master/.travis.yml#L22 ;-)

kossnocorp:
It's covered by `npm run lint`

package.json
```diff
@@ -71,11 +72,16 @@
...
     "systemjs-check": "babel-node scripts/systemjs_check.js"
   },
   "standard": {
     "parser": "babel-eslint",
-    "ignore": ["dist/"]
+    "ignore": [
+      "dist/"
+    ],
+    "rules": {
+      "flowtype-errors/show-errors": 2
```
leshakoss: @amilajack @kossnocorp doesn't work for me 🤔 I don't think `standard` allows custom rules

leshakoss:
Closed because `standard` doesn't allow custom rules

PR #287, opened 2016-12-06
Fixed typescript imports from individual modules. Fixes #264

### LINKED ISSUE #264: Typings doesn't work if you use it as a targeted import
```
import {addDays} from 'date-fns';
```
works while

```
import * as addDays from 'date-fns/add_days';
```

Doesn't. 

There is a very weird behavior on Typescript's part, but having this package actually include the typings actually causes more issues than to have people just do a ```npm install @types/date-fns``` which works brilliantly. Per documentation of TS 2.0+, @types is the official new way for non-TS native project anyway. I think you should remove the definition file and just add a note in README for TS users to use the alternative.

Otherwise, the only way to fix this is to completely rewrite the date-fns typings I think.

mattlewis92:


mattlewis92:
@LeshaKoss I've fixed the merge conflicts with master so this can be merged now

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ This change log follows the format documented in [Keep a CHANGELOG].
 
 ## [Unreleased]
 
+- Fixed typescript imports from individual modules
```
LeshaKoss: @mattlewis92 hi, please resolve merge conflict again 😅 and change `typescript` to `TypeScript`

mattlewis92:
@LeshaKoss done!

LeshaKoss:
@mattlewis92 alright thanks, I'll test it right now

LeshaKoss:
@mattlewis92 I tested it, looks ok, going to merge soon

I have a question. I know nothing about TypeScript.
`import * as addDays from 'date-fns/add_days'` is working now but
`import addDays from 'date-fns/add_days'` is not (`error TS1192: Module ''date-fns/add_days'' has no default export.`) Is that supposed to happen?

mattlewis92:
Yup that's right. Typescript doesn't add a default export like babel does so you need to do `import * as addDays from 'date-fns/add_days';`

LeshaKoss:
Alright, thanks

mattlewis92:
Awesome, thanks! 👍

LeshaKoss:
@mattlewis92 so now your fix was [officially](https://date-fns.org/docs/Change-Log#1.15.1-2016-12-07) released as [v1.15.1](https://github.com/date-fns/date-fns/releases/tag/v1.15.1). (diff: https://github.com/date-fns/date-fns/compare/v1.15.0...v1.15.1) Thanks again, looking forward to your future contributions 😉

PR #296, opened 2016-12-06
Converts to ES6 to enable tree-shaking

### LINKED ISSUE #179: Enable tree shaking for date-fns
- [ ] Introduce modules index file and add it to `package.json` as `main`
- [ ] Add docs

See http://www.2ality.com/2015/12/webpack-tree-shaking.html

alexgorbatchev:
Closes #179

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,3 @@
 {
-  "presets": ["es2015"]
+  "presets": ["es2015", "babel-preset-power-assert"]
 }
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,4 @@
 {
-  "presets": ["es2015", "babel-preset-power-assert"]
+  "presets": ["es2015", "babel-preset-power-assert"],
+  "plugins": ["add-module-exports"]
 }
--- a/package.json
+++ b/package.json
@@ -13,7 +13,9 @@
   "devDependencies": {
     "babel-cli": "^6.6.5",
     "babel-eslint": "^7.1.1",
+    "babel-loader": "^6.2.8",
     "babel-preset-es2015": "^6.6.0",
+    "babel-preset-power-assert": "^1.0.0",
     "flow-bin": "^0.36.0",
     "fs-promise": "^1.0.0",
     "glob-promise": "^2.0.0",
@@ -44,9 +46,9 @@
     "snazzy": "^5.0.0",
     "standard": "^8.6.0",
     "systemjs": "^0.19.39",
+    "systemjs-plugin-babel": "0.0.17",
     "uglify-js": "^2.6.1",
-    "webpack": "^1.12.0",
-    "webpack-espower-loader": "^1.0.1"
+    "webpack": "^1.12.0"
   },
   "scripts": {
     "test": "karma start config/karma.js",
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "babel-cli": "^6.6.5",
     "babel-eslint": "^7.1.1",
     "babel-loader": "^6.2.8",
+    "babel-plugin-add-module-exports": "^0.2.1",
     "babel-preset-es2015": "^6.6.0",
     "babel-preset-power-assert": "^1.0.0",
     "flow-bin": "^0.36.0",
@@ -81,6 +82,8 @@
   },
   "standard": {
     "parser": "babel-eslint",
-    "ignore": ["dist/"]
+    "ignore": [
+      "dist/"
+    ]
   }
 }
--- a/scripts/release_npm.sh
+++ b/scripts/release_npm.sh
@@ -37,6 +37,16 @@ done
 #
 cp -r ./src $dir/es6
 
+#
+# Copy original ES6 index.js to the ./es6 folder
+#
+cp ./index.js $dir/es6/
+
+#
+# Compile index.js to the ES5 ./ folder
+#
+$(npm bin)/babel --out-file $dir/index.js ./index.js
+
 cp dist/date_fns_docs.json $dir/docs.json
 find "$dir" -type f -name "test.js" -delete
 find "$dir" -type f -name "benchmark.js" -delete
--- a/scripts/release_npm.sh
+++ b/scripts/release_npm.sh
@@ -24,6 +24,19 @@ do
   cp scripts/sub_module_package.json "$module/package.json"
 done
 
+#
+# This will compile all ES6 files on top of the already copied files leaving
+# all non *.js files in place as expected
+#
+(cd src && $(npm bin)/babel --out-dir ../$dir **/*.js)
+
+#
+# Copy ES6 files into the ./src folder so that one could call
+#
+#   import foo from 'date-fns/es6/foo';
+#
+cp -r ./src $dir/es6
+
 cp dist/date_fns_docs.json $dir/docs.json
 find "$dir" -type f -name "test.js" -delete
 find "$dir" -type f -name "benchmark.js" -delete
--- a/scripts/systemjs_check.js
+++ b/scripts/systemjs_check.js
@@ -3,6 +3,16 @@ import listLocales from './_lib/list_locales'
 import SystemJS from 'systemjs'
 import path from 'path'
 
+const pluginPath = path.resolve(__dirname, '../node_modules/systemjs-plugin-babel')
+
+SystemJS.config({
+  map: {
+    'plugin-babel': path.join(pluginPath, 'plugin-babel.js'),
+    'systemjs-babel-build': path.join(pluginPath, 'systemjs-babel-browser.js')
+  },
+  transpiler: 'plugin-babel'
+})
+
 Promise.all(importFns().concat(importLocales())).then(
   () => {
     console.log('SystemJS support is OK')
--- a/src/add_days/index.js
+++ b/src/add_days/index.js
@@ -1,4 +1,4 @@
-var parse = require('../parse/index.js')
+import parse from '../parse/index.js'
 
 /**
  * @category Day Helpers
@@ -22,4 +22,4 @@ function addDays (dirtyDate, amount) {
   return date
 }
 
-module.exports = addDays
+export default addDays
--- a/src/add_days/test.js
+++ b/src/add_days/test.js
@@ -1,8 +1,8 @@
 // @flow
 /* eslint-env mocha */
 
-var assert = require('power-assert')
-var addDays = require('./')
+import assert from 'power-assert'
+import addDays from '.'
 
 describe('addDays', function () {
   it('adds the given number of days', function () {
--- a/config/webpack.js
+++ b/config/webpack.js
@@ -7,7 +7,7 @@ var config = {
   output: getOutputConfig(),
   module: {
     loaders: [
-      {test: /\.js$/, loader: 'webpack-espower', exclude: /node_modules/},
+      {test: /\.js$/, exclude: /node_modules/, loaders: ['babel']},
       {test: /\.json$/, loader: 'json'}
     ]
   }
--- a/index.js
+++ b/index.js
@@ -1,155 +1,153 @@
-module.exports = {
-  addDays: require('./add_days/index.js'),
-  addHours: require('./add_hours/index.js'),
-  addISOYears: require('./add_iso_years/index.js'),
-  addMilliseconds: require('./add_milliseconds/index.js'),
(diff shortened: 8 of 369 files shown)


alexgorbatchev:
hm... so there are actually no failures in the test run but it seems to be exiting with 1...

mattlewis92:
I think this will break things for folks using tools like webpack1 as `import fn from 'date-fns/fn';` will now have an ES6 module syntax in it which non ES6 aware bundlers can't understand.

mattlewis92:
For this to work the build script will need to transpile the separate ES6 functions into commonjs to maintain BC. It will then need to copy the ES6 source into something like a `esm` directory and add a `module` field to the `package.json` that points to `esm/index.js`

alexgorbatchev:
@mattlewis92 thanks for feedback! I updated `release_npm.sh` to transpile es6 to es3 to maintain backward compatibility... at the same time, it copies `src` into `es6` folder to enable `import foo from 'date-fns/es6/foo'`

What do you think?

```
#
# This will compile all ES6 files on top of the already copied files leaving
# all non *.js files in place as expected
#
(cd src && $(npm bin)/babel --out-dir ../$dir **/*.js)

#
# Copy ES6 files into the ./src folder so that one could call
#
#   import foo from 'date-fns/es6/foo';
#
cp -r ./src $dir/es6
```

mattlewis92:
Nice job! There's just a couple more small changes to make
- [ ] remove the dist changes from the PR as these are auto generated on release - and check the script that builds them transpiles and bundles the files together into ES5 ok
- [ ] convert `index.js` to ES6, and transpile it into ES5 as well
- [ ] update the package.json's `main` property to point to the transpiled ES5 `index.js` entry point
- [ ] update the package.json's `module` property to point to the ES6 `index.js` entry point

Currently just converting the individual functions to ES6 modules won't give much benefit from tree shaking as it's effectively what you're already doing with `import foo from 'date-fns/foo'`. Tree shaking will only be effective when doing `import {foo} from 'date-fns';`

Hope that helps! 😃

havenchyk:
@mattlewis92 I'd also add `jsnext:main` field  https://github.com/lodash/lodash/blob/es/package.json#L10 and keep index.js in ES6 with imports

kossnocorp:
Thank you for your work, it's a great job!

I'm concerned about the latest ES usage because it will require precompilation step not only for the dist but for the npm package as well. Please give me time to think about all cons and pros. Initially, I planned to enable tree-shaking by providing an alternative `index.js` file. I want to know if it's possible to do, then I will prefer the latter approach (as it simpler). If it's not then the latest ES is the only solution.

alexgorbatchev:
@kossnocorp should i continue working on this PR?

kossnocorp:
@alexgorbatchev please 🙏  given that webpack 2 is going to support `import` statement, I think we're good to go!

.babelrc
```diff
@@ -1,3 +1,3 @@
 {
-  "presets": ["es2015"]
+  "presets": ["es2015", "babel-preset-power-assert"]
```
kossnocorp: I think this should be scoped to the env because it might affect the production env.

scripts/release_npm.sh
```diff
@@ -16,6 +16,19 @@ do
...
+#
+(cd src && $(npm bin)/babel --out-dir ../$dir **/*.js)
+
+#
+# Copy ES6 files into the ./src folder so that one could call
+#
+#   import foo from 'date-fns/es6/foo';
+#
+cp -r ./src $dir/es6
+
```
kossnocorp: 👍

scripts/systemjs_check.js
```diff
@@ -3,6 +3,16 @@ import listLocales from './_lib/list_locales'
 import SystemJS from 'systemjs'
 import path from 'path'
 
+const pluginPath = `${__dirname}/../node_modules/systemjs-plugin-babel`
```
kossnocorp: Please use `path.resolve`: `path.resolve(__dirname, '../node_modules/systemjs-plugin-babel')` so we won't have glitches at Windows.

scripts/systemjs_check.js
```diff
@@ -3,6 +3,16 @@ import listLocales from './_lib/list_locales'
...
+
+SystemJS.config({
+  map: {
+    'plugin-babel': `${pluginPath}/plugin-babel.js`,
+    'systemjs-babel-build': `${pluginPath}/systemjs-babel-browser.js`
+  },
+  transpiler: 'plugin-babel'
+});
```
kossnocorp: Can tell a bit about how it works? I'm not too familiar SystemJS.

And btw 👮 sorry, but no semicolons here :trollface: we're following [JavaScript "Standard" Style](http://standardjs.com/)

src/add_days/index.js
```diff
@@ -22,4 +22,4 @@ function addDays (dirtyDate, amount) {
   return date
 }
 
-module.exports = addDays
+export default addDays
```
kossnocorp: I think it would be cool to have just `export default function addDays (dirtyDate, amount) {` (here and everywhere), so a code reader could quickly spot module exports.

Now you need to scroll to the bottom to see it.

package.json
```diff
@@ -13,7 +13,9 @@
   "devDependencies": {
     "babel-cli": "^6.6.5",
     "babel-eslint": "^7.1.1",
+    "babel-loader": "^6.2.8",
     "babel-preset-es2015": "^6.6.0",
+    "babel-preset-power-assert": "^1.0.0",
```
kossnocorp: What's the difference between this package and `babel-plugin-espower` (I use it in my another project)?

src/add_days/test.js
```diff
@@ -1,8 +1,8 @@
 // @flow
 /* eslint-env mocha */
 
-var assert = require('power-assert')
-var addDays = require('./')
+import assert from 'power-assert'
+import addDays from './'
```
havenchyk: isn't it enough `import addDays from '.'`?

kossnocorp: @havenchyk this is legacy from previous Node.js versions. Now it's not actual so could be replaced with `.` 👍

package.json
alexgorbatchev: https://github.com/power-assert-js/babel-preset-power-assert has the details :)

src/add_days/index.js
alexgorbatchev: I agree... couldn't figure out quickly enough how to do this search and replace... don't think i'm up for the task of doing this manually :) any tips?

mattlewis92:
I think this might be a breaking change for typescript users as there will now be a default export added to each function i.e you will need to do
Import foo from 'date-fns/foo';

Instead of this previously:
Import * as foo from 'date-fns/foo';

kossnocorp:
@mattlewis92 ouch. The it's a good candidate for `2.0`.

kossnocorp:
`import foo from 'date-fns/foo'` looks like that we need 👍

mattlewis92:
Yup the new behaviour in this PR is definitely the correct one. 

It will also break common js usage as you'll need to do
require('date-fns/foo').default however that can be rectified by using this babel plugin:
https://www.npmjs.com/package/babel-plugin-add-module-exports

kossnocorp:
`babel-plugin-add-module-exports` sounds like that we need.

With 2.0 we're going to rename `date-fns/add_days` to `date-fns/addDays`, so please use this scheme where it's possible.

Btw I like `date-fns/es/addDays` path.

In the future, we'll have `fp` and `utc` submodules, it will be like:

```
date-fns
date-fns/utc
date-fns/fp
date-fns/fp/utc
date-fns/es
date-fns/es/utc
date-fns/es/fp
date-fns/es/fp/utc
```

But it's not a big deal when you can just `import {addDays} from 'date-fns/fp/utc'`.

src/add_days/test.js
alexgorbatchev: fixed

scripts/systemjs_check.js
alexgorbatchev: fixed

scripts/systemjs_check.js
alexgorbatchev: fixed

alexgorbatchev:
I'm not clear if we want `import { foo } from ...` or `import foo from ...`?

kossnocorp:
@alexgorbatchev `import addDays from 'date-fns/es/addDays` but `import {addDays} from 'date-fns/es`

kossnocorp:
@mattlewis92 please correct me if I wrong. `babel-plugin-add-module-exports` will allow us to keep support of `const addDays = require('date-fns/addDays')`? If so then it's not breaking change.

alexgorbatchev:
@kossnocorp i believe `const addDays = require('date-fns/addDays')` works as is right now, if you look at the generated code, it has `exports.default = addDays`... eg

```js
'use strict';

Object.defineProperty(exports, "__esModule", {
  value: true
});

var _index = require('../parse/index.js');

var _index2 = _interopRequireDefault(_index);

function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }

/**
 * @category Day Helpers
 * @summary Add the specified number of days to the given date.
 *
 * @description
 * Add the specified number of days to the given date.
 *
 * @param {Date|String|Number} date - the date to be changed
 * @param {Number} amount - the amount of days to be added
 * @returns {Date} the new date with the days added
 *
 * @example
 * // Add 10 days to 1 September 2014:
 * var result = addDays(new Date(2014, 8, 1), 10)
 * //=> Thu Sep 11 2014 00:00:00
 */
function addDays(dirtyDate, amount) {
  var date = (0, _index2.default)(dirtyDate);
  date.setDate(date.getDate() + amount);
  return date;
}

exports.default = addDays;
```

package.json
kossnocorp: Cool!

mattlewis92:
@kossnocorp there are actually 2 breaking changes this would introduce. To give some context, previously each function would be 
```
module.exports = foo;
```

and now after transpilation
```
module.exports.default = foo;
```

This will break commonjs in this way
```
var foo = require('date-fns/foo'); // currently
var foo = require('date-fns/foo').default; // after merging this PR
```

This BC *can* be solved with this plugin: https://www.npmjs.com/package/babel-plugin-add-module-exports

However the other breaking change is this (but only for typescript users. babel, rollup etc infer `module.exports = foo;` as the default export, typescript doesn't and requires you to explicitly do `module.exports.default = foo;`)
```
import * as foo from 'date-fns/foo';
console.log(foo); // currency logs foo 
console.log(foo); // after merging this PR logs {default: foo} // 
```

AFAIK there is no babel plugin to workaround this issue. I don't think it would be too difficult though to write a plugin though that transpiles
```
export default foo;
```

into 
```
module.exports = foo;
```

This would allow backwards compatibility to be kept.

src/add_days/index.js
mattlewis92: You could write a codemod to do it: https://github.com/facebook/jscodeshift

kossnocorp:
> AFAIK there is no babel plugin to workaround this issue. I don't think it would be too difficult though to write a plugin though that transpiles

I hope we'll find a time to do it, because it sounds like a decent plan to me.

mattlewis92:
I'll take a go at it this weekend 😄

alexgorbatchev:
I've been thinking more about the breaking changes. Are there reasons not to release es6 version only going forward (transpiled)? Cutting a new major release won't affect anyone currently using the library and upgrade path would be relatively straight forward. I feel everything should be moving to es6 and more we push, then we will get there faster :)

Thoughts?

mattlewis92:
@alexgorbatchev I think you're right to only include this as a major release - writing a babel plugin to remove the default export from transpiled modules to try and preserve BC wouldn't work due to how the index.js requires the default export from each module (which would be removed by the babel plugin). It might be better to wait though to gather any required breaking changes together into one major bump.

p.s. can you add the `main` (to transpiled es5 index), `module` (to the es6 index) and `js:next` (to the es6 index) entry points in the package.json so ES6 aware module bundlers will be able to pick up the ES6 versions?

kossnocorp:
@alexgorbatchev thank you a lot for starting this discussion and your contribution to date-fns! I going to re-open this PR to https://github.com/date-fns/date-fns/tree/v2 and test it against different use cases. If needed we'll update and finalize this PR (we'll keep authorship).

kossnocorp:
See: https://github.com/date-fns/date-fns/pull/379

PR #297, opened 2016-12-06
Add italian locale

albertorestifo:
I've put my nationality to good use and created an Italian locale.

I based it off the Spanish locale files, given how similar the two languages are.

### DIFF
--- a/src/locale/it/build_format_locale/index.js
+++ b/src/locale/it/build_format_locale/index.js
@@ -1,0 +1,84 @@
+var buildFormattingTokensRegExp = require('../../_lib/build_formatting_tokens_reg_exp/index.js')
+
+function buildFormatLocale () {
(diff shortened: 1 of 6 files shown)


albertorestifo:
I made a second mistake. I'll rebase and then open a new PR

src/locale/it/build_format_locale/index.js
```diff
@@ -0,0 +1,84 @@
...
+    switch (rem100 % 10) {
+      case 1:
+        return number + 'º'
+      case 2:
+        return number + 'º'
+      case 3:
+        return number + 'º'
+    }
+  }
+  return number + 'º'
```
guigrpa: If I'm not missing something, you can remove all the previous lines of this function and it will provide exactly the same results

guigrpa: Clarification: you can just leave `return number + 'º'`

PR #312, opened 2016-12-08
Fix locale specs

ertrzyiks:
Specs for ordinal results call `assert` with API of `assert.equal`. It has started with `en` locale and got copy-pasted to others. I've fixed it for `en` and `de`. It should be ok in #294 already.

PR #320, opened 2016-12-09
Fix incorrect locale tests (closes #317)

### LINKED ISSUE #317: Check other locales for incorrect tests
https://github.com/date-fns/date-fns/pull/312

LeshaKoss:
oh boy


rbudiharso:
It looks alright, did something broke with it? I'll take a look at it
sometime today if there's anything broken with the Indonesian locale.

LeshaKoss:
@rbudiharso we had not tested ordinal numbers in all `format` locales properly (see: #312), so for example in Indonesian locale the ordinal of 101 was `pertama`. It doesn't really affect formatting results because `format` right now never outputs ordinal numbers bigger than 31

PR #322, opened 2016-12-09
Change `Object.keys` to loop

LeshaKoss:
`Object.keys` shouldn't work in IE8


c-dante:
An alternative to changing the implementation is to include the [polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys#Polyfill).

kossnocorp:
@c-dante I think it would be overkill to use polyfill in this case.

PR #325, opened 2016-12-10
add missing `isSoLastWeek`

claus:
see https://twitter.com/kossnocorp/status/807512029158395904

### DIFF
--- a/src/is_so_last_week/index.js
+++ b/src/is_so_last_week/index.js
@@ -1,0 +1,27 @@
+var isSameWeek = require('../is_same_week/index.js')
+var addWeeks = require('../add_weeks/index.js')
+
+/**
+ * @category Week Helpers
+ * @summary Is the given date so last week?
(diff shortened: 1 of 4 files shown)


src/is_so_last_week/index.js
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @example
+ * // Is 7 December 2016 so last week?
+ * var result = isSoLastWeek(
+ *   new Date(2016, 12, 7)
+ * )
+ * //=> true
+ */
+function isSoLastWeek (dirtyDate, options) {
+  return isSameWeek(new Date(), addWeeks(dirtyDate, 1), options)
```
kossnocorp: It was easier to just `return false` here 😝

kossnocorp:
@LeshaKoss let's release this as an easter egg. Just filter it out in docs and the `dist`.

PR #330, opened 2016-12-12
Fix SS and SSS format with leading zeros

pauldijou:
According to docs, `SSS` format should always be a 3 characters string, with leading zeros if necessary. Same for `SS`. That wasn't the case. Added tests and fixed the problem.


LeshaKoss:
@pauldijou oh oops omg, thank you!

LeshaKoss:
@pauldijou thank you for your fix! It was released with [1.18.0](https://github.com/date-fns/date-fns/releases/tag/v1.17.0) (diff: https://github.com/date-fns/date-fns/compare/v1.17.0...v1.18.0) 🌟

pauldijou:
That was like... super super fast. Thanks a lot.

PR #331, opened 2016-12-12
Added date range overlap check methods

### LINKED ISSUE #319: Method suggestion: Return the number of days that overlap in two date ranges
I had to write a method for returning the number of overlapping days in date ranges several times in different projects already and always found it a missing feature from the existing libraries.

Would you consider adding it to date-fns?

I can submit a PR with it if you're interested.

asia-t:
Added date range methods to determine whether and by how many days two date ranges overlap.

Resolves #319

### DIFF
--- a/src/are_ranges_overlapping/index.js
+++ b/src/are_ranges_overlapping/index.js
@@ -1,0 +1,44 @@
+var parse = require('../parse/index.js')
+
+/**
+ * @category Range Helpers
+ * @summary Is the given date range overlapping with another date range?
+ *
+ * @description
+ * Is the given date range overlapping with another date range?
+ *
+ * @param {Date|String|Number} dirtyInitialRangeStartDate - the start of the initial range
+ * @param {Date|String|Number} dirtyInitialRangeEndDate - the end of the initial range
+ * @param {Date|String|Number} dirtyComparedRangeStartDate - the start of the range to compare it with
+ * @param {Date|String|Number} dirtyComparedRangeEndDate - the end of the range to compare it with
+ * @returns {Boolean} whether the date ranges are overlapping
+ * @throws {Error} startDate of a date range cannot be after its endDate
+ *
+ * @example
+ * // For overlapping date ranges:
+ * isOverlappingRange(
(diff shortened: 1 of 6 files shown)


LeshaKoss:
@asia-t Hello! Thank you for your contribution!

I feel like names of these function aren't really consistent with the other functions. `isOverlappingRange` should probably be more naturally named as `isRangeOverlapping`. Also I am not sure if `num` should be shortened like this in `getNumOverlappingDays`. Maybe `getNumberOfOverlappingDays`? Or `getOverlappingDays` or `getOverlappingDaysInRange`? (we didn't use `number` at all in `getDaysInYear`, `getDaysInMonth` and `getISOWeeksInYear`) Your thoughts?

asia-t:
@LeshaKoss Hi! Yeah, I was struggling with the naming a bit. I like your suggestions - `isRangeOverlapping` and `getOverlappingDaysInRange` sound more natural. I'll make the changed today. Thanks!

LeshaKoss:
@asia-t also, I am not sure about it, but maybe `areRangesOverlapping` is better than `isRangeOverlapping`? There are two ranges after all

asia-t:
Hey @LeshaKoss ! I went for plural in both for consistency's sake. So, it's now `areRangesOverlapping` and `getOverlappingDaysInRanges`. Does it sound ok?

LeshaKoss:
@asia-t yeah, it's good! Thank you!

LeshaKoss:
@asia-t thank you for your contribution! ⭐️ Your functions were released as [1.20.0](https://github.com/date-fns/date-fns/releases/tag/v1.20.0) (diff: https://github.com/date-fns/date-fns/compare/v1.19.0...v1.20.0)

asia-t:
Hey @LeshaKoss ! Something is not quite right with the function `getOverlappingDaysInRanges` in the documentation on the site. It looks like I messed up the format by putting a const between the doc part and the function name :/ Sorry about that.

If that is the case, I submitted a PR that fixes it:
https://github.com/date-fns/date-fns/pull/337

PR #334, opened 2016-12-13
Added Greek locale files (el)

teoulas:
Support for Greek.

Greek has genders and some formats ("D MMMM" and "DD MMMM") should output the month in genitive.

### DIFF
--- a/src/locale/el/build_format_locale/index.js
+++ b/src/locale/el/build_format_locale/index.js
@@ -1,0 +1,87 @@
+var buildFormattingTokensRegExp = require('../../_lib/build_formatting_tokens_reg_exp/index.js')
+
+function buildFormatLocale () {
+  var months3char = ['Ιαν', 'Φεβ', 'Μαρ', 'Απρ', 'Μαϊ', 'Ιουν', 'Ιουλ', 'Αυγ', 'Σεπ', 'Οκτ', 'Νοε', 'Δεκ']
+  var monthsFull = ['Ιανουάριος', 'Φεβρουάριος', 'Μάρτιος', 'Απρίλιος', 'Μάιος', 'Ιούνιος', 'Ιούλιος', 'Αύγουστος', 'Σεπτέμβριος', 'Οκτώβριος', 'Νοέμβριος', 'Δεκέμβριος']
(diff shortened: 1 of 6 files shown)


LeshaKoss:
@teoulas thanks a lot! 🌟

> "D MMMM" and "DD MMMM"

What about "Do MMMM"?

src/locale/el/build_format_locale/index.js
```diff
@@ -0,0 +1,85 @@
...
+  // Generate ordinal version of formatters: M -> Mo, D -> Do, etc.
+  var ordinalGenders = {
+    'M': 'ος',
+    'D': 'η',
+    'DDD': 'η',
+    'd': 'η',
+    'Q': 'ο',
+    'W': 'η'
+  }
+  Object.keys(ordinalGenders).map(function (formatterToken) {
```
LeshaKoss: @teoulas Please don't use `Object.keys` (it doesn't work in IE8, sadly). Separate array which is a copy of `Object.keys(ordinalGenders)` is ok

teoulas: IE8? Wow, I completely forgot that! I'll change it.

teoulas:
Good point about "Do MMMM". It's not a common format, but momentjs supports it correctly. I will take a look.

LeshaKoss:
@teoulas 

> Good point about "Do MMMM". It's not a common format, but momentjs supports it correctly. I will take a look.

In Russian locale we made it like this:

https://github.com/date-fns/date-fns/blob/master/src/locale/ru/build_format_locale/index.js#L75

teoulas:
@LeshaKoss thanks for the tip, that's exactly how I implemented it as well!

LeshaKoss:
@teoulas great work! Please fix the linting problem and we're good to go https://travis-ci.org/date-fns/date-fns/builds/183575458#L230

teoulas:
Fixed linting and squashed commits. Good to go. 🚀

LeshaKoss:
@teoulas thank you for your contribution! 🇬🇷 Greek locale was released with [1.19.0](https://github.com/date-fns/date-fns/releases/tag/v1.19.0) (diff: https://github.com/date-fns/date-fns/compare/v1.18.0...v1.19.0)

teoulas:
👍  Thanks for the super quick response and new release!

Issue #324, opened 2016-12-10
Filipino (Philippines) language support


RIanDeLaCruz:

RIanDeLaCruz:
Hi all. Claiming this one. :)

LeshaKoss:
@RIanDeLaCruz thank you!

LeshaKoss:
closed in #339

PR #339, opened 2016-12-15
Locale filipino

RIanDeLaCruz:
For filipino locale as in #324


LeshaKoss:
@RIanDeLaCruz hello, thanks a lot for your contribution! We [generally](https://github.com/date-fns/date-fns/tree/master/src/locale) use two-letter codes for locales, except the cases where the second part of a four-letter code refers to a variant of macrolanguage (like in `zh_cn` and `zh_tw` which are respectively Simplified and Traditional Chinese locales). So, we need either 2 or 4 letter code. Are Tagalog and Filipino different things? I see that moment [uses](https://github.com/moment/moment/blob/develop/src/locale/tl-ph.js) `tl_ph` code for Filipino, so I guess the latter is a variant of the former. Are there any other used variants of Tagalog? If there are none, I would prefer the locale to be coded as just `tl` ; or `tl_ph` otherwise or if it is for any reason wrong to make it just `tl`. Your thoughts?

LeshaKoss:
Well, maybe I am wrong and three letter code is ok here. This resource http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry lists `fil` as the correct code for Filipino

RIanDeLaCruz:
Filipino is the official language while Tagalog is just a dialect from which Filipino was based on. I'm leaning towards using fil because it's the official term. Would you think that's fine, @LeshaKoss?

LeshaKoss:
@RIanDeLaCruz I tried to understand language code standards but failed miserably. Let's leave it as `fil` 😅

LeshaKoss:
@RIanDeLaCruz thanks again for the contribution! ⭐️ Is it ready to be shipped?

RIanDeLaCruz:
Yep! :)

LeshaKoss:
@RIanDeLaCruz thanks yet again! 🇵🇭 Filipino locale was released with [v1.21.0](https://github.com/date-fns/date-fns/releases/tag/v1.21.0) (diff: https://github.com/date-fns/date-fns/compare/v1.20.1...v1.21.0)

PR #345, opened 2016-12-19
Expose I18n Translations Outside Of A Date Context

jameshopkinsASOS:
I've found the [I18n support](https://date-fns.org/docs/I18n) of `date-fns` to be very useful.

However, at the moment, the translations can only be used in the context of a fully formed timestamp. This means that I cannot utilise these translation dictionaries to render out a textual list of months, for example, without being very hacky, and providing a superfluous timestamp to each.

In order to solve this, I've created a new method `.getTranslation` on each `locale` object, that, given a formatter token (as per [`format`](https://date-fns.org/docs/format)), and an optional index, returns a single translated phrase at that index. If the optional index isn't provided, the whole list is returned.

For example:

```js
const daLocale = require('date-fns/locale/da');

daLocale.getTranslation('MMM')
$> ['jan', 'feb', 'mar', 'apr', 'maj', 'jun', 'jul', 'aug', 'sep', 'okt', 'nov', 'dec']

daLocale.getTranslation('dddd', 3)
$> onsdag
```

There is now a translations dictionary for each locale (located in `./locale/*/translations/index.js`), from which this new method obtains its values. This dictionary is also used for each method that requires it in each `build_format_locale`.

Hope it proves of some use!


src/locale/el/build_format_locale/index.js
```diff
@@ -74,7 +65,7 @@ function buildFormatLocale () {
   formatsWithGenitive.forEach(function (formatterToken) {
     formatters[formatterToken + ' MMMM'] = function (date, commonFormatters) {
       var formatter = formatters[formatterToken] || commonFormatters[formatterToken]
-      return formatter(date, commonFormatters) + ' ' + monthsGenitive[date.getMonth()]
+      return formatter(date, commonFormatters) + ' ' + getTranslation('MMMM-genitive', date.getMonth())
```
LeshaKoss: @jameshopkinsASOS hello, thanks a lot for your contribution! That is indeed very useful! But as I see, the intention of it is to provide an universal interface for translations in any locale. But there:
`getTranslation('MMMM-genitive', date.getMonth())`
is an unique token for the Greek locale. Maybe it makes sense to make the '-genitive' variant of token an option to the function that would be ignored in the other locales? Like
`getTranslation('MMMM', date.getMonth(), {case: 'genitive')` or something like that. Not sure about the API though. Your thoughts?

jameshopkinsASOS: HI @LeshaKoss, thanks for the prompt feedback.

> ... Maybe it makes sense to make the '-genitive' variant of token an option to the function that would be ignored in the other locales? 

I went with the current approach, since the genitive derivative seemed to be used exclusively with the `MMMM` (full month textual representation) formatter token, therefore I assumed (in the context of what is already implemented in `date-fns`), that it's a subset of this textual representation.

Therefore, I'd be hesitant about passing it in as an additional option for the aforementioned reason, in that it wouldn't make sense(?) in the case of any other format other than `MMMM`.

I'm no expert on date `I18ln`, so please feel free to correct me where needed :)

LeshaKoss: @jameshopkins for example, in the Russian locale:
`A: ['ночи', 'утра', 'дня', 'вечера']`
these are too in the genitive case. We just haven't find the use for nominative (default) case here (yet)

jameshopkins: If this is the case @LeshaKoss, then why isn't the [variable in the current implementation](https://github.com/date-fns/date-fns/blob/master/src/locale/ru/build_format_locale/index.js#L11) namespaced with `Genitive` like the ['full month' equivalent](https://github.com/date-fns/date-fns/blob/master/src/locale/ru/build_format_locale/index.js#L7) seems to be?

Pending an outcome, I'm happy to amend where necessary.

jameshopkins:
I've provided [a better interface](https://github.com/date-fns/date-fns/pull/345/commits/6bcb68647c34485e749543cafb9d782f77c1cc00) for non-normative translations.

This enforces the convention that keys of normative translations (i.e `MMMM`) purely described their format, whereas their non-normative counterparts (i.e `MMMM-genitive`), describe their format, *as well* as their casing.

jameshopkins:
Hi @LeshaKoss, has there been any update on a resolution for my comments?

jameshopkins:
Happy to fix the conflict, if we can get a resolution on the comment I left?

Cheers

James

leshakoss:
@jameshopkinsASOS @jameshopkins alright. Sorry for long lasting silence. I didn't know (and still don't) what to do with your PR. One the one hand, having a list of names in languages certainly would be useful. On the other hand, it isn't a task for `date-fns`, since it has something to do with dates only tangentially (there are language data libraries like this one: http://www.unicode.org/cldr/charts/29/summary/en.html that do have this goal)

I want `date-fns` locale interface to be as flexible as possible. Languages are hard, and there may be some languages, in which having a predetermined list of names for date units may be impossible due to some unique grammar constructions, like grammar cases or something else.

Anyway, I still want the interface like you suggested and implemented, but maybe not in the shape you implemented. Please let us (`date-fns` team) think some more on this matter. Please don't close your PR. I don't promise it will be merged before `v2.0.0`, but we are working hard on `v2.0.0` right now: https://github.com/date-fns/date-fns/tree/v2 so maybe it makes sense to include your feature as the part of this major release, or at least some time after it.

Again, sorry for the delay.

@kossnocorp could you please share your thoughts on this issue?

leshakoss:
@jameshopkinsASOS @jameshopkins hello again! If you don't mind, I want to hijack your pull request and send it to `v2-locales` branch as #468. I will continue your work on it from this point

PR #346, opened 2016-12-19
Code generation experiments

LeshaKoss:


scripts/_lib/list_locales.js
```diff
@@ -6,7 +6,21 @@ export default function listLocales () {
...
     .filter((locale) => /^[^._]/.test(locale))
     .map((locale) => ({
-      name: locale,
+      name: camelize(locale),
+      path: `./locale/${locale}`,
       fullPath: `./src/locale/${locale}/index.js`
     }))
 }
+
+function camelize (str) {
```
havenchyk: would you mind to use `lodash` for such cases and not to reinvent the wheel?

LeshaKoss: @havenchyk `_.camelCase` would not suit here (I wanted to transform `zh_cn` into `zhCN` not `zhCn`)

havenchyk: @LeshaKoss I see, `upperFirst(camelCase(str))` will solve your problem, but in this case maybe your solution is better

LeshaKoss:
#368

LeshaKoss:
(there were more changes in this PR but #355 made them needless)

PR #355, opened 2016-12-23
Auto generate typescript definitions from documentation on release

mattlewis92:
Sorry I took a bit longer than anticipated to get to this, LMK if you'd like me to change anything 😃


scripts/release_npm.sh
```diff
@@ -30,4 +32,4 @@ find "$dir" -type f -name "benchmark.js" -delete
 cd "$dir" || exit
 npm publish
 cd - || exit
-rm -rf "$dir"
+rm -rf "$dir" typings.d.ts
```
LeshaKoss: I don't think we should delete it, let it stay in the repo like index.js and dist

mattlewis92: My thinking was that it can cause confusion - people might spend their time making PRs to the typings files when it would get overwritten on the next release.

LeshaKoss: Well, `index.js` is generated too. I don't think it matters

mattlewis92: Fair enough, I'll keep it in then

scripts/generate_typings.js
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  return typeString
+}
+
+function getModuleDefinition (name, moduleSuffix = '') {
+  return `
+declare module 'date-fns/${camelCaseToSnakeCase(name)}${moduleSuffix}' {
+  import {${name}} from 'date-fns';
+  export = ${name};
```
LeshaKoss: Maybe we should `Standard` style http://standardjs.com for TypeScript files too to keep it consistent with the library? Like no `;`s and such. Unless TS has it's own standard style, of course

scripts/generate_typings.js
```diff
@@ -0,0 +1,103 @@
...
+    const caseCorrectedType = correctTypeCase(type)
+    if (forceArray) {
+      return `${caseCorrectedType}[]`
+    }
+
+    return caseCorrectedType
+  }).join(' | ')
+
+  if (typeString === 'Date | string | number') {
+    return 'DateOrStringOrNumber'
```
LeshaKoss: Why not just 'Date | string | number'? Would be easier to copy paste the code to separate files, if needed

mattlewis92: I left that in so it was easier to diff with the existing typings.d.ts file, I'll remove it now

mattlewis92:
@LeshaKoss I've implemented all the changes you requested, let me know if you'd like me to change anything else 😃

LeshaKoss:
@mattlewis92 your changes were released with [v1.23.0](https://github.com/date-fns/date-fns/releases/tag/v1.23.0) (diff: https://github.com/date-fns/date-fns/compare/v1.22.0...v1.23.0). Thanks for the contribution, keep 'em coming 😉

RIanDeLaCruz:
@mattlewis92 Does this include locale? :) There is an open issue (#372) looking for TypeScript definitions for locale

mattlewis92:
Not currently, but it should be pretty trivial to add it 😃

RIanDeLaCruz:
Should we add it @LeshaKoss ?

LeshaKoss:
Yeah, sure

PR #348, opened 2016-12-21
Add support for durations

webpapaya:
This is a work in progress PR which will add support for durations to date-fns. It will mostly migrate functions from https://www.npmjs.com/package/pomeranian-durations to date-fns in order to create one comprehensive library to work with dates, times and durations.

closes https://github.com/date-fns/date-fns/issues/277
closes https://github.com/date-fns/date-fns/issues/284

PR #364, opened 2017-01-04
Add `subDuration` and `addDuration`.

webpapaya:
This PR adds 2 additional methods. This is the first PR which tries to incrementally add duration support to date-fns. The first PR adds  two new public methods which add or subtract an ISO8601 duration string to or from a given date.

In this PR there are also a couple of other functions scoped under `durations/` which are needed to implement `addDuration` and `subDuration`. Some of them might become public (documented) in one of the next PRs.


leshakoss:
Closing due to [this](https://github.com/date-fns/date-fns/pull/348#issuecomment-384172446

PR #367, opened 2017-01-04
Arabic language support

AbdallahAHO:
Arabic Language (Modern Standard Arabic - Al-fussha) support for date-fns.

### DIFF
--- a/src/locale/ar/build_distance_in_words_locale/index.js
+++ b/src/locale/ar/build_distance_in_words_locale/index.js
@@ -23,8 +23,8 @@ function buildDistanceInWordsLocale () {
     },
 
     aboutXHours: {
-      one: 'حوالي ساعة واحدة',
-      other: 'حوالي {{count}} ساعات'
+      one: 'ساعة واحدة تقريباً',
+      other: '{{count}} ساعات تقريباً'
     },
 
     xHours: {
@@ -38,8 +38,8 @@ function buildDistanceInWordsLocale () {
     },
 
     aboutXMonths: {
-      one: 'حوالي شهر واحد',
-      other: 'حوالي {{count}} أشهر'
+      one: 'شهر واحد تقريباً',
+      other: '{{count}} أشهر تقريباً'
     },
 
     xMonths: {
--- a/src/locale/ar/build_distance_in_words_locale/index.js
+++ b/src/locale/ar/build_distance_in_words_locale/index.js
@@ -1,0 +1,99 @@
+function buildDistanceInWordsLocale () {
+  var distanceInWordsLocale = {
+    lessThanXSeconds: {
+      one: 'أقل من ثانية واحدة',
+      other: 'أقل من {{count}} ثواني'
+    },
+
+    xSeconds: {
+      one: 'ثانية واحدة',
+      other: '{{count}} ثواني'
+    },
+
+    halfAMinute: 'نصف دقيقة',
+
+    lessThanXMinutes: {
+      one: 'أقل من دقيقة',
+      other: 'أقل من {{count}} دقيقة'
+    },
+
+    xMinutes: {
+      one: 'دقيقة واحدة',
+      other: '{{count}} دقائق'
+    },
+
+    aboutXHours: {
+      one: 'حوالي ساعة واحدة',
+      other: 'حوالي {{count}} ساعات'
+    },
+
+    xHours: {
+      one: 'ساعة واحدة',
+      other: '{{count}} ساعات'
+    },
+
+    xDays: {
+      one: 'يوم واحد',
+      other: '{{count}} أيام'
+    },
+
+    aboutXMonths: {
+      one: 'حوالي شهر واحد',
+      other: 'حوالي {{count}} أشهر'
+    },
+
+    xMonths: {
+      one: 'شهر واحد',
+      other: '{{count}} أشهر'
+    },
+
+    aboutXYears: {
+      one: 'عام واحد تقريباً',
+      other: '{{count}} أعوام تقريباً'
+    },
+
+    xYears: {
+      one: 'عام واحد',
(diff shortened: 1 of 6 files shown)


src/locale/ar/build_distance_in_words_locale/index.js
```diff
@@ -0,0 +1,99 @@
...
+      other: 'حوالي {{count}} أشهر'
+    },
+
+    xMonths: {
+      one: 'شهر واحد',
+      other: '{{count}} أشهر'
+    },
+
+    aboutXYears: {
+      one: 'عام واحد تقريباً',
```
ahmedelgabri: I see here that you are using تقريبا while up you are using حوالي, I guess تقريبا is the best for all of them

src/locale/ar/index.js
```diff
@@ -0,0 +1,11 @@
+var buildDistanceInWordsLocale = require('./build_distance_in_words_locale/index.js')
+var buildFormatLocale = require('./build_format_locale/index.js')
+
+/**
+ * @category Locales
+ * @summary English locale.
```
ahmedelgabri: Shouldn't this be Arabic? :)

AbdallahAHO: Missed this! :sweat_smile:

src/locale/ar/build_distance_in_words_locale/index.js
AbdallahAHO: Good point, Changed :+1:

src/locale/ar/build_format_locale/index.js
```diff
@@ -0,0 +1,79 @@
...
+  })
+
+  return {
+    formatters: formatters,
+    formattingTokensRegExp: buildFormattingTokensRegExp(formatters)
+  }
+}
+
+// Arabic Language doesn't have fixed prefix for ordinal Numbers similar to "st", "nd", "rd" or "th"
+// We Will need
```
LeshaKoss: What is the problem with ordinal numbers? It's ok to return just numbers if there are none, we do it in Icelandic locale IIRC. If so, please remove this comment

ahmedelgabri: @LeshaKoss I agree, makes more sense. As far as I know, ordinals are only used in formatting, right? https://date-fns.org/docs/format if this is the only case then I think numbers will be fine.

LeshaKoss: @ahmedelgabri yes, the locale returned by buildFormatLocale is used only in `format` ATM

src/locale/ar/build_format_locale/index.js
```diff
@@ -0,0 +1,79 @@
...
+
+function buildFormatLocale () {
+  // Note: in English, the names of days of the week and months are capitalized.
+  // If you are making a new locale based on this one, check if the same is true for the language you're working on.
+  // Generally, formatted dates should look like they are in the middle of a sentence,
+  // e.g. in Spanish language the weekdays and months should be in the lowercase.
+  var months3char = ['يناير', 'فبراير', 'مارس', 'أبريل', 'مايو', 'يونيو', 'يوليو', 'أغسطس', 'سبتمبر', 'أكتوبر', 'نوفمبر', 'ديسمبر']
+  var monthsFull = ['كانون الثاني يناير', 'شباط فبراير', 'آذار مارس', 'نيسان أبريل', 'أيار مايو', 'حزيران يونيو', 'تموز يوليو', 'آب أغسطس', 'أيلول سبتمبر', 'تشرين الأول أكتوبر', 'تشرين الثاني نوفمبر', 'كانون الأول ديسمبر']
+  var weekdays2char = ['أح', 'إث', 'ثل', 'أر', 'خم', 'جم', 'ست']
+  var weekdays3char = ['أحد', 'إثن', 'ثلاث', 'أرب', 'خمس', 'جمع', 'سبت']
```
ahmedelgabri: As far as I know in Arabic there is no 2 or 3char for days & for me it looks odd. If you check moment for example you will see that there is no 2 or 3 they just drop "ال" for the short version. & for the very short they use one letter.

https://github.com/moment/moment/blob/develop/locale/ar.js#L79-L80

This is even the case with nearly all Arabic dialects too
https://github.com/moment/moment/tree/develop/locale

For Egypt for example the 2char for Sunday اح won't sound nice at all.

LeshaKoss: @ahmedelgabri @AbdallahAHO it's ok to make it not 2 or 3 characters but just the shortest and a short abbreviation of the month/the day of the week. We do this in a lot of locales. I think it's 👍

AbdallahAHO: @ahmedelgabri, I agree with the weird أح, I was more into using moment.js format at first, but I wasn't sure if it's okay to make it more than 2 or 3 characters. However, It's fixed it now. :+1:

src/locale/ar/build_format_locale/index.js
AbdallahAHO: Comment removed! :+1:

LeshaKoss:
@AbdallahAHO @ahmedelgabri looks good! Is it ready to merge?

ahmedelgabri:
@LeshaKoss it's ok with me to be merged yes :)

LeshaKoss:
@AbdallahAHO thanks a lot for your awesome contribution! ⭐️ Arabic locale was released as [v1.24.0](https://github.com/date-fns/date-fns/releases/tag/v1.24.0) (diff: https://github.com/date-fns/date-fns/compare/v1.23.0...v1.24.0)

AbdallahAHO:
That's my first contribution to a public repo! :tada: 
Thank you @LeshaKoss for making date-fns possible, and Thanks for your help @ahmedelgabri, I am a big fan of your work.

PR #370, opened 2017-01-05
Flow typings generator

LeshaKoss:


LeshaKoss:
@mattlewis92 hello, I modified your generator to generate the flow typings as well, could you please check if I broke anything TypeScript-related? Thanks

LeshaKoss:
@mattlewis92 that was intended, I thought that is more resembling to standard.js style guide. Thanks for double checking!

PR #375, opened 2017-01-06
parse → toDate & new parse function (closes #219)

### LINKED ISSUE #219: Add an ability to pass the format to parse function, so it would be possible to parse dates in any format

leshakoss:


leshakoss:
@kossnocorp please review

fxck:
Any idea when is this gonna be released?

leshakoss:
@fxck probably this month

fxck:
Alright, cool, are you keeping builds / @next releases somewhere by any chance?

kossnocorp:
@fxck not yet, but alpha is coming really soon.

fxck:
Sorry to bother you again, but is that alpha anywhere near? I would help a ton. Thank you, thank you.

kossnocorp:
It is. We can provide you with an alpha, but the API could change as we are settling up some details ATM, so I'm not sure if it's a good idea. Which locales you use (if any)?

fxck:
I don't mind a slightest bit if the API changes. Just Czech locale at the moment.

PR #378, opened 2017-01-06
`getTime` function (closes #372)

### LINKED ISSUE #372: get unix timestamp of given date
Is it current supported? In moment we have **moment.valueOf()**

leshakoss:


src/get_time/test.js
```diff
@@ -0,0 +1,22 @@
...
+  })
+
+  it('accepts a string', function () {
+    var result = getTime(new Date(1483228800000).toISOString())
+    assert(result === 1483228800000)
+  })
+
+  it('accepts a timestamp (and returns it unchanged)', function () {
+    var result = getTime(new Date(1483228800000).getTime())
+    assert(result === 1483228800000)
```
silvenon: Since all of these timestamps are the same, maybe instead put it in a variable and reuse that? I know that hardcoding is in the spirit of testing, but in this case I think it would make the test easier to read.

OlivierCo: 👍

kossnocorp: I like the idea of having:

```js
var ts = 1483228800000
var result = getTime(new Date(ts).getTime())
assert(result === ts)
```

But moving it to a shared context (e.g. into the `describe` body) will introduce unnecessary cognitive load 🤔 . So I prefer to have it copy-pasted.

silvenon: Whatever you think is best. In my case I was distracted by trying to compare those numbers, only to realize that they are the same.

kossnocorp:
👍

PR #387, opened 2017-01-11
Add baseDate argument to parse

leshakoss:


CHANGELOG.md
```diff
@@ -144,7 +144,7 @@ rather than spread arguments.
 
   // v2.0.0 onward
   toDate('2016-01-01')
-  parse('2016-01-01', 'YYYY-MM-DD')
+  parse('2016-01-01', 'YYYY-MM-DD', new Date())
```
kossnocorp: Maybe it makes sense to put to the `options`?

leshakoss: @kossnocorp it makes sense, but would make the function non-pure:
```javascript
// when code is executed on 2016-12-31
parse('January', 'MMMM') //=> 2016-01-01
// when code is executed on 2017-01-01
parse('January', 'MMMM') //=> 2017-01-01
```

kossnocorp: @leshakoss please proceed, you're right. But please clarify the reason in the docs. Also, do we throw an exception when the arity is wrong?

leshakoss: @kossnocorp we don't do that in any function. Should we?

kossnocorp: Not sure, let's skip it for now and think about later. Please create an issue.

leshakoss:
@kossnocorp please review

src/parse/test.js
```diff
@@ -5,19 +5,23 @@ var assert = require('power-assert')
 var parse = require('./')
 
 describe('parse', function () {
+  beforeEach(function () {
+    this._baseDate = new Date(1986, 3 /* Apr */, 4, 10, 32, 0, 900)
+  })
```
kossnocorp: Let's not use state in tests, it's safe to use the same date object, so do like that:

```js
describe('parse', function () {
  const baseDate = new Date(1986, 3 /* Apr */, 4, 10, 32, 0, 900)
  // ...
```

leshakoss: @kossnocorp here or everywhere? There's quite a lot cases of this

leshakoss: @kossnocorp done

whitecolor:
@leshakoss 
Why is `baseDate` is required (in TS typings)?

leshakoss:
@whitecolor it being required makes the function pure. Before this PR, the current date was used as a base date (from which `parse` takes missing info), and therefore the function was returning different results for the same arguments when the time is passed. You can return the former behaviour like this:
```js
const impureParse = (dateString, formatString, options) =>
  parse(dateString, formatString, new Date(), options)
```

whitecolor:
I understand that this is needed for immutability in case when input doesn't contain year, though I belive is more rare case then providing full date. If you want this param still be required, at least I would allow `null` because code with just dummy `new Date()` *looks impure* too. Personally I would make this users responsibility to provide `baseDate` if they need it.

leshakoss:
@whitecolor it is the library's (as well as TS typing system's) responsibility to prevent errors caused by side effects, though.

And I think to have `baseDate = new Date()` is perfectly fine — it just explicitly states that in case of parsing ambiguity, the parser should default to `new Date()`. Yes, it looks impure, but that is the whole point of pure functions — to highlight this code as a side effect, and therefore as a potential source of bugs.

`baseDate` isn't used in your use case, but if you would extend your code to also parse, for example, dates in MM/DD format, this behaviour of `parse` function would suggest checking the value of baseDate carefully (as opposed to running silently, which could unexpectedly create a bug year later)

whitecolor:
> it is the library's (as well as TS typing system's) responsibility to prevent errors caused by side effects

But it is logic error. And actually you could make a handing of this case when there is no year and it is require and `thow` the error about it.

I just really think that demanding this param is redundant, most cases won't require it and users will just be puzzled: *what does `baseDate` mean and why I do I need to provide it?*. And in case if user supply only `dd/mm` it should be a logical error of him if he doesn't provide a date (a year) - so this problem is not caused by side effect, but mistakes in user's logic first of all.

This is not principle question for me, but I think API would be cleaner without this param required in general case.

leshakoss:
@whitecolor I think, throwing an exception when an optional argument is not provided is much worse API design.

whitecolor:
>I think, throwing an exception when an optional argument is not provided is much worse API design

ок, хозяин - барин)

PR #393, opened 2017-01-13
Add FP function generation (closes #253)

### LINKED ISSUE #253: FP (functional programming style functions)
Create special type of functions that allow FP-style curring:

```js
import format from 'date-fns/fp/format'
import partial from 'lodash/partial'
// ...
const formatShort = partial(format, 'YYYY-MM-DD')
```

versus:

```js
import format from 'date-fns/format'
// ...
const formatShort = (date) => format(date, 'YYYY-MM-DD')
```

leshakoss:


scripts/build_fp.js
```diff
@@ -0,0 +1,62 @@
...
+    .join('\n')
+
+  const fpFnWithOptionsLines = ['// This file is generated automatically. Please don\'t change it.']
+    .concat('')
+    .concat(`var fn = require('../../${camelCaseToSnakeCase(name)}/index.js')`)
+    .concat(`var convertToFpWithOptions = require('../_lib/convertToFpWithOptions/index.js')`)
+    .concat('')
+    .concat(`var ${name}WithOptions = convertToFpWithOptions(fn, ${arity})`)
+    .concat('')
+    .concat(`module.exports = ${name}WithOptions`)
```
kossnocorp: Let's generate ES6 versions here, ES5 versions will be compiled by Babel.

leshakoss:
@mattlewis92 hello, could you please give me a hand here? I tried to make TypeScript typings for FP functions #253 but I don't know how to do it correctly

mattlewis92:
Happy to help 😄 I can't see from the diff what issues you're facing, can you give me a specific example of where you're running into trouble?

leshakoss:
@mattlewis92 I generated a bunch of FP functions like these: https://github.com/date-fns/date-fns/pull/393/commits/39af0f2a7511e929fa75609226dcf4227b46de84#diff-0777f5ac1c488eba0e7fe9c0d2629658R1

They come in two variations, with options and without them:
```javascript
// regular function, from ./src/add_days
addDays(date, amount, options)

// FP function, from ./src/fp/addDays:
addDays(amount)(date)

// FP function with options, from ./src/fp/addDaysWithOptions
addDaysWithOptions(options)(amount)(date)
```

Basically, they are the same as regular functions but the order of arguments is reversed and arguments are curried:
```
const addDays = (b) => (a) => addDays(a, b)
const addDaysWithOptions = (options) => (b) => (a) => addDays(a, b, options)
```

I wanted to generate TypeScript typings for them:
https://github.com/date-fns/date-fns/pull/393/commits/a4cfe8fc05da7530abc882fb4dc9b7ecea6ba59f#diff-9b9d1a591a5fff70b4e5f52e2dbf48cbR937
But I don't really know if I did it correctly and if there is some idiomatic way to make typings for them. (I tried to check it in online TypeScript playground, but I don't think there's a way to check how separate typing files like `./typings.d.ts` work there. I don't know much about TypeScript module declarations and namespaces and stuff). Could you please help me with it?

mattlewis92:
Ah ok, maybe adding an e2e typescript test would be a good idea, similar to how definitely typed works?

leshakoss:
Could you please elaborate what do you mean or give an example?

mattlewis92:
So every lib on definitely typed has the typings and then a test file that uses the typings in some example usage. The idea is that if tsc can compile them then the typings are valid. For example: https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/webpack

leshakoss:
@mattlewis92 Oh, gotcha. I'll think about it. Well, for now, is
```typescript
declare module 'date-fns/fp' {
  let addDays: (number) => (Date | string | number) => (Date)
  namespace addDays {}
}

declare module 'date-fns/fp/addDays' {
  import {addDays} from 'date-fns/fp'
  export = addDays
}

declare module 'date-fns/fp/addDays/index' {
  import {addDays} from 'date-fns/fp'
  export = addDays
}
```
is the correct TypeScript declaration for this function?
```javascript
// ./src/fp/addDays
import _addDays from './../add_days/index.js'
const addDays = (b) => (a) => _addDays(a, b)
export default addDays
```
or should it be declared some other way? (like
```typescript
declare module 'date-fns/fp' {
  function addDays (number): ((Date | string | number) => (Date))
  namespace addDays {}
}
```
or something like this, I don't know)

mattlewis92:
Uh I'm not too sure off the top of my head, I will get back to you tomorrow 😃

leshakoss:
@mattlewis92 thanks 🙂

griffinmichl:
I was just typing up this exact feature request when I came across this PR.  
  
Thanks for the awesome work y'all are doing! Any idea when v2 will drop?

PR #394, opened 2017-01-14
add getNextMonth feature

OlivierCo:
- add getNextMonth feature to date-fns


leshakoss:
See https://github.com/date-fns/date-fns/pull/401#issuecomment-273828200

PR #395, opened 2017-01-14
Improve Japanese locale

ykzts:
"Nごろ" is a sentence not used in Japan. and... The word "半分" does not mean "half a minute".


leshakoss:
@ykzts thanks for your fixes! 🇯🇵 They were released with [v1.26.0](https://github.com/date-fns/date-fns/releases/tag/v1.26.0) (diff: https://github.com/date-fns/date-fns/compare/v1.25.0...v1.26.0)

PR #400, opened 2017-01-16
Added strftime function.

ShogunPanda:
Hi!
This PR adds strftime, basing on the format function.

Hope this helps!
  Shogun


ShogunPanda:
FYI: I fixed the tests but the Travis build failed. Anyone knows why?

ShogunPanda:
Forget about my comment above. It was a lint problem.
Now everything is fresh and clean! :)

leshakoss:
@ShogunPanda thanks a lot for your contribution! We (date-fns team) want to cover all use cases for date manipulation in JavaScript, however we don't want to duplicate already covered functionality (in this case, functionality of `format`). That's why I think this function should be a separate npm package, like `strftime-date-fns` or something like that. Looking forward to see this as a separate library!

ShogunPanda:
Cool, will release it soon :)

ShogunPanda:
Done! Released on both npm and bower as: [date-fns-strftime](https://www.npmjs.com/package/date-fns-strftime)

PR #401, opened 2017-01-16
add getNextYear feature

OlivierCo:
- add getNextYear feature
- add default value to getYear param, if no params given, so dirtyDate = new Date()
(can be use without params and return the current year)

getYear()
// => 2017

### DIFF
--- a/src/get_next_year/index.js
+++ b/src/get_next_year/index.js
@@ -1,0 +1,26 @@
+var parse = require('../parse/index.js')
+var getYear = require('../getYear/index.js')
+
+/**
+ * @category Month Helpers
+ * @summary Get the next year of the given date.
+ * 
+ * @description
+ * Get the next year of the given date.
+ * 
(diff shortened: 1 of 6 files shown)


leshakoss:
@OlivierCo hello, thanks a lot for your contribution! However, we (date-fns team) try to keep our API for date manipulation as full and simple as possible, without duplication of already covered use cases. `getNextYear` can be easily substituted with `const getNextYear = date => getYear(date) + 1`, and your other function: #394 can be replaced with `const getNextMonth = date => (getMonth(date) + 1) % 12`. We already started to remove duplicating functionality (see #377) and `getNextYear` and `getNextMonth` would create a new class of functions: `getNextISOYear`, `getNextMillisecond`, `getLastYear`, ... that would make the library bigger without adding new functionality. If you use these a lot in your projects, we encourage to create new npm packages based on date-fns, so you may want to do that!

PR #379, opened 2017-01-07
ES Modules (closes #179)

kossnocorp:
The PR converts code into ES that enables tree-shaking. See the parent PR: https://github.com/date-fns/date-fns/pull/296

## TODO

- [ ] Make sure that all locales are exported as an object
- [x] Get rid of `module.exports` in the source code
- [ ] Finalize `build_package` script
- [ ] Add smoke tests for the built package

PR #404, opened 2017-01-18
ES Modules (closes #179, closes #414) 

### LINKED ISSUE #179: Enable tree shaking for date-fns

### LINKED ISSUE #414: Expose `module` field in package.json
To support sane modern es6 aware bundlers do their job better.

ref: https://github.com/rollup/rollup/wiki/pkg.module
~~_(there was some wiki page that explain few variants and why they are rejected but can't find it)_~~

ooh, found it: https://github.com/nodejs/node/wiki/ES6-Module-Detection-in-Node, specifically https://github.com/nodejs/node/wiki/ES6-Module-Detection-in-Node#option-4-meta-in-packagejson. It is "rejected" but is best working approach currently and is used by Webpack and Rollup

leshakoss:
The PR converts code into ES that enables tree-shaking. See the parent PR: https://github.com/date-fns/date-fns/pull/379

## TODO

- [x] Make sure that all locales are exported as an object
- [x] Get rid of `module.exports` in the source code
- [x] Finalize `build_package` script
- [x] Add smoke tests for the built package


package.json
```diff
@@ -11,10 +11,8 @@
   "typings": "./typings.d.ts",
   "main": "index.js",
   "module": "es/index.js",
-  "dependencies": {
-    "babel": "^6.5.2"
```
leshakoss: @kossnocorp should babel be in non-dev dependencies?

kossnocorp: @leshakoss nope

corysimmons:
So master is still on 1.x, v2 is still on 1.x.

How can we npm install this functionality?

leshakoss:
@corysimmons you can use `scripts/buildPackage.sh` script to compile the package from `v2` branch:

```bash
git clone https://github.com/date-fns/date-fns.git date-fns-v2
cd date-fns-v2
git checkout v2
yarn
env PACKAGE_OUTPUT_PATH="$(pwd)/../YOUR-MODULE/node_modules/date-fns" ./scripts/buildPackage.sh
cd ..
```
(date-fns v2 will appear in `./YOUR-MODULE/node_modules/date-fns`)

I would advise against doing in any code, production or not, since we are planning to change I18n API before an alpha release.

corysimmons:
Okay thank you.

Any super-general ETA on an API-stable release of this functionality? Our team is interested in using this vs. moment, but mostly for the weight reduction, and tree-shaking would be 👌

leshakoss:
@corysimmons probably a month or so

knieber:
So is tree-shaking still not usable with any release? We've tried destructured importing and have tree-shaking enabled, but still seeing the entire library being brought into our prod build. We've used v1.30 and v2.0.0-alpha27

TrySound:
@knieber Which bundler do you use?

knieber:
@TrySound webpack 4

TrySound:
Do you use `mode: production`?

knieber:
@TrySound yes. We are using Next.js and it "should" be setting that for us when running their built in `build` command. It's possible that it's some config with next that's the problem. So, is tree-shaking already support by date-fns in v2.0.0-alpha27?

TrySound:
Module field with esm syntax is provided. Big part of api is function. So treeshaking should work.
https://unpkg.com/date-fns@2.0.0-alpha.27/package.json

And I noticed similar problem in next.js too earlier. Not sure how to solve it.

knieber:
@TrySound Ah, I see. That was my misunderstand. Yes, it seems like there are some issues with tree-shaking this package in next.js. Thanks for the help!

corysimmons:
Is it just this package or is tree-shaking just broken in Next?

PR #406, opened 2017-01-18
add backers and sponsors

piamancini:
your open collective backers and sponsors can now appear directly on your readme. 
see how it'll look: https://github.com/apex/apex#backers
More info: https://github.com/opencollective/opencollective/wiki/Github-banner
Also add badges on top

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,6 @@
 # ![](http://cdn.date-fns.org/README-logo.svg) date-fns
+[![OpenCollective](https://opencollective.com/date-fns/backers/badge.svg)](#backers) 
+[![OpenCollective](https://opencollective.com/date-fns/sponsors/badge.svg)](#sponsors)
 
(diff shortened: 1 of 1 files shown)


kossnocorp:
@piamancini thanks you, I appreciate your effort but I'm going to close this PR. I don't like the current approach as it makes dozens of unnecessary HTTP requests 😬 I prefer to keep README as simple as possible. In the future when I'll have time to manage Open Collective and run a campaign, I'll add backers to https://date-fns.org/ instead.

piamancini:
Sounds good! 
Quick question: Do you want to keep your collective open? If so, please edit the Sponsor tier to remove the reference to their logo showing up on your readme. It's better to be transparent and avoid confusions. You can log in https://opencollective.com/date-fns and manage them from Edit Collective on the menu bar.

PR #416, opened 2017-01-24
Dutch locale: fix hours/years plural and in/over confusion

rubenstolk:
I think that @jtangelder will agree with this fix:

#### `over` confusion
- [x] "over XXX" in English is being translated as "meer dan XXX" (more than)
- [x] "in XXX" in English should be translated as "over XXX"

#### `plural` confusion
- [x] In Dutch it's more common to use "years" as singular, even if it's plural (`over 4 jaar` rather than `over 4 jaren`
- [x] In Dutch it's more common to use "hours" as singular, even if it's plural (`over 4 uur` rather than `over 4 uren`
- [x] "Half a minute" should be translated as "Een halve minuut"

### DIFF
--- a/src/locale/nl/build_distance_in_words_locale/index.js
+++ b/src/locale/nl/build_distance_in_words_locale/index.js
@@ -10,7 +10,7 @@ function buildDistanceInWordsLocale () {
       other: '{{count}} seconden'
     },
 
-    halfAMinute: 'halve minuut',
+    halfAMinute: 'een halve minuut',
 
     lessThanXMinutes: {
       one: 'minder dan een minuut',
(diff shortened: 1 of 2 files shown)


jtangelder:
I agree!

rubenstolk:
Yep, done, also added a change for "Een halve minuut"

leshakoss:
@rubenstolk thanks a lot for your fixes! ⭐️ I finally got around to release them as [v1.27.2](https://github.com/date-fns/date-fns/releases/tag/v1.27.2) (diff: https://github.com/date-fns/date-fns/compare/v1.27.1...v1.27.2)

PR #432, opened 2017-02-12
Rename `es` submodule to `esm`

leshakoss:


CHANGELOG.md
```diff
@@ -239,6 +239,14 @@ rather than spread arguments.
   for consistency and future features.
   See [docs/Options.js](https://github.com/date-fns/date-fns/blob/master/docs/Options.js)
 
+- Added support for ECMAScript modules via `'date-fns/esm'` subpackage:
```
kossnocorp: [ECMAScript **M**odules](http://www.ecma-international.org/ecma-262/6.0/#sec-modules) (also add the link, please)

CHANGELOG.md
```diff
@@ -239,6 +239,14 @@ rather than spread arguments.
...
+
+  ```javascript
+  import format from 'date-fns/esm/format'
+  // is same as
+  const format = require('date-fns/format')
+  ```
```
kossnocorp: The whole point of having ESM is tree-shaking. I think it's worth to emphasize it: `import {format} from 'date-fns/esm'`.

leshakoss: @kossnocorp tree-shaking doesn't work by default without any action, isn't it?

kossnocorp: It's not but ESM intended to be used with the bundles that support tree-shaking (like webpack or Rollup). There are no point in using it if tree-shaking is not enabled.

leshakoss: @kossnocorp well there's also TypeScript where you should either use 'date-fns/esm' or add `* as` to import:
```typescript
import * as format from 'date-fns/format'
// or
import format from 'date-fns/esm/format'

kossnocorp: Good point, worth to reflect in the documentation too.

PR #434, opened 2017-02-12
Parallelize Travis CI tests

leshakoss:


scripts/test_tz_extended.sh
```diff
@@ -344,7 +344,7 @@ then
     "Pacific/Wallis"
   do
     printf "Run test in time zone $tz\n"
-    env TEST_TZ=true USE_STATIC_TESTS=true TZ=$tz npm run test-ci \
+    env TEST_TZ=true USE_STATIC_TESTS=true TZ=$tz yarn run test-ci \
       &>tmp/last_test_output.txt || (cat tmp/last_test_output.txt && exit 1) || exit 1
   done
-fi
+#fi
```
kossnocorp: Did you comment this out to test it?

leshakoss: yep. Though maybe we _should_ test timezones in PRs too, it would save us some last-minute-before-releasing fixes

kossnocorp: You're right. How long it takes to build the whole suite now? Also what will happen with the tests counter we have?

leshakoss: @kossnocorp test counters are separate for each suite.

But it all doesn't matter because paralleling tests for some reason doesn't work too:
https://travis-ci.org/date-fns/date-fns/jobs/200849169#L542

leshakoss: I think we could parallelize even more by splitting `test-tz-extended` to multiple sub-tests. Or use something faster instead of Chromium

PR #436, opened 2017-02-15
Adopt camelCase naming schema (closes #220)

### LINKED ISSUE #220: Adopt camelcase file naming schema
Why does `date-fns` use lowercase for file names?
`import isSameDay from 'date-fns/is_same_day'`

Lodash uses camelCase and look like has no any problems
`import keyBy from 'lodash/keyBy'`

leshakoss:


scripts/buildIndex.js
```diff
@@ -1,7 +1,7 @@
 import fs from 'fs'
 import path from 'path'
-import listFiles from './_lib/list_files'
-import listFpFiles from './_lib/list_fp_files'
+import listFiles from './_lib/listFiles'
+import listFpFiles from './_lib/listFpFiles'
```
kossnocorp: FP is an abbr, hence should be `listFPFiles`. When the var name starts with an abbr, then it should be all lower-cased:

- `listFPFiles`
- `fpFiles`

PR #443, opened 2017-02-17
Implicit conversion of all arguments (closes #442, closes #389)

### LINKED ISSUE #389: Handle the wrong number of arguments / Check the type of arguments
What should the functions do when user didn't pass the right number of arguments or the passed arguments are not of expected type? Should this case be handled?

### LINKED ISSUE #442: Implicit conversion of all arguments
`addDays(new Date(), '10')` should add 10 days (it doesn't now)
See also #389

leshakoss:


leshakoss:
Explicit?

Issue #437, opened 2017-02-15
Format day of month ('Do') french translation error


peterpeterparker:
Hi,

I think that the format of the day of month ('Do') isn't correct in french, it doesn't follow the same rules as in english or german for example or even doesn't follow the same logic as day of the year of week in french.

day of month ('Do'):

1st, 2nd, ..., 31st in english is correct
or
1., 2., ... 31. in german is correct too

but for day of the month in french, you not gonna says
1er, 2e, ... 31e 

but 

1er, 2, ... 31 

like "le 1er janvier 2017" but like "le 2 janvier 2017", ... "le 31 janvier 2017"

Therefore I think that the function ordinal in src/locale/fr/build_format_locale/index.js should be extended, when the number is greater than 1, the code "number + 'e'" should only apply for day of the week and day of the year where it should only return "number" when day of the month.

Best regards

kossnocorp:
Thank you for the report! cc @date-fns/i18n-french

fbonzon:
I'm trying to fix the exact same problem in Moment.js. (And might fix it here too.) Please read related issue https://github.com/moment/moment/issues/3375#issuecomment-280949260

The `Do` token is correctly implemented here, in my understanding. It's a question of definition and documentation. The definition of `Do` is "day of month with ordinal suffix". Example usage:

#### English
2017-03-01 → `Do` day of the month → 1st day of the month
2017-03-23 → `Do` day of the month → 23rd day of the month
2017-03-24 → `Do` day of the month → 24th day of the month

#### French
2017-03-01 → `Do` jour du mois → 1er jour du mois
2017-03-23 → `Do` jour du mois → 23e jour du mois
2017-03-24 → `Do` jour du mois → 24e jour du mois

When writing long date format in French, we want to use ordinal for day 1 (token `Do`) and cardinal for other days (token `D`). There is currently no way to express this in the locale files.

Solutions I can think of:

1. Create a new token with definition "day of month for long date format (with cardinal or ordinal depending on locale)". For French, it would return the same value as `Do` for day 1, and same value as `D` otherwise.

Name it `Dom` for Day of month (latest suggestion in the above referred issue). Or `F` (next letter after D, and E is already taken) ?

Or maybe this problem is too specific of the French language, doesn't merit a dedicated token for all languages, and is semantically better handled within the French locale code. This brings second solution:

[...]

PR #448, opened 2017-02-23
Honor grammatical gender in French ordinals

fbonzon:
Ran `npm run lint` and `npm run test`.

### DIFF
--- a/src/locale/fr/build_format_locale/index.js
+++ b/src/locale/fr/build_format_locale/index.js
@@ -11,27 +11,27 @@ function buildFormatLocale () {
   var meridiemFull = ['du matin', 'de l’après-midi', 'du soir']
 
   var formatters = {
-    // Month: Jan, Feb, ..., Dec
+    // Month: Jan, Feb, …, Dec
     'MMM': function (date) {
       return months3char[date.getMonth()]
     },
 
-    // Month: January, February, ..., December
+    // Month: January, February, …, December
     'MMMM': function (date) {
       return monthsFull[date.getMonth()]
     },
 
-    // Day of week: Su, Mo, ..., Sa
+    // Day of week: Su, Mo, …, Sa
     'dd': function (date) {
       return weekdays2char[date.getDay()]
     },
 
-    // Day of week: Sun, Mon, ..., Sat
+    // Day of week: Sun, Mon, …, Sat
     'ddd': function (date) {
       return weekdays3char[date.getDay()]
     },
 
-    // Day of week: Sunday, Monday, ..., Saturday
+    // Day of week: Sunday, Monday, …, Saturday
     'dddd': function (date) {
       return weekdaysFull[date.getDay()]
     },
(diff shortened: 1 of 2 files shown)


fbonzon:
@date-fns/i18n-french

src/locale/fr/build_format_locale/index.js
```diff
@@ -76,9 +84,18 @@ function buildFormatLocale () {
   }
 }
 
-function ordinal (number) {
+// Gender:
+// 0 = masculin
```
leshakoss: I would prefer `masculineOrdinal` and `feminineOrdinal` to be separate functions, to get rid of magic numbers

src/locale/fr/build_format_locale/index.js
```diff
@@ -63,10 +63,18 @@ function buildFormatLocale () {
...
   ordinalFormatters.forEach(function (formatterToken) {
     formatters[formatterToken + 'o'] = function (date, formatters) {
-      return ordinal(formatters[formatterToken](date))
+      return ordinal(formatters[formatterToken](date), 0)
+    }
+  })
+  // Words with feminine grammatical gender: semaine
+  ordinalFormatters = ['W']
+  ordinalFormatters.forEach(function (formatterToken) {
+    formatters[formatterToken + 'o'] = function (date, formatters) {
```
leshakoss: Just `formatters['Wo'] = ...` without `forEach` loop is okay, I think

leshakoss: Or even put it right in `formatters` object above

src/locale/fr/build_format_locale/index.js
```diff
@@ -76,9 +84,18 @@ function buildFormatLocale () {
...
+    switch (gender) {
+      case 0:
+      default:
+        return '1er'
+      case 1:
+        return '1re'
+    }
   }
 
   return number + 'e'
```
leshakoss: What about what @peterpeterparker said here: https://github.com/date-fns/date-fns/issues/437#issue-207933466?
> but for day of the month in french, you not gonna says
1er, 2e, ... 31e
> 
> but
> 
> 1er, 2, ... 31

Could you please give me examples where it should be that way (1er, 2, 3 ...) and where it should be the other way (1er, 2e, 3e)? I am not sure I understand

src/locale/fr/build_format_locale/test.js
```diff
@@ -291,7 +291,7 @@ describe('fr locale > buildFormatLocale', function () {
 
     describe('Wo', function () {
       it('returns ordinal result of W formatter', function () {
-        assert(buildFormatLocale().formatters.Wo(null, {W: function () { return 1 }}) === '1er')
+        assert(buildFormatLocale().formatters.Wo(null, {W: function () { return 1 }}) === '1re')
```
leshakoss: 👍

src/locale/fr/build_format_locale/index.js
fbonzon: This is an independant problem. Like said in #437, the definition of token `Do` is "day of month with ordinal", and correctly implemented with 1er, 2e, 3e, …

Like others tokens with `o` at the end, it only says that it's the ordinal variant of the same token without `o` (the cardinal variant). It doesn't say it's the token to use when writing dates in long format (with month name). Examples:

EN: The `Do` day of the month of `MMMM`→ The 2nd day of the month of June
FR: Le `Do` jour du mois de `MMMM`→ The 2e jour du mois de juin

1er, 2, 3, … is used in long date formats, where the day of the month is followed by `MMM` or `MMMM`. It's neither `D` or `Do`, it depends on day of month. `Do` for day 1, `D` otherwise.

When speaking or writing dates in English, both cardinal figures (June 2, token `D`) and ordinal figures (June 2nd, token `Do`) are possible. Depends on context, intended style. A given date format will use `D` or `Do` consistently for every day of the month.

When speaking or writing dates in French, there is no choice really, the rule is always the same: use ordinal for day 1 (suffix "er", format `Do`), and cardinal (format `D`) for other days. That's for the specific context of dates with month.

I don't think it's correct to make `Do` return 1er, 2, 3, … because it can be used in other contexts. See example above "The `Do` day of the month".

fbonzon:
All comments addressed.

fbonzon:
Obsoleted by #449

PR #449, opened 2017-02-23
Fix ordinals in French locale

### LINKED ISSUE #437: Format day of month ('Do') french translation error

fbonzon:


docs/i18n.md
```diff
@@ -155,4 +155,13 @@ use this quick guide:
...
 
+### Priority of Tokens
+
+A locale in date-fns is specified by defining formatters for the various tokens that the [format](https://date-fns.org/docs/format) function understands. To handle special cases, it is possible to group several tokens in a bigger token, and write a formatter for it as a whole. Example from the French locale:
+
+- Formatter for `Do` returns 1er, 2e, 3e, …
+- Formatter for `Do MMMM` returns 1er janvier, 2 janvier, 3 janvier, … Note the ordinal suffix disappears when day of month > 1.
+
+When formatting a date using a format string, the priority of tokens is **reverse alphabetical order**. E.g. first the format string is searched for `Do MMMM`, then it is searched for `Do`, because `Do MMMM` comes before `Do` when sorted alphabetically in reverse order.
+
```
leshakoss: @fbonzon well, that's nice and all, but I think this text belongs not in I18n docs but rather in comments somewhere in `format` function, or at least in `fr` locale.

Firstly, this information is a bit irrelevant for all people who read I18n doc — it's about internal algorithm of `format` and doesn't help for all the people who want to understand how to use locales.

Secondly, if I understand the algorithm correctly, there's not really any ambiguity how `format` should split a string of tokens: if there is 'Do MMMM', it applies 'Do MMMM', if there's 'Do' or 'MMMM' without another one, it does that. Well, 'MMMMMM' is ambiguous, but just 'greedy matching' is enough to tell how `format` works in this case.

Thirdly, locales don't have to be that way — they are just a regexp and a collection of functions and can be anything, so this info may be not true for all languages.

Fourthly, 'Do MMMM' in French locale and similar tokens in Russian locale seem like dirty hacks to me rather than real formatter tokens, so we probably should move the explanation how they work somewhere to comments in code.

I feel like we should have more extended jsdoc for locales that we could show on date-fns.org, with examples and stuff like that, but I still haven't got around to do that, so maybe we could start with French locale and move the text above into 'locales/fr/index.js' jsdoc

fbonzon: OK. Moved to code comments.

I agree doc for end users and developers should be separate. Note that's why I added the text here inside the "Adding New Language" section, which is not intended for users, but developers contributing to locales. Probably you should move that section to the CONTRIBUTING file too.

fbonzon:
Comments addressed.

PR #454, opened 2017-02-28
Remove exception in format (closes #304)

### LINKED ISSUE #304: format returns "undefined NaN, 0NaN" for invalid dates
Here is the test case:

https://runkit.com/581c7c34d33aac0014dfc5d3/5847e8805c9e5c001426d5de/branches/master

It should probably use `isValid` before trying to format and throw in this case?

leshakoss:
Reverts #392

### DIFF
--- a/src/format/test.js
+++ b/src/format/test.js
@@ -315,9 +315,8 @@ describe('format', function () {
   })
 
   describe('edge cases', function () {
-    it('throws an exception if the date isn\'t valid', function () {
-      var block = format.bind(null, new Date(NaN), 'MMMM D, YYYY')
-      assert.throws(block)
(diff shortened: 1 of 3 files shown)


src/format/test.js
```diff
@@ -315,9 +315,8 @@ describe('format', function () {
   })
 
   describe('edge cases', function () {
-    it('throws an exception if the date isn\'t valid', function () {
-      var block = format.bind(null, new Date(NaN), 'MMMM D, YYYY')
-      assert.throws(block)
+    it('returns String(\'Invalid Date\') if the date isn\'t valid', function () {
```
kossnocorp: `it("returns String('Invalid Date') if the date isn't valid", function () {` next time, please

PR #460, opened 2017-03-10
Fix corner cases

leshakoss:


src/isSameDay/test.js
```diff
@@ -36,4 +36,20 @@ describe('isSameDay', function () {
...
+    assert(result === false)
+  })
+
+  it('returns false if the second date is `Invalid Date`', function () {
+    var result = isSameDay(
+      new Date(1987, 1 /* Feb */, 11),
+      new Date(NaN)
+    )
+    assert(result === false)
+  })
```
kossnocorp: It's worth to add a tests case with `isSameDay(new Date(NaN), new Date(NaN))`

src/isAfter/test.js
```diff
@@ -44,4 +44,20 @@ describe('isAfter', function () {
...
+    assert(result === false)
+  })
+
+  it('returns false if the second date is `Invalid Date`', function () {
+    var result = isAfter(
+      new Date(1987, 1 /* Feb */, 11),
+      new Date(NaN)
+    )
+    assert(result === false)
+  })
```
kossnocorp: Add test case for `isAfter(new Date(NaN), new Date(NaN))`

src/areIntervalsOverlapping/test.js
```diff
@@ -167,4 +167,42 @@ describe('areIntervalsOverlapping', function () {
...
+        {start: new Date(2016, 10, 3), end: new Date(2016, 10, 7)},
+        {start: new Date(NaN), end: new Date(2016, 10, 5)}
+      )
+      assert.throws(block)
+    })
+
+    it('throws an exception if the end date of the compared time interval is `Invalid Date`', function () {
+      var block = areIntervalsOverlapping.bind(
+        null,
+        {start: new Date(2016, 10, 3), end: new Date(2016, 10, 7)},
```
kossnocorp: Let's use `RangeError` for the exception.

src/closestIndexTo/test.js
```diff
@@ -47,4 +47,23 @@ describe('closestIndexTo', function () {
...
+
+  it('returns NaN if any date in the given array is `Invalid Date`', function () {
+    var date = new Date(2014, 6 /* Jul */, 2)
+    var result = closestIndexTo(date, [
+      new Date(2015, 7 /* Aug */, 31),
+      new Date(NaN),
+      new Date(2012, 6 /* Jul */, 2)
+    ])
+    assert(isNaN(result))
+  })
```
kossnocorp: Let's add test case when the given array contains an unexpected value such as `undefined`.

CHANGELOG.md
```diff
@@ -328,6 +328,10 @@ This change log follows the format documented in [Keep a CHANGELOG].
   )
   ```
 
+- **BREAKING**: functions now throw `RangeError` if optional values passed to `options`
+  are not `undefined` or have expected values.
+  This change is introduced for consistency with ECMAScript's `Intl` which does the same.
```
kossnocorp: Add a comment about arithmetic operations with `NaN`

PR #467, opened 2017-03-22
Remove dots in `dd` and `ddd` in Polish format locale

justrag:
The dots after weekdays2char and weekdays3char are kinda silly - you can always add them manually if you need them but you cannot remove them. And they're not in any way standard.

### DIFF
--- a/src/locale/pl/build_format_locale/index.js
+++ b/src/locale/pl/build_format_locale/index.js
@@ -3,7 +3,7 @@ var buildFormattingTokensRegExp = require('../../_lib/build_formatting_tokens_re
 function buildFormatLocale () {
   var months3char = ['sty', 'lut', 'mar', 'kwi', 'maj', 'cze', 'lip', 'sie', 'wrz', 'paź', 'lis', 'gru']
   var monthsFull = ['styczeń', 'luty', 'marzec', 'kwiecień', 'maj', 'czerwiec', 'lipiec', 'sierpień', 'wrzesień', 'październik', 'listopad', 'grudzień']
-  var weekdays2char = ['nd', 'pn', 'wt', 'śr', 'cz', 'pt', 'sob']
+  var weekdays2char = ['nd', 'pn', 'wt', 'śr', 'cz', 'pt', 'sb']
   var weekdays3char = ['ndz', 'pon', 'wt', 'śr', 'czw', 'pt', 'sob']
   var weekdaysFull = ['niedziela', 'poniedziałek', 'wtorek', 'środa', 'czwartek', 'piątek', 'sobota']
   var meridiem = ['w nocy', 'rano', 'po południu', 'wieczorem']
--- a/src/locale/pl/build_format_locale/index.js
+++ b/src/locale/pl/build_format_locale/index.js
@@ -3,8 +3,8 @@ var buildFormattingTokensRegExp = require('../../_lib/build_formatting_tokens_re
 function buildFormatLocale () {
   var months3char = ['sty', 'lut', 'mar', 'kwi', 'maj', 'cze', 'lip', 'sie', 'wrz', 'paź', 'lis', 'gru']
   var monthsFull = ['styczeń', 'luty', 'marzec', 'kwiecień', 'maj', 'czerwiec', 'lipiec', 'sierpień', 'wrzesień', 'październik', 'listopad', 'grudzień']
-  var weekdays2char = ['nd.', 'pn.', 'wt.', 'śr.', 'cz.', 'pt.', 'sob.']
-  var weekdays3char = ['ndz.', 'pon.', 'wt.', 'śr.', 'czw.', 'pt.', 'sob.']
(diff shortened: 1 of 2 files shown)


leshakoss:
@date-fns/i18n-polish

justrag:
I revised the tests, too. And fixed "sob" -> "sb" in 2chars.

ertrzyiks:
Actually, there are rules standardizing usage of a dot in shortcuts and one of them says to put dot when the last letter in a shortcut is NOT the last letter of a word. Also you can not add a dot manually only for `pl` locale, you will need to do this for all locales if you decide to have a dot after `dd`.

However, I think it makes sense to remove dots, but for the sake of consistency. I can not find any other locale with a dot here.

piotrd:
I just took a look at how it is handled e.g. in macOS, and in locale preferences I've found the following options for names of days:

image

So as you can see, there is a version of 3-character day shortcuts with and without a dot. Maybe that's the way to go, i.e. introduce 2 new formats: `weekdays2charWithDot` and `weekdays3charWithDot`?

leshakoss:
@ertrzyiks @piotrd `dd` and `ddd` formatters don't have to be exactly 2 and 3 characters, they are just "short" and "shorter" versions of weekdays and can be any number of characters which make more sense in localized formatted date. As for this locale, I suggest using these values (which I randomly found on internet): http://www.unicode.org/cldr/charts/29/summary/pl.html#1825 http://www.unicode.org/cldr/charts/29/summary/pl.html#1839 I don't know how much is this a standard of some sort or anything so maybe I am wrong on this

justrag:
Actually, the version I'm interested in most is the one I commonly see in the wall calendars: pn wt śr cz pt sb nd. Short and to the point.

piotrd:
I have nothing against the proposed format (although wall calendars are no authority to me), but I can imagine someone else coming in a few days and asking to take that back :)

On a more serious side, I value consistency above individual preferences. I took a look at other locale in date-fns (not all, but most of them) and found dots only in french ones (and only in 3char version). So I think the change proposed in this PR is consistent with majority and can be merged.

ertrzyiks:
or, if we take that `dd` is a short symbol and `ddd` is an abbreviation then we can follow french translation and remove dots only from the shortest version. Similar idea is behind unicode.org tables linked by @leshakoss

piotrd:
This way or another, @leshakoss or @kossnocorp should make the decision and IMHO enforce consistency across all locale.

leshakoss:
personal preference < consistency < rules/standards of languages

It's the same issue as with `a`/`A`/`aa` formatters and AM/PM; and with capital first letter of months and weekdays: different languages have different rules and I would prefer `format` to format dates like if native speaker wrote them, as opposed to direct translation of English language locale.

For `dd` (aka `weekdays2char `), calendars sound like a good point of reference (I think `Su`, `Mo`, ..., `Sa` of `en` locale are taken from calendars too)

`ddd` (`weekdays3char`) is supposed to look like an abbreviation in the middle of the text or, for example, in some official document.

It is true that most `date-fns` locales use the same 2- and 3-letter formats for weekdays, but I don't really know how many of those are artifacts of direct translation of English language locale and should be fixed.

piotrd:
If we treat `weekdays2char` as the shortest meaningful format, then I don't see a reason to keep dots in it.

However, for the `weekdays3char`, as @leshakoss suggested, I would stick to version officially acknowledged as correct by Polish language dictionaries - with a dot:

> poniedziałek => pon./ pn. 
wtorek => wt. 
środa => śr. 
czwartek => czw. 
piątek => piąt./ pt. 
sobota => sob. 
niedziela => niedz.

(which also means we have Sunday abbreviated incorrectly in locale)

References: [1](http://portalwiedzy.onet.pl/140198,,,,dni_tygodnia_-_pisownia_skrotow,haslo.html?drukuj=1), [2](http://sjp.pwn.pl/poradnia/haslo/skroty-dni-tygodnia;2315.html), [3](http://www.edupedia.pl/words/index/show/401176_slownik_poprawnej_polszczyzny-dni_tygodnia.html).

> It is true that most date-fns locales use the same 2- and 3-letter formats for weekdays, but I don't really know how many of those are artifacts of direct translation of English language locale and should be fixed.

Good point. I don't know either.

justrag:
Ammended the PR according to discussion.

> var weekdays2char = ['nd', 'pn', 'wt', 'śr', 'cz', 'pt', 'sb']
> var weekdays3char = ['niedz.', 'pon.', 'wt.', 'śr.', 'czw.', 'piąt.', 'sob.']

leshakoss:
@justrag thanks a lot for your contribution and you guys @date-fns/i18n-polish for the discussion! ⭐️ This fix was released as [v1.28.2](https://github.com/date-fns/date-fns/releases/tag/v1.28.2) (diff: https://github.com/date-fns/date-fns/compare/v1.28.1...v1.28.2)

justrag:
You're welcome and thanks for the org invitation :)
(I think my PR actually removes more characters than it adds ;) )

PR #471, opened 2017-04-05
Remove `isDate`

leshakoss:


nunocastromartins:
Sorry to comment on this old pull request, but why was this functionality removed?
Calling `instanceof Date` fails when `Date` objects are passed across frames.
Other libraries such as lodash instead do something like:
```js
typeof date === 'object' && Object.prototype.toString.call(date) === '[object Date]'
```
See lodash implementation (which seems to have a performance optimisation for Node.js): https://github.com/lodash/lodash/blob/master/isDate.js

I see no reason why this function shouldn't be provided by date-fns.
Thank you for your work!

nunocastromartins:
Just adding that a proper implementation of `isDate` would solve #541.
Calling `instanceof` is indeed not a good practice on a functional library such as date-fns since pure functions shouldn't access global state (and `Date` is a global variable).

Just leaving my two cents here.

kossnocorp:
@nunocastromartins this is a very good point. @leshakoss WDYT?

Issue #472, opened 2017-04-07
setDay() should handle `weekStartsOn` differently (or the documentation needs to change!)


gausie:
At the moment setDay() works as follows

```javascript
  it('allows to specify which day is the first day of the week', function () {
    var result = setDay(new Date(2014, 8 /* Sep */, 1), 0, {weekStartsOn: 1})
    assert.deepEqual(result, new Date(2014, 8 /* Sep */, 7))
  })
```

(for reference the 1st of September 2014 was a Monday)

This is initially very confusing to me, and I would expect the function to work as follows:

```javascript
setDay(new Date(2014, 8 /* Sep */, 1), 0, {weekStartsOn: 1}) // 1 September 2014
setDay(new Date(2014, 8 /* Sep */, 1), 2, {weekStartsOn: 1}) // 3 September 2014
setDay(new Date(2014, 8 /* Sep */, 1), 0, {weekStartsOn: 2}) // 26 August 2014
```

I think it makes more sense for the `weekStartsOn` option to affect interpretation of the second argument as well as the first. My reasoning for this is as follows: if you are writing code that uses Monday as the 0th day in a week, you should still be able to use date-fns and just specify the weekStartsOn option to any relevant calls. As it stands, you would have to do modulus arithmetic on the second argument to make use of this function at all.

Once #473 is merged, this is an easy fix.

leshakoss:
Closed. See comments in #473

PR #473, opened 2017-04-07
Allow getDay to accept weekStartsOn option

gausie:
Adds an options parameter with weekStartsOn to getDay

For example

```javascript
// A Friday with standard options
getDay(new Date(2017, 3, 7)) // 5

// A Friday with the week beginning on a Monday
getDay(new Date(2017, 3, 7), { weekStartsOn: 1 }) // 4
```

Also updates other functions using weekStartsOn to rely on getDay, removing duplicated logic

NB: In working on this PR I discovered issue #472.

### DIFF
--- a/src/end_of_week/index.js
+++ b/src/end_of_week/index.js
@@ -1,3 +1,4 @@
+var getDay = require('../get_day/index.js')
 var parse = require('../parse/index.js')
 
 /**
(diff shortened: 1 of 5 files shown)


leshakoss:
@gausie so, with your fix, if you pass `{weekStartsAt: 1}` then 0 will be Monday?

Current behaviour of `date-fns` is chosen to allow something like this:
```javascript
const SUNDAY = 0
const MONDAY = 1
// etc

setDay(new Date(), SUNDAY, {weekStartsAt: MONDAY})
```

gausie:
Yes that is my intention. To be honest I hadn't considered the usage you describe. In my current project we have a number of arrays representing day info where the 0th element represents Monday's info, 1st Tuesday etc. Using the method as changed in this PR allows us to point to the correct element in that array easily.

In any case, this PR doesn't affect setDay and I think stands on its own!

leshakoss:
@gausie many people use the same `options` for all `date-fns` functions, and having functions that handle weekdays by different ways will lead to confusion. As I think, `date-fns` should always handle 0 as Sunday, 1 as Monday etc. because that's what JavaScript's `Date` object do. In other functions setting `options.weekStartsAt` doesn't change that behaviour. Sorry if that affects your workflow, I suggest you to make a separate library based on `date-fns` for your needs. Thank you for using `date-fns` anyway!

PR #487, opened 2017-05-12
Add step to eachDay function

BDav24:
Hi, do you think this option would be useful?


BDav24:
Hi @kossnocorp, may I let you merge this feature then? Given the low complexity, you will go a lot faster than I would, setting up the whole env ;)

kossnocorp:
Okay! My plan is to merge and release it to `v1` and then backport to `v2`. I'll ping once it's released.

PR #498, opened 2017-05-29
Refine test, build and release systems

kossnocorp:
- The PR refines `scripts` dir structure by grouping scripts into 3 categories: test, build and release. 
- Release system now is fully automated. From now on, to initiate a release, you need to create a GitHub release, and then Travis CI will prepare and publish it.
- Remove internal scripts from `package.json` to for readability sake. 
- Get rid of `babel-node`.
- Get rid of Bower and UMD.
- Update docs here and there.


docs/i18n.md
```diff
@@ -10,10 +10,6 @@
 
 ## Usage
 
-**Important! Note that non-CommonJS packages (Bower, CDN)
-don't support I18n yet.** Please [vote for the issue](https://github.com/date-fns/date-fns/issues/232)
-if you want to make it happen.
-
 There are just a few functions that support I18n:
 
 - [`format`](https://date-fns.org/docs/format)
```
leshakoss:
```markdown
 - [`format`](https://date-fns.org/docs/format)
 - [`formatDistance`](https://date-fns.org/docs/formatDistance)
 - [`formatDistanceStrict`](https://date-fns.org/docs/formatDistanceStrict)
 - [`formatRelative`](https://date-fns.org/docs/formatRelative)

 To use a locale, you need to import it and then pass
 as an option to a function:
 
 ```js
 import {formatDistance} from 'date-fns/esm'
 // Import Esperanto locale
 import {eo} from 'date-fns/esm/locale')
 
 var result = formatDistance(
   new Date(2016, 7, 1),
   new Date(2015, 0, 1),
   {locale: eo} // Pass the locale as an option
 )
 //=> 'pli ol 1 jaro'
```
```

File: scripts/build/fp.js [hunk 2]
```diff
@@ -1,5 +1,14 @@
...
+ * @file
+ * The script generates the FP functions using the docs JSON file.
+ *
+ * It's a part of the build process.
+ */
+
+const fs = require('fs')
+const jsDocs = require('../../docs.json')
 
 const generatedAutomaticallyMessage = '// This file is generated automatically by `scripts/buildFP.js`. Please, don\'t change it.'
```
leshakoss: // This file is generated automatically by `scripts/build/fp.js`

scripts/build/indices.js
```diff
@@ -1,9 +1,26 @@
...
+ *
+ * It's a part of the build process.
+ */
+
+const fs = require('fs')
+const path = require('path')
+const listFns = require('../_lib/listFns')
+const listFPFns = require('../_lib/listFPFns')
+
+const generatedAutomaticallyMessage = "// This file is generated automatically by `scripts/buildIndex.js`. Please, don't change it."
```
leshakoss: // This file is generated automatically by `scripts/build/indices.js`.

scripts/build/typings.js
```diff
@@ -1,6 +1,15 @@
...
+ * The script generates Flow and TypeScript typing files.
+ *
+ * It's a part of the build process.
+ */
+
+const fs = require('fs')
+const listLocales = require('../_lib/listLocales')
+const jsDocs = require('../../docs.json')
 
 const generatedAutomaticallyMessage = '// This file is generated automatically by `scripts/buildTypings.js`. Please, don\'t change it.'
```
leshakoss: ... by `scripts/build/typings.js`

scripts/release/tweet.js
```diff
@@ -19,7 +29,6 @@ fetch(zapierHookURL, {
   }),
   headers: {'Content-Type': 'application/json'}
 })
-  .then(() => console.log('+ Done!'))
```
leshakoss: (btw, does zapier return the tweet link?)

kossnocorp: @leshakoss nope, it's just a hook.

kossnocorp:
@leshakoss please take a look, I addressed your review comments.

scripts/build/docs.js
```diff
@@ -1,11 +1,21 @@
...
+ *
+ * It's a part of the build process.
+ */
+
+const fsp = require('fs-promise')
+const path = require('path')
+const cloneDeep = require('lodash.clonedeep')
+const jsDocParser = require('jsdoc-to-markdown')
+const listFns = require('../_lib/listFns')
+const docsConfig = require('../docs')
```
leshakoss: `'../../docs/index.js'`

index.js because just `'../../docs'` resolves to `'date-fns/docs.json'`.

Also in `'date-fns/docs/index.js`:

```js
const path = require('path')

module.exports = {
```
instead of
```js
import path from 'path'

export default {
```

scripts/build/build.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# The script unifies the build scripts.
+#
+# It's the entry point for the build process.
+
+set -e
+
+./scripts/build/docs.js
```
leshakoss: I'm OK with it, though I prefer when they were package.json scripts, because yarn and npm announced which script they ran — right now the progress is completely silent. Maybe log in console? Like this:
```sh
printf "\x1B[32mBuilding docs...\x1B[0m\n"
```

scripts/test/travis.sh
```diff
@@ -0,0 +1,42 @@
...
+
+function prebuild {
+  env BUILD_TESTS=true webpack --config ./config/webpack.js
+}
+
+if [ "$TEST_SUITE" == "main" ]
+then
+  yarn lint
+  flow check
+  ./scripts/test/systemJS.js
```
leshakoss: Same as with building scripts, the progress is silent. If this thing failed: https://travis-ci.org/date-fns/date-fns/jobs/238663435#L223 it would be hard to understand which part of script it is. Maybe make them package.json scripts again?

kossnocorp:
@leshakoss I didn't test it either, but I guess we'll test it really soon 😬

scripts/build/docs.js
kossnocorp: @leshakoss good catch 👍

scripts/build/build.sh
kossnocorp: @leshakoss it's a good point :+1:

Although I don't want to see it in `package.json` because that's internal scripts not intended to be used separately. Another reason is that when there are so many scripts it's hard to digitize the list which is a problem for contributors.

Luckily there is `set -x` which does exactly what we want.

PR #503, opened 2017-06-12
hungarian localization added

horvathlg:


Gustorn:
I'm looking for Hungarian support as well, a small note on the current implementation (even though it might be somewhat obsolete with v2 right around the corner): for `distanceInWordsLocale.halfAMinute` the correct spelling is `fél perce` (in two words) instead of the current `félperce`.

horvathlg:
@Gustorn you are right, I'll fix this.

kossnocorp:
Thank you, a lot!

I'll merge and release it as v1.29 soon. I'll leave a comment when it's shipped.

kossnocorp:
You won't believe, but the soon is finally here 😆 I've just backported build scripts from v2 and finally managed to release a new v1 version: https://date-fns.org/v1.29.0/docs/Change-Log#1.29.0-2017-10-11

Thank you again!

Issue #554, opened 2017-09-11
Fix danish locale support for long and relative formats


stefanbugge:
The danish (da) localization files should be updated to support long and relatives formats

leshakoss:
Closing due to the lack of activity

PR #555, opened 2017-09-11
Fixing danish locale support for long and relative formats #554

stefanbugge:
Fixing long and relative formats for the danish locale.
Additionally the weekday and month patterns have also been fixed.

All tests have been corrected as well.


kossnocorp:
The fix is released `v2.0.0-alpha.3`. See the change log: https://gist.github.com/kossnocorp/a307a464760b405bb78ef5020a4ab136#v200-alpha3

PR #558, opened 2017-09-13
Global flow typings

leshakoss:
Possible solution for https://github.com/date-fns/date-fns/issues/534#issuecomment-329200505 in #534 

Note for future: scripts/build/typings.js certainly needs some refactoring

PR #559, opened 2017-09-14
Add a test case for Flow + ESM

kossnocorp:
Right now it's failing, but if #558 works well, it will be fixed upon merge.


leshakoss:
@kossnocorp please take a look at my commits in this branch. The test was failing not because there was index.js.flow missing, but because it doesn't convert `import`s to `require`s. Now it works and even checks typings

Issue #572, opened 2017-09-27
Typings's broken on latest Typescript (2.5.3)


dashmug:
I am not sure why but something in v2.5.3 has broken compilation of my project.

`Could not find a declaration file for module 'date-fns/format'. '/path/to/my/project/node_modules/date-fns/format/index.js' implicitly has an 'any' type.`

I'm using `date-fns` like this

```js
import * as format from 'date-fns/format'
```

For now, I've reverted it back to Typescript v2.5.2 as it is working with no problems.

egorshulga:
I confirm the issue.
Date-fns version is 2.0.0-alpha.6.

dashmug:
I can also confirm that it works if we import it using

```
import { format } from 'date-fns'
```

luisbelloch:
TS 2.5.3 changes [here](https://github.com/Microsoft/TypeScript/milestone/55?closed=1). I cannot tell the change that broke the typings.

Works for me using Typescript 2.5.2 and date-fns 1.28.5

kossnocorp:
If someone could investigate the reason and help me to fix it, I will much appreciate it! ❤️  I'm not a TS user so it's hard to me dig into it.

kossnocorp:
@dashmug which version you have btw?

dashmug:
@kossnocorp 
I'm using `date-fns v1.28.5` and `typescript v2.5.3`.

This looks like a bug with Typescript. Here's what I found so far...

If I comment out [lines 1128 to 1131 of typings.d.ts](https://github.com/date-fns/date-fns/blob/v1/typings.d.ts#L1128-L1131) 

```ts
// declare module 'date-fns/format' {
//     import { format } from 'date-fns'
//     export = format
// }

```

and move it into another file in my project (say `typings/date-fns.d.ts`), 

```ts
declare module 'date-fns/format' {
    import { format } from 'date-fns'
    export = format
}
```

it just works as usual.

Could it be that Typescript does not allow multiple `declare module` statements in a single file anymore? I don't see any commit for that.

plantain-00:
According to feedback from typescript maintainers: 
>A solution that should work in both old and new versions of TypeScript would be to create a file date-fns/format/index.d.ts like is done with date-fns/format/index.js.flow, instead of adding many declare module in date-fns/typings.d.ts. Then there would be no need for adding typings to package.json since the definitions would be where we expect them to be anyway.

[...]

PR #573, opened 2017-10-03
Update typings based on new Typescript version.

dashmug:
Issue: https://github.com/date-fns/date-fns/issues/572


src/add_iso_years/index.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module 'date-fns/add_iso_years' {
```
bradleyayers: You don't need the `declare module` syntax here. You should be able to simple have:

```
import {addISOYears} from 'date-fns'
export = addISOYears
```

src/locale/ar/index.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'date-fns/locale/ar' {}
```
bradleyayers: These locale files can simply be `export {}`

src/add_iso_years/index.d.ts
dashmug: Thanks for the feedback. Do we only need the `declare module` syntax if the declaration file is not named `index.d.ts`?

scripts/generate_typings.js
```diff
@@ -149,39 +133,46 @@ function getTypeScriptFnDefinition (fn) {
...
+  const name = fn.content.name
+  const snakeCaseName = fn.file.snakeCaseName
+  const filename = `./src/${snakeCaseName}/index.d.ts`
+  const moduleName = `date-fns/${snakeCaseName}`
 
-  const definition = `declare module 'date-fns/${name}' {}`
+  const typingString = [`declare module '${moduleName}' {`]
+    .concat(`  import {${name}} from 'date-fns'`)
+    .concat(`  export = ${name}`)
+    .concat('}')
```
leshakoss: Could you please add a new line at the end of the file? Like this: `.concat('}\n')`

scripts/generate_typings.js
```diff
@@ -149,39 +133,46 @@ function getTypeScriptFnDefinition (fn) {
...
+
+  fns.forEach(generateTypeScriptFnTyping)
+}
+
+function generateTypeScriptLocaleTyping (locale) {
+  const snakeCaseName = locale.snakeCaseName
+  const name = `locale/${snakeCaseName}`
+  const filename = `./src/locale/${snakeCaseName}/index.d.ts`
+
+  const typingString = `declare module 'date-fns/${name}' {}`
```
leshakoss: Same here

package-lock.json
```diff
@@ -0,0 +1,8223 @@
+{
```
leshakoss: We use yarn for date-fns development, and the package doesn't have any non-dev dependencies, so this file is not really needed. Could you please delete it from the PR?

dashmug:
@leshakoss @bradleyayers 
I'll test this more (including [@bradleyayers's feedback above](https://github.com/date-fns/date-fns/pull/573#discussion_r142312895)) later.

Thank you for the quick feedback.

src/add_iso_years/index.d.ts
bradleyayers: It's only needed if it the declaration file is not for the adjacent JavaScript module.

dashmug:
@leshakoss I have tested this on the following typescript versions and can confirm that it's working:
* v2.4.2
* v2.5.2
* v2.5.3 (latest)

Anything else I need to do to get this ready for merging?

dashmug:
@leshakoss Wait, I take it back. I wasn't testing it right.

timwangdev:
Any updates on this? @dashmugm , Thanks for the PR btw.

ForbesLindesay:
My current workaround is a postinstall script of:

```js
const fs = require('fs');

const dirname = __dirname + '/../node_modules/date-fns';
const modules = fs
  .readdirSync(dirname)
  .filter(
    name =>
      fs.statSync(dirname + '/' + name).isDirectory() &&
      /^[a-z0-9]+(\_[a-z0-9]+)*$/.test(name),
  );

modules.forEach(name => {
  const camelName = name.replace(/_(.)/g, (_, c) => c.toUpperCase());
  fs.writeFileSync(
    dirname + '/' + name + '/index.d.ts',
    [
      'import {' + camelName + "} from 'date-fns';",
      'export = ' + camelName + ';',
    ].join('\n'),
  );
});
```

kossnocorp:
Thank you a lot! I just finished backporting build scripts from v2 and going to ship this update.

kossnocorp:
Sorry for fake news, but I had to revert the merge. See: https://github.com/date-fns/date-fns/issues/572#issuecomment-335781855

If anyone here wants to help, please chime in!

timwangdev:
@kossnocorp 
@ForbesLindesay 's postinstall script works on my production project. We could just run the script once and call it a commit.

kossnocorp:
@timwangdev this is great!

Am I properly understand that the working solution is to introduce `index.d.ts` next to every function with such content?

```js
import { functionName } from 'date-fns';
export = functionName;
```

timwangdev:
@kossnocorp Yep!

kossnocorp:
Nice! Then I'll update the scripts and release a new version.

cbengtson85:
@kossnocorp 
Just started migrating to date-fns from moment and running into this issue.  Is there an ETA when the new version will be released?  Thanks!

ForbesLindesay:
@kossnocorp  any chance you could push a release soon?

Issue #276, opened 2016-12-05
😡 Broken deployment: Travis CI didn't upload new versions to CDN


kossnocorp:
See: http://cdn.date-fns.org/v1.12.1/date_fns.min.js

umidbekkarimov:
Did you try https://unpkg.com ? just add `dist` dir to npm and it would be accessible from https://unpkg.com/date-fns@1.12.1/dist/date_fns.js

kossnocorp:
I've decided to drop support of UMD/CDN versions as our build system is complex enough already and I don't want to spend time on features that nobody use. For long time CDN was broken and not a single person complained.

For those who're dying to have CDN version of date-fns and Moment.js is not an option, please contact me.

PR #586, opened 2017-10-19
Update docs to replace date-fns' CDN with cdnjs

ClementParis016:
While building a quick prototype using date-fns, I found that the CDN doesn't contain all the versions of the library, and reading #276 confirmed it.

So this is just a small docs update to replace date-fns own CDN with cdnjs, which itself provides all the versions. This way newcomers won't wonder what they did wrong when they are trying to load a version from the CDN which is not available.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -38,7 +38,7 @@ dates.sort(dateFns.compareAsc)
 ## Installation
 
 The library is available as an [npm package](https://www.npmjs.com/package/date-fns),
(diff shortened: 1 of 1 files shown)


ianmcnally:
I hit the outdate documentation today, when wanting to create a Codepen with date-fns. Cdnjs worked like a charm, and got me back on the project I was working on.

kossnocorp:
Better late than never 🤪

PR #590, opened 2017-10-20
Add Hebrew locale

Nirlah:


src/locale/he/_lib/formatDistance/index.js
```diff
@@ -0,0 +1,91 @@
+var formatDistanceLocale = {
+  lessThanXSeconds: {
+    one: 'פחות משניה',
+    other: 'פחות מ־{{count}} שניות'
+  },
+
+  xSeconds: {
+    one: 'שניה 1',
```
TxHawks: I don't think the `1` prefix and suffix is needed in all the `one` properties in Hebrew, `כמעט שנה` is cleaner and more idiomatic to the language than `כמעט שנה 1`

Nirlah:
Hey, will update this weekend.

Nirlah:
Hey @amitzw,
Looked into it today, it seems there is quite a few breaking changes to i18n system.
Won't have time for it in the upcoming weeks, so feel free to take over.

Issue #592, opened 2017-10-22
Tests just hang and never quit


jish:
When running the test suite using `yarn run test` Chrome opens, the test suite runs, and then the test suite and Chrome never close.

jish:
Upon further inspection this looks like a development test server, rather than a one-off test runner. Maybe we could document the difference between the two and how and when to use each one.

PR #593, opened 2017-10-22
Document how to run the test suite

jish:
The contributing guide should explain how to run the test suite.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,7 +60,7 @@ Please follow the main contributing rules, to maintain date-fns' top quality:
 
 - Write tests.
 
-- Run tests: `$ yarn test -- --single-run`.
+- Run tests: `$ yarn test --single-run`.
 
 - [Write documentation](#documentation).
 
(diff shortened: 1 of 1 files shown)


abritinthebay:
It would be `npm test` surely? No dependency on Yarn and npm is installed with Node. Why add documentation for an inferior and additional dependency? Seems strange.

jish:
I have updated the documentation to use `npm test` instead.

jish:
According to https://github.com/date-fns/date-fns/issues/592 and https://github.com/date-fns/date-fns/issues/614 `yarn test -- --single-run` should be used to run tests. I have updated this branch to reflect that. 🙂

CONTRIBUTING.md
```diff
@@ -60,6 +60,8 @@ Please follow the main contributing rules, to maintain date-fns' top quality:
 
 - Write tests.
 
+- Run tests: `$ yarn test -- --single-run`.
```
tusbar: Since Yarn 1.0, `--` is not required anymore to forward options.

```
yarn test --single-run
```

jish:
I have removed the extraneous `--`

PR #595, opened 2017-10-26
small fix in Czech locale

SilenY:


leshakoss:
@SilenY thanks for your contribution!

PR #598, opened 2017-11-01
Fix types break on typescript 2.5.3 for v1 branch

plantain-00:
#572


src/add_days/index.d.ts
```diff
@@ -0,0 +1,6 @@
+
```
leshakoss: Please remove these two newlines

plantain-00:
I removed the empty lines.

PR #600, opened 2017-11-05
Adding isFirstWeekInMonth helper

KuznetsovIlya:
Use-case: diagram gantt or another timescale in week-scale, when first week-item in month needs special style or logic.

### DIFF
--- a/src/isFirstWeekInMonth/index.js
+++ b/src/isFirstWeekInMonth/index.js
@@ -1,0 +1,32 @@
+import toDate from '../toDate/index.js'
+import startOfWeek from '../startOfWeek/index.js'
+import addDays from '../addDays/index.js'
+
+/**
+ * @name isFirstWeekInMonth
+ * @category Week Helpers
+ * @summary Is the given date the first week of a month?
+ *
+ * @description
+ * Is the given date the first week of a month?
+ *
+ * @param {Date|String|Number} date - the date to check
+ * @param {Options} [options] - the object with options. See [Options]{@link https://date-fns.org/docs/Options}
+ * @param {0|1|2} [options.additionalDigits=2] - passed to `toDate`. See [toDate]{@link https://date-fns.org/docs/toDate}
+ * @returns {Boolean} the date is the first week of a month
+ * @throws {TypeError} 1 argument required
+ * @throws {RangeError} `options.additionalDigits` must be 0, 1 or 2
+ *
+ * @example
+ * // Is 1 September 2017 the first week of a month?
+ * var result = isFirstWeekInMonth(new Date(2017, 8, 1))
+ * //=> true
+ */
+export default function isFirstWeekInMonth (dirtyDate, dirtyOptions) {
+  if (arguments.length < 1) {
+    throw new TypeError('1 argument required, but only ' + arguments.length + ' present')
+  }
+
+  const date = toDate(dirtyDate, dirtyOptions)
+  return addDays(startOfWeek(date), 6).getDate() <= 7
+}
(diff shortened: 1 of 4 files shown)


src/isFirstWeekInMonth/index.js
```diff
@@ -0,0 +1,32 @@
...
+ * //=> true
+ */
+export default function isFirstWeekInMonth (dirtyDate, dirtyOptions) {
+  if (arguments.length < 1) {
+    throw new TypeError('1 argument required, but only ' + arguments.length + ' present')
+  }
+
+  const date = toDate(dirtyDate, dirtyOptions)
+  return addDays(startOfWeek(date), 6).getDate() <= 7
+}
```
leshakoss: So this is how I to do it correctly:
```js
return addDays(startOfWeek(dirtyDate, dirtyOptions), 6, dirtyOptions).getDate() <= 7
```
I did not check the correctness of the algorithm. Could you please confirm that is works for edge cases? (like the date is first day of month; the date is last day of the first week; same for `weekStartsOn` set on something that isn't 0)

src/isFirstWeekInMonth/test.js
```diff
@@ -0,0 +1,49 @@
+// @flow
+/* eslint-env mocha */
+
+import assert from 'power-assert'
+import isFirstWeekOfMonth from '.'
```
leshakoss: I see that you renamed the function from `isFirstWeekOfMonth`. I like this name (with `of`) more as it uses the same logic as other functions with `of` or `in` in the name. Compare `getDayOfYear` with `getDaysInMonth`

src/isFirstWeekInMonth/test.js
```diff
@@ -0,0 +1,49 @@
...
+  it('returns true if the given date is in the first week of month - 2', function () {
+    var result = isFirstWeekOfMonth(new Date(2017, 9 /* Oct */, 30))
+    assert(result === true)
+  })
+
+  it('returns false if the given date is not in the first week of month', function () {
+    var result = isFirstWeekOfMonth(new Date(2017, 10 /* Nov */, 6))
+    assert(result === false)
+  })
+
```
leshakoss: Please add `weekStartsOn`-specific tests:
https://github.com/date-fns/date-fns/blob/master/src/startOfWeek/test.js#L20-L54
https://github.com/date-fns/date-fns/blob/master/src/startOfWeek/test.js#L111-L115

src/isFirstWeekInMonth/test.js
```diff
@@ -0,0 +1,49 @@
...
+  it('throws `RangeError` if `options.additionalDigits` is not convertable to 0, 1, 2 or undefined', function () {
+    // $ExpectedMistake
+    var block = isFirstWeekOfMonth.bind(null, new Date(2017, 9 /* Oct */, 30), {additionalDigits: NaN})
+    assert.throws(block, RangeError)
+  })
+
+  it('throws TypeError exception if passed less than 1 argument', function () {
+    assert.throws(isFirstWeekOfMonth.bind(null), TypeError)
+  })
+})
```
leshakoss: All stuff I mentioned also applies to `getWeekOfMonth`: https://github.com/date-fns/date-fns/issues/601#issuecomment-341995752
I leave this review (and PR) for the case if we and the date-fns community will choose to add `isFirstWeekOfMonth` function instead of implementing `getWeekOfMonth`

src/isFirstWeekInMonth/index.js
leshakoss: Also edge cases to consider:
- the first week is a single day
- the first day of the second week

leshakoss: - the last day of the previous month (e.g. I expect that this function would result in `false` for 31 October 2017)

leshakoss:
Closing this due to it being replaced with `getWeekOfMonth`. See #621 and discussion in #601

PR #616, opened 2017-11-17
Selectable default locale v1

dignite:
# Why?

Users that don't need to format dates with English locale might not want to include the english locale in their javascript bundle to optimize performance towards the end user. The size of the English locale is around 2 KB.

- By hiding English locale behind `date-fns/src/locale/default_locale` the module redirect configuration will make more sense and won't redirect the cases where you want to explicitly user English locale.
- The same technique could be used to select a sensible default locale in a project, although that is probably not recommended.

# How?

- Introduce and use the module `locale/default_locale`
- Add documentation on how to exclude the English locale by showing an example Webpack configuration

### DIFF
--- a/docs/i18n.md
+++ b/docs/i18n.md
@@ -21,6 +21,8 @@ There are just a few functions that support I18n:
 - [`distanceInWordsToNow`](https://date-fns.org/docs/distanceInWordsToNow)
 - [`distanceInWordsStrict`](https://date-fns.org/docs/distanceInWordsStrict)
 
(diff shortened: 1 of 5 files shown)


dignite:
An idea for v2 could be to add the indirection of the defaultLocale module but instead of adding the module replacement technique to the documentation (the one with a disclaimer) we could add versions of the functions that use defaultLocale that explicitly need a locale sent in.

For example `format` could have an alternative `formatWithoutDefaultLocale` and to avoid code duplication `format` could call `formatWithoutDefaultLocale` where the bulk of the logic would be.

dignite:
One year later, this is never going to land, closing 😢

PR #617, opened 2017-11-17
Feature/continuous testing

SpiderUnicorn:
Added support for running tests in node with jest.

This is preferable during development due to a much improved speed when running the tests. It also allows for watching for changes and only running tests related to a change. This means you get feedback of a failing test within seconds.

The configuration is minimal. Tests can be run in watch mode by running:

jest --watch


kossnocorp:
That is a good idea, sorry that it took so long to approach it. We definitely need an ability to run tests in Node.

At the moment we have plenty of failing tests and I going to investigate those in near future.

kossnocorp:
I continued the work in #1192

Issue #540, opened 2017-08-15
Support eachYear eachMonth eachWeek eachHour eachMinute


BiteBit:
Hi, team.
**eachDay** is a very convenient function under some situation. but it's not comprehensive, i think **eachYear**,**eachMonth**,**eachWeek**,**eachHour** and **eachMinute** will be very useful too.
Is there any plan for this?
Thanks.

brigand:
@BiteBit could you write a more complete proposal with example inputs and outputs?

kossnocorp:
At the _moment_ the date-fns team has no throughput for that, but PR(s) is welcome!

maraisr:
I'd be happy to take this on - but need some context?

Did you want something like a `eachDay` function that returns.... what?

cc @brigand  @kossnocorp

strarsis:
Any changes for v2?
I found an implementation in date-fns codebase, but apparently it isn't available anymore in v.2:
https://github.com/date-fns/date-fns/blob/17ba85a89843f8914f8d33cb85ab0a23c1db5a8d/src/eachMonthOfInterval/index.js

EvanHahn:
I think all of these functions exist now. Should this issue be closed?

PR #618, opened 2017-11-17
Added eachYear, eachMonth, eachWeek #540

mborgbrant:
Implementented:

*eachYearInterval
*eachMonthInterval
*eachWeekInterval

from ticket #540


mcmxc:
Why would these methods be unavailable to import? I'm on v2.8.1

PR #625, opened 2017-11-22
UTC functions

leshakoss:
For this PR:
- [x] Add tests for UTC functions
- [x] Add UTC functions
- [ ] Make `format` timezones always output UTC timezone (and add private option `options._timezone` to modify this behaviour)
- [ ] Add indices
- [ ] Add typings
- [ ] Add FP tests
- [ ] Add FP fns
- [ ] Add documentation

In future:
- [ ] Add to-UTC and from-UTC helper functions to src/_lib
- [ ] Convert all non-UTC functions to UTC wrappers

### DIFF
--- a/src/utc/addDays/index.js
+++ b/src/utc/addDays/index.js
@@ -1,0 +1,33 @@
+import toDate from '../toDate/index.js'
+
+/**
+ * @name addDays
+ * @category Day Helpers
+ * @summary Add the specified number of days to the given date.
+ *
+ * @description
+ * Add the specified number of days to the given date.
+ *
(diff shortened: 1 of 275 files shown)


hlehmann:
@leshakoss can it be based on #707 ?

leshakoss:
@hlehmann I dunno, I didn't look at it. Right now releasing the v2 is much much bigger priority than TZ support

hlehmann:
Yes, it's just that I mainly use UTC dates and I need it to make a proper switch. Also with esnext features such as pipeline operator date-fns will have a huge benefit over moment/luxon.

Good efforts have been made in #707.

PR #634, opened 2017-11-27
Update long & relative formats for 'nl' locale

bitcrumb:
We noticed that the `nl` locale did not yet contain the proper long format specifiers (see: https://en.wikipedia.org/wiki/Date_and_time_notation_in_the_Netherlands). Also, the relative date formatting did not use the Dutch language yet.

In this pull request I've updated both. Since this is my first contribution to this project, I'm happy to receive feedback and update the PR accordingly if that would be necessary.

### DIFF
--- a/src/locale/nl/_lib/formatLong/index.js
+++ b/src/locale/nl/_lib/formatLong/index.js
@@ -1,12 +1,12 @@
 import buildFormatLongFn from '../../../_lib/buildFormatLongFn/index.js'
 
 var formatLong = buildFormatLongFn({
-  LT: 'h:mm aa',
-  LTS: 'h:mm:ss aa',
-  L: 'MM/DD/YYYY',
-  LL: 'MMMM D YYYY',
-  LLL: 'MMMM D YYYY h:mm aa',
-  LLLL: 'dddd, MMMM D YYYY h:mm aa'
+  LT: 'HH:mm',
+  LTS: 'HH:mm:ss',
+  L: 'DD/MM/YYYY',
+  LL: 'D MMMM YYYY',
+  LLL: 'D MMMM YYYY HH:mm',
+  LLLL: 'dddd, D MMMM YYYY HH:mm'
 })
 
 export default formatLong
--- a/src/locale/nl/test.js
+++ b/src/locale/nl/test.js
@@ -131,32 +131,32 @@ describe('nl locale', function () {
     describe('long formats', function () {
       it('LT', function () {
         var result = format(date, 'LT', {locale: locale})
-        assert(result === '10:32 a.m.')
+        assert(result === '10:32')
       })
 
       it('LTS', function () {
         var result = format(date, 'LTS', {locale: locale})
-        assert(result === '10:32:00 a.m.')
+        assert(result === '10:32:00')
       })
 
       it('L', function () {
         var result = format(new Date(2017, 6 /* Jul */, 2), 'L', {locale: locale})
-        assert(result === '07/02/2017')
+        assert(result === '02/07/2017')
       })
 
       it('LL', function () {
         var result = format(date, 'LL', {locale: locale})
-        assert(result === 'april 4 1986')
+        assert(result === '4 april 1986')
       })
 
       it('LLL', function () {
         var result = format(date, 'LLL', {locale: locale})
-        assert(result === 'april 4 1986 10:32 a.m.')
+        assert(result === '4 april 1986 10:32')
       })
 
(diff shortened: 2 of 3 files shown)


src/locale/nl/_lib/formatLong/index.js
```diff
@@ -1,12 +1,12 @@
...
 var formatLong = buildFormatLongFn({
-  LT: 'h:mm aa',
-  LTS: 'h:mm:ss aa',
-  L: 'MM/DD/YYYY',
-  LL: 'MMMM D YYYY',
-  LLL: 'MMMM D YYYY h:mm aa',
-  LLLL: 'dddd, MMMM D YYYY h:mm aa'
+  LT: 'HH:mm',
+  LTS: 'HH:mm:ss',
+  L: 'DD/MM/YYYY',
```
marlonbaeten: I believe the official Dutch formatting is DD-MM-YYYY

src/locale/nl/test.js
```diff
@@ -228,32 +228,32 @@ describe('nl locale', function () {
...
     it('next week', function () {
       var result = formatRelative(new Date(1986, 3 /* Apr */, 6, 12, 0), baseDate, {locale: locale})
-      assert(result === 'zondag at 12:00 p.m.')
+      assert(result === 'zondag om 12:00')
     })
 
     it('after the next week', function () {
       var result = formatRelative(new Date(1986, 3 /* Apr */, 11, 16, 50), baseDate, {locale: locale})
-      assert(result === '04/11/1986')
+      assert(result === '11/04/1986')
```
marlonbaeten: This should be 11-04-1986

src/locale/nl/_lib/formatLong/index.js
bitcrumb: When I set my region to _Belgium_ on my Mac, the date format uses slashes. However, when I set it to _Netherlands_ the date is displayed using dashes. See:
screen shot 2017-12-18 at 21 18 46
screen shot 2017-12-18 at 21 18 53

I guess that means that we should actually go one step further and have two folders `src/locale/nl-BE` and `src/locale/nl-NL` instead of only one `src/locale/nl`.

Would you agree?

marlonbaeten: I agree. But I think that would be a breaking change, or can we make an alias of `src/locale/nl` to `src/locale/nl-NL`?

marlonbaeten: Maybe having `src/locale/nl` and `src/locale/nl-BE` would be more conform the existing conventions, see `src/locale/fr` and ``src/locale/fr-CH` for instance.

leshakoss: Breaking changes are ok while date-fns is in alpha. There will be plenty of them yet

I approve `nl` and `nl-BE`.

[Guide for locale names](https://github.com/date-fns/date-fns/blob/master/docs/i18nContributionGuide.md#choosing-a-directory-name-for-a-locale)
[Guide for different locales with the same language](https://github.com/date-fns/date-fns/blob/master/docs/i18nContributionGuide.md#creating-a-locale-with-the-same-language-as-another-locale)

bitcrumb: Thx @leshakoss!
I will update this PR accordingly :)

bitcrumb:
@marlonbaeten @leshakoss Sorry that it took so long to continue with this pull request. Please do verify. I took the approach as suggested and added an extra locale `nl-BE` which mainly imports from `nl` except for the long format.

I also executed the build script & updated some documentation. Any feedback is welcome.

marlonbaeten:
@leshakoss could you merge these additions? Thanks in advance!

MBing:
Thank you for your contributions Lode. 

For country/region reference I wouldn't rely on Apple. In Belgium education and industry has clear goals when writing letters to do so according to the ISO 31-0 and also the BIN norms NBN Z 01-002.

A date is written like `YYYY-MM-DD` and numbers are grouped per 3 with spaces iso a `.` as a separator. The BIN norms are a formal industry standard and should be our aim when trying to be consistent as an implementation for the majority. 

References:
[VVKSO BIN norms](http://ond.vvkso-ict.com/vvksomain/brochure/Vouwblad%20BIN-normen%20Oranje.pdf)
[ISO 31-0 & Belgische BIN normen NBN Z 01-002](http://taaladvies.net/taal/advies/vraag/1/10000000_10_000_000/)

PR #641, opened 2017-12-07
Fix short version of Norwegian day names

kbrabrand:
In the current Norwegian locale, the short versions of the day names are the same as the narrow. The three day versions are commonly used in Norway, and having the correct ones in `date-fns` would be great.

This PR adds that.


kbrabrand:
@leshakoss Any chance to get this one in? Small change and we have a customer waiting 🙏

PR #650, opened 2017-12-25
Fix Croatian tests

manico:
Enabled and fixed tests for `hr` locale.


manico:
Hi, will this be merged?
I would create serbian and bosnian translations after this.

PR #652, opened 2017-12-26
Fixing Ukrainian not translated part and rename locale to UK

korzhyk:


src/locale/uk/index.js
```diff
@@ -11,7 +11,7 @@ import buildTokensRegExp from '../_lib/buildTokensRegExp/index.js'
  * @category Locales
  * @summary Ukrainian locale.
  * @language Ukrainian
- * @iso-639-2 ukr
+ * @iso-639-1 uk
```
leshakoss: @korzhyk we need @iso-639-2 tag for generating documentation for date-fns.org, please return it back

leshakoss:
Thanks a lot for your contribution! I will try to release this as a part of v2.0.0-beta.1 as soon as possible (hopefully a few days). I will let you know when it’s released.

Issue #660, opened 2018-01-04
Finalize API for v2.0.0-beta.1


leshakoss:
Hi there, date-fns community.

As you probably already know, we've been promising to release v2 for about a year now. Now it is time to set a completely arbitrary deadline to make final changes, which is February the 1st 2018. By then I hope to set a stable API, finish the code and release the beta. After that, date-fns will suffer no breaking changes until v3.

In this issue, I ask all of you to comment on which features are you missing in the [current v2 API](https://date-fns.org/v2.0.0-alpha.7/docs/Getting-Started) that couldn't be added in an idiomatic way without making a breaking change.

The list of breaking changes to be made I've got so far:
- Rename `ua` locale to `uk`: #651
- Use [unicode date format skeletons](http://www.unicode.org/reports/tr35/tr35-31/tr35-dates.html#Date_Format_Patterns) for `format` and `parse`: #520
  - Also it will probably involve either moving [long formats](https://github.com/date-fns/date-fns/blob/master/src/format/index.js#L66) from `format` to a separate function, or rethinking description of the long formats in the locales. 
- (optional) Rethink [`options.unit`](https://github.com/date-fns/date-fns/blob/master/src/formatDistanceStrict/index.js#L35) enum in `formatDistanceStrict` (i.e. `'year'` instead of `'Y'`)

Update: all the planned stuff is done. Beta coming soon!

leshakoss:
Also since as of now I am working on date-fns alone, I will update this comment with my progress so far, so you will know whether the development is advancing or is stuck again.

leshakoss:
I need a few more weeks. New deadline is 2018-02-12

benmccann:
I'd love to get timezone support in. I've sent https://github.com/date-fns/date-fns/pull/676 as a proof of concept. As of right now, I believe it's backwards compatible, so wouldn't necessarily have to make it in before API freeze, but I'd love to get your feedback on it to confirm approach. In the case that you would like to see a different method of doing it that alters the API and are able to review soon it might give me time to address the design before the deadline.

leshakoss:
@benmccann timezones probably wouldn't be a breaking change, but if they would, releasing them as v3 sooner that we would expect is ok for me.

As for API that you suggested, I will take a closer look at it some time later, but I would probably vote against using Intl and would instead suggest some form of compatible with Intl cross-platform API (we aim to support browsers IE8+)

benmccann:
[...]

Issue #180, opened 2016-01-01
IANA support


kossnocorp:
- IANA time zone identifiers support.
- Time zone manipulations.

Please left a comment if you are interesting in the feature.

havenchyk:
👍

olavk:
Definitely interested. Doing a major overhaul on our project right now, and would like to switch out moment.js but it seems for now we're stuck with it

vbilici:
Our website heavily relies on time zone, this feature would be great!

kossnocorp:
It will be awesome if everyone who is interested in the functionality will post use-cases so that we can see what users need, prioritize and release desired functionality earlier. Thanks 🎉

kossnocorp:
For instance I would love to know the most wanted functions.

reywright:
```
export function isoDateConversion (isoDate){
    if (isoDate){
        var d = isoDate.split(/[-T:]/);
        return new Date(d[0], d[1] - 1, d[2], d[3], d[4]);
    } else {
        return isoDate
    }
}
```

When I'm using this library I have to end up using this function a lot to normalize the dates I get back across different user timezones. It would be nice if I could (doesn't matter how) use the utility in a way where it gives me back what I want without accounting for the users timezone. That's my personal use case and how I handle it now.

kedoska:
In my case I need to set up the end-user time zone (starting with a _default_).

Application starts setting up (for example) `moment.tz.setDefault("Europe/London")` and customer can change it in his/her _settings page_.

So abstracting it will be:
 * a list that represents all the available _tz_.
 * a function to set/get the current _tz_.
 * a `format` function considering the current _tz_

/cc @kossnocorp

kucherenkovova:
+1

kossnocorp:
I started working on it: https://github.com/date-fns/tzdata-fns

The goal is to parse IANA source files and generate code that will help to convert time between (and within) timezones without glitches.

The main problem is that JavaScript doesn't have API to work with timezones. Traditionally precise timezone calculations are done on the server-side that have access to [(traffic-heavy) IANA DB](http://www.iana.org/time-zones).

For example to substract years in just `Europe/Moscow` you need to consider all changes are made to the timezone of the region. 

See history of changes for a single time zone:

```
⟩ tzdb/zdump -v Europe/Moscow
Europe/Moscow  -9223372036854775808 = NULL
Europe/Moscow  -9223372036854689408 = NULL
[...]

PR #676, opened 2018-01-28
IANA timezone support

benmccann:
This uses the [DateTimeFormat API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat#Browser_compatibility). IE users only need the [Intl.DateTimeFormat timezone polyfill](https://github.com/yahoo/date-time-format-timezone). Edge users do not need a polyfill

### DIFF
--- a/src/toDate/index.js
+++ b/src/toDate/index.js
@@ -64,6 +64,7 @@ var patterns = {
  * @param {*} argument - the value to convert
  * @param {Options} [options] - the object with options. See [Options]{@link https://date-fns.org/docs/Options}
  * @param {0|1|2} [options.additionalDigits=2] - the additional number of digits in the extended year format
+ * @param {string} [options.timeZone] - the time zone to use. Default is local time. Valid values are 'UTC' or Area/Location such as 'America/Cancun'.
  * @returns {Date} the parsed date in the local time zone
  * @throws {TypeError} 1 argument required
  * @throws {RangeError} `options.additionalDigits` must be 0, 1 or 2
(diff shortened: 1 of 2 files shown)


src/toDate/index.js
```diff
@@ -64,6 +64,7 @@ var patterns = {
  * @param {*} argument - the value to convert
  * @param {Options} [options] - the object with options. See [Options]{@link https://date-fns.org/docs/Options}
  * @param {0|1|2} [options.additionalDigits=2] - the additional number of digits in the extended year format
+ * @param {string} [options.tz] - the time zone to use. Default is local time. Valid options are 'UTC' or Area/Location such as 'America/Cancun'.
```
marnusw: The `tz` option is nice and short, but I wonder if it's not worthwhile, just in the interest of consistency across different APIs, to name this option `timeZone` as used in the `Intl` API itself?

benmccann: I debated that myself. Given that both you and @leshakoss have indicated a preference for it, I've updated it accordingly

marnusw:
I have written some documentation for helper functions that might be built around this idea to solve the issues presented in #180 and will create a separate PR with those when they are ready and depending on further discussion in #660. For now any feedback on the [documentation](https://github.com/marnusw/date-fns/blob/timezones/docs/timeZones.md) would be most welcome.

benmccann:
I'll leave this open for reference for the time being, but @marnusw has taken this PR even further, so probably best that we focus on that one in terms of review and being merged

kossnocorp:
Wow, this is a really elegant solution, love it! Thank you a lot! We'll evaluate this as soon as we finish with v2. Sorry for waiting.

Restuta:
@kossnocorp I wonder if there any news on this? Any way we can help?

palpich:
+1 for this feature

src/toDate/index.js
```diff
@@ -330,3 +338,63 @@ function dayOfISOWeekYear (isoWeekYear, week, day) {
   date.setUTCDate(date.getUTCDate() + diff)
   return date
 }
+
+// The code below is borrowed from Luxon under the MIT license
```
david-induro, line 342: "borrowed", in perpetuity

PR #677, opened 2018-01-30
correct "sideEffect" field in package.json

TxHawks:
Based on https://github.com/webpack/webpack/tree/next/examples/side-effects
the field name is `sideEffects`, not `side-effects`.


jaylinski:
Duplicate of date-fns/date-fns/pull/711.

PR #682, opened 2018-02-09
Added isLegitimate

mrenty:
Added a function to check if the passed arguments convert to a date which exists.

`isLegitimate(2018, 0, 31)` returns `true`, while `isLegitimate(2018, 1, 31)` returns `false`.


leshakoss:
@mrenty we actually had a function like that but it was scrapped because it caused some special cases with FP fns and generally made a lot of trouble. Let me check if you could find it

leshakoss:
@mrenty here it is: https://github.com/date-fns/date-fns/pull/195

leshakoss:
Anyway, the reasons we removed it are still valid, so yeah. I'll take a look at it next week and think about it, but we'll probably close this PR

mrenty:
@leshakoss alright no problem, let me know if you have any other remarks and/or questions 👍

Issue #510, opened 2017-06-21
Parse() not working in IE/Firefox consistantly


kingsleymasters:
This is against v1.28.5

Noticed an issue when Parsing ISO style dates (but not valid ISO format) with a trailing Z 

See the fiddle below (in IE11 or Firefox) it works in Chrome.
https://jsfiddle.net/p56cfcjd/4/

Parsess correctly in all browsers
`dateFns.parse('2012-01');`

Fails in IE11/Firefox but parses in Chrome
` dateFns.parse('2012-01Z');`

SLOBYYYY:
Guys, this is a big problem and more than 5 months have passed since. Any news or workaround regarding this?

kingsleymasters:
Unfortunately had to go back to using moment.js because of this bug :(  I couldn't see an easy fix.

leshakoss:
Fixed in master branch by #690

PR #690, opened 2018-02-16
Fixed issue when parsing ISO style dates (but not valid ISO format) with a trailing Z (Issue #510)

dorakung:
This PR fixed the issue [https://github.com/date-fns/date-fns/issues/510](url)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,7 @@ for the list of changes made since `v2.0.0-alpha.1`.
 - [Support for long and relative formats for Swedish locale](https://github.com/date-fns/date-fns/pull/570) (thanks to [@alexandernanberg](https://github.com/alexandernanberg)).
 
 ### Changed
+- **BREAKING**: fixed when parsing ISO style dates (but not valid ISO format) with a trailing Z (e.g 2012-01Z), it returns Invalid Date for FireFox/IE11 (Issue #510)
 
 - **BREAKING**: function submodules now use camelCase naming schema:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,7 +96,7 @@ for the list of changes made since `v2.0.0-alpha.1`.
 - [Support for long and relative formats for Swedish locale](https://github.com/date-fns/date-fns/pull/570) (thanks to [@alexandernanberg](https://github.com/alexandernanberg)).
 
 ### Changed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -96,6 +96,7 @@ for the list of changes made since `v2.0.0-alpha.1`.
 - [Support for long and relative formats for Swedish locale](https://github.com/date-fns/date-fns/pull/570) (thanks to [@alexandernanberg](https://github.com/alexandernanberg)).
 
 ### Changed
+- **BREAKING**: fixed when parsing ISO style dates (but not valid ISO format) with a trailing Z (e.g 2012-01Z), it returns Invalid Date for FireFox/IE11 (Issue #510)
```
leshakoss: It's not breaking, it's a bug fix

leshakoss: Also please convert issue number to a link (i.e. `[#510](https://github.com/date-fns/date-fns/issue/510)`)

src/toDate/test.js
```diff
@@ -219,6 +219,12 @@ describe('toDate', function () {
           assert.deepEqual(result, new Date('2014-10-25T13:46:20+07:00'))
         })
       })
+      context('when the year and the month are specified', function () {
+        it('sets timezone correctly on yyyy-MMZ format', function () {
+          var result = toDate('2012-01Z')
+          assert.deepEqual(result, new Date('2011-12-31T18:00:00+07:00'))
```
leshakoss: @dorakung please make it `'2012-01-01T00:00:00+00:00'`. It is the same thing, but shows the intention clearer

richie50:
image



Results =>
image

parse routine does not work on the latest firefox. Thank @leshakoss

leshakoss:
@richie50 I am not sure what you're trying to do. Could you please elaborate? (I don't see any connection between your two screenshots) And also could you please create a new issue for that?

richie50:
@leshakoss for some reason on firefox the dateString Object with the field _date: "07-May-2018"_ can not be converted properly to a date object by the routine **_parseDate(arg1, arg2)._** . **(Refer to the above code)**.I did trace the code and i noticed on firefox the new Date("07-May-2018") _returns null_ but this does not happen on chrome. For now i'm going to change the  "07-May-2018" to an acceptable format that can be parsed by the Javascript Date Object. Thanks.

leshakoss:
@richie50 if the string passed to `parse` fn is not in the [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) format, then it falls back to `new Date` constructor, which could work differently in different browsers

richie50:
@leshakoss thanks. 👍

PR #691, opened 2018-02-16
Long formats

leshakoss:
```js
var result = format(new Date(), 'PPPPpppp')
//=> 'Friday, February 16th, 2018 at 7:52:21 PM GMT+01:00'
```


ralyodio:
this doesn't work:

`dateFns.format(new Date(), 'PPPPpppp')`

Issue #692, opened 2018-02-16
differenceInYears() returns -0 when comparing the same date


hanshenrik:
```
differenceInYears('2018-02-01',` '2018-02-01')
// -0

differenceInYears(new Date(), new Date())
// -0

differenceInYears('2018-02-02', '2018-02-01')
// 0
```

Shouldn't the same year always be returned as a "positive" zero value?

Solvable by using `Math.abs(value)` if the value is 0. Perhaps a bigger discussion if _all_ values should be returned as positive numbers.

ghostd:
I think we should have a positve 0 (as proposed by the PR of armenavetisyan), but keep the sign for the non-zero results. thi can be used to know if we are before/after an other date.

webpapaya:
I think this is not only an issue with differenceInYears but some other differenceIn* functions as well.

````js
{
  differenceInCalendarDays: 0,
  differenceInCalendarISOWeeks: 0,
  differenceInCalendarISOYears: 0,
  differenceInCalendarMonths: 0,
  differenceInCalendarQuarters: 0,
  differenceInCalendarWeeks: 0,
  differenceInCalendarYears: 0,
  differenceInDays: -0,
  differenceInHours: 0,
  differenceInISOYears: -0,
  differenceInMilliseconds: 0,
  differenceInMinutes: 0,
  differenceInMonths: -0,
  differenceInQuarters: -0,
  differenceInSeconds: 0,
  differenceInWeeks: -0,
  differenceInYears: -0,
}
````
I think -0 should never be returned by any of the functions. It would be nice if @armenavetisyan could also take the other functions in his PR into account. Thanks.

**Comment by kossnocorp** [comment]
Hm, on one hand, that's an annoying inconsistency, but on another hand -0 and 0 for JS are the same thing:

```js
0 == -0
//=> true

0 === -0
//=> true

JSON.stringify(-0)
//=> "0"
```

I think that it doesn't make sense to add `abs` everywhere just for sake of it unless there are real concerns.

@leshakoss WDYT?

kossnocorp:
@leshakoss figured out that the problem is in `Math.round`. I still not sure how this is a problem, but since it could be fixed using simple `|| 0` and @leshakoss is so into it, here is my 👍 for making it consistent across the code base.

leshakoss:
It was not `Math.round`, I just tried to guess without looking in the code 😅

kossnocorp:
@leshakoss nice try

[...]

PR #697, opened 2018-02-25
[#692] Prevent negative zero return for the same years

armenavetisyan:
#692


kossnocorp:
Thank you for the attempt. The original issue is solved by #739

PR #707, opened 2018-03-10
Time zone helpers including IANA time zone support

marnusw:
I've built on the addition in #676 by @benmccann and tried to figure out what an API / helper functions for time zone support might look like which would solve the use cases presented in #180. Key additions are the two time zone helper functions:

- `setTimeZone` for getting a `Date` instance with the time at a given time zone
- `dateAtTimeZone` for getting a `Date` with the UTC value of a time in a given time zone

The `toDate`, `format` and `parse` functions also have additional time zone support added.

For a more complete discussion of the new functionality see the [Time zone documentation](https://github.com/marnusw/date-fns/blob/timezones/docs/timeZones.md) included in this PR.

The `date-fns` library aims to support IE8+ (see #660). This PR proposes using the `Intl` API for IANA time zone support which requires a [polyfill](https://github.com/yahoo/date-time-format-timezone) in older browsers. As a compromise, all new time zone functionality fully supports the use of time zone offset values (`+0200`, `-04:00` etc.) which will work back to IE8. If you wish to use IANA time zones a polyfill will be needed though.


src/_lib/tzMakeDateTimeFormat/index.js
```diff
@@ -0,0 +1,27 @@
+const dtfCache = {}
+
+export default function tzMakeDateTimeFormat (timeZone, locale) {
```
benmccann: You shouldn't need to pass in a locale here. This code examines the date to calculate the timezone offset, which is not language-dependent. The fact that this code uses `'en-US'` internally does not change the results for users of any other language and is just an internal implementation detail

marnusw: I found that in some cases the three letter time zone printed by `Intl.DateTimeFormat` differs for different locales. For example, when the locale is `en-US` the `Europe/Paris` time zone prints as GMT+2, but when using `en-GB` it prints as CEST. It is really unfortunate that this happens, because aside from accommodating this there wound't be any need to pass a locale as you say.

benmccann: I'm afraid I'm still not seeing the issue. `makeDtf` is only called by `tzOffset` in my PR which returns a number so where is the formatting to GMT+2 or CEST happening?

You're saying that in the method `tzOffset` that the variable `dtf` has a different value depending on the locale. And then that causes a problem in which case? I.e. if `partsOffset(dtf, date)` or `hackyOffset(dtf, date)` is called?

marnusw:
Sorry, I should have added that the same method is used in ´format´ to get
the three letter time zone name for token ZZZ. I think the locale is only
actually passed in that case which is the only instance where it is needed.
(If it's passed in elsewhere where the formatted tz is not used that would
be an oversight.)


*@benmccann* commented on this pull request.

src/_lib/tzMakeDateTimeFormat/index.js
marnusw: Sorry, I should have added that the same method is used in `format` to get
the three letter time zone name for token `ZZZ`. That is the only
time the locale is actually passed into `makeDtf`. When used to tokenize the date no locale is passed and it will always use the cached `en-US` DTF as you had in your implementation.

aphillipo:
Is there a chance of this being merged, (and a new release created) it would completely remove the need for Luxon which other people want to add to our code base. Date-fns is far superior apart from in to support! Thanks.

src/_lib/tzMakeDateTimeFormat/index.js
benmccann: Makes sense. Thanks for taking this effort further!

benmccann:
@marnusw looks like this PR will need to be rebased

benmccann:
@leshakoss any thoughts on this PR?

marnusw:
@benmccann You're right, I'll make some time to rebase this in the near future. If we get any positive feedback on the possibility of having it merged I'll make it a priority.

benmccann:
@leshakoss just a quick reminder about this PR. Thanks!

mesqueeb:
I also require `IANA Time Zone Database` support for working with Google Calendar API.
I just need to retrieve the current timezone in the `IANA` format.
Is there a way to already use this?

majelbstoat:
This would also be incredibly helpful for handling the rendering of dates in a server-side rendered react app. The inability to set a timezone causes the server and client to render different strings, which causes a flash of incorrect dates, and warnings from react about differing rendered results.

marnusw:
@mesqueeb I think this might actually not solve your problem even if this PR is merged. As far as I have found the `Intl` API takes an IANA time zone as input and you can use it to format dates in that time zone, but it can't be used to retrieve the current time zone and in fact doesn't have the option to format a date to IANA time zone (only to long and short local names, e.g. Eastern Standard Time / EST).

I would imagine the Google Calendar API might accept offsets as well, in which case you might use `new Date().getTimezoneOffset()` to get what you need.

aphillipo:
I would really love this to be merged - what is the big problem - it seems well tested and I've ended up having to copy the code into my codebase... Not ideal!

robclancy:
Really? A date library without timezone support?

motin:
Related: https://www.npmjs.com/package/date-fns-timezone

marnusw:
This became a bit of a nightmare to rebase and since it will not be merged until v2 is released in any case I have released https://www.npmjs.com/package/date-fns-tz yesterday which is based on this PR. I hope that it will not only provide the answer to some time zone related use cases but also form part of the eventual discussion on including time zone support in `date-fns` itself.

kossnocorp:
@marnusw thank you a lot for working on it! I'm going to promote your package in [Twitter](https://twitter.com/date_fns). It would be nice to have a dedicated doc (e.g. https://date-fns.org/v2.0.0-alpha.25/docs/Time-Zones) on the website with the list of possible solutions. I would appreciate if you will help with that.

marnusw:
Sure, thanks @kossnocorp! I will create a PR with a docs page in the next few days.

eljefedelrodeodeljefe:
Great effort, however I am not sure whether using `Intl` is the best approach especially for browsers, in order just to get IANA timezones. The list is rather small, but the ICU dependency not available or available without full locales in many environments, e.g. Node out of the box, or breaking horribly in IE not even allowing parsing JS. If this is first time use of Intl for date-fns, I suggest making testing a priority since it's like sneaking in a npm dep.

marnusw:
An interesting issue came up in [date-fns-tz#15](https://github.com/marnusw/date-fns-tz/issues/15) where someone expected `utcToZonedTime` to parse any string as if it is UTC, even when the string itself has no time zone component, `utcToZonedTime('2019-03-20 10:51:43', 'America/New_York')`, which made me think about extending `parse` to accept a `timeZone` option as well, especially now that `alpha.27` does not do automated string parsing anymore.

I have left some of my initial thoughts on the issue and thought I would post about it here so anyone that has commented on this thread can give input if so inclined. I will appreciate any thoughts.

aryehb:
@leshakoss Now that [v2 has been released](https://github.com/date-fns/date-fns/releases/tag/v2.0.0), can this PR be reconsidered?

aphillipo:
Yes I've been extremely pleased with it!

nicolad:
@leshakoss any updates on this?
It would be nice to pick `date-fns-tz` features from ``date-fns`.

KingDarBoja:
Does this PR will be updated on the meantime and reviewed? It seems a important feature for handling timezones that should be on this library. In the meantime gonna use https://github.com/marnusw/date-fns-tz from same author. Cheers!

marnusw:
I will close this now since there are no plans to merge this at the moment and `date-fns-tz` has become a popular library by now in its own right.

PR #711, opened 2018-03-20
Fix sideEffects flag

alexandernanberg:
Uses the updated flag name instead of the old one. https://github.com/webpack/webpack/tree/master/examples/side-effects


kossnocorp:
Thank you!

PR #727, opened 2018-04-14
Japanese translations

mesqueeb:


mesqueeb:
I updated all Japanese files.
I'm not sure about `match/index.js` how to write these...

PR #744, opened 2018-05-15
Remove `new Date` fallback in `toDate` (closes #743)

### LINKED ISSUE #743: Remove `new Date` fallback in `toDate`
This fallback leads to all functions working differently in different browsers (sinse `toDate` is used to clean the date arguments in all functions). For example, `toDate('07-May-2018')` returns `Invalid Date` in Firefox, but correctly parsed date in Chrome. See also #742

leshakoss:
See #743

### DIFF
--- a/src/toDate/test.js
+++ b/src/toDate/test.js
@@ -228,9 +228,10 @@ describe('toDate', function () {
     })
 
     describe('failure', function () {
-      it('fallbacks to `new Date` if the string is not an ISO formatted date', function () {
+      it('returns `Invalid Date` if the string is not an ISO formatted date', function () {
         var result = toDate(new Date(2014, 8 /* Sep */, 1, 11).toString())
-        assert.deepEqual(result, new Date(2014, 8 /* Sep */, 1, 11))
+        assert(result instanceof Date)
+        assert(isNaN(result))
       })
     })
   })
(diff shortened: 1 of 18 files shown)


src/toDate/test.js
```diff
@@ -259,18 +260,54 @@ describe('toDate', function () {
...
 
-  it('throws `RangeError` if `options.additionalDigits` is not convertable to 0, 1, 2 or undefined`', function () {
-    // $ExpectedMistake
-    var block = toDate.bind(null, '+12340702', {additionalDigits: 3})
-    assert.throws(block, RangeError)
+  describe('argument conversion', function () {
+    it('implicitly converts instance of Number into a number', function () {
+      // eslint-disable-next-line no-new-wrappers
+      var timestamp = new Number(new Date(2016, 0, 1, 23, 30, 45, 123).getTime())
+      // $ExpectedMistake
```
kossnocorp: Hm, why Flow does complaints here?

src/toDate/test.js
```diff
@@ -259,18 +260,54 @@ describe('toDate', function () {
...
+      var result = toDate(timestamp)
+      assert.deepEqual(result, new Date(2016, 0, 1, 23, 30, 45, 123))
+    })
+
+    it('implicitly converts instance of String into a string', function () {
+      // eslint-disable-next-line no-new-wrappers
+      var dateString = new String('2014-02-11')
+      // $ExpectedMistake
```
kossnocorp: Same question here

src/toDate/test.js
leshakoss: Oh, it is probably because the argument type is `*` right now. Should we make it `Date|String|Number`?

kossnocorp: Let's do it

Issue #777, opened 2018-06-11
[V2] fr (french) language support


Lakston:
I'm working on a PR for French language support to match the new API and making good progress.

I have a couple of questions though :

- I have rewritten all the tests but I have a lot of errors, some coming from my test, some coming from the lib tests, even if I comment all my test, running `yarn test` yields errors, is that normal ?

- Concerning my test errors I'm running in a lot of errors like this :
```
      long format
        ✖ short date
          Chrome 67.0.3396 (Mac OS X 10.13.4)
        TypeError: formatLong.date is not a function
            at dateLongFormatter (webpack:///src/format/_lib/longFormatters/index.js:4:24 <- /Users/fab/Code/date-fns/test.js:32988:26)
            at dateTimeLongFormatter (webpack:///src/format/_lib/longFormatters/index.js:35:11 <- /Users/fab/Code/date-fns/test.js:33019:13)
            at webpack:///src/format/index.js:378:15 <- /Users/fab/Code/date-fns/test.js:31460:15
            at Array.map (<anonymous>)
            at format (webpack:///src/format/index.js:374:5 <- myFolder/date-fns/test.js:31456:61)
            at Context.<anonymous> (webpack:///src/locale/fr/test.js:136:28 <-myFolder/date-fns/test.js:77616:44)
```

any hints on what I'm doing wrong ?

- Should I submit a WIP PR or should I wait until I square things out regarding those test errors ?

Thanks for your time !

leshakoss:
@Lakston please submit the PR first so we can take a look at those errors :-)

Lakston:
Thanks ! I submitted the PR, it can be seen [here](https://github.com/date-fns/date-fns/pull/778)

PR #778, opened 2018-06-12
Udpate French locale to match the new V2 API

Lakston:
There are some problems with the tests and I'm bad at regex so I'll have to work more on those.

edit: forgot to mention the associated issue: #777

### DIFF
--- a/src/locale/fr/_lib/formatRelative/index.js
+++ b/src/locale/fr/_lib/formatRelative/index.js
@@ -1,10 +1,10 @@
 var formatRelativeLocale = {
-  lastWeek: "dddd 'dernier à' LT",
-  yesterday: "'hier à' LT",
-  today: "'aujourd’hui à' LT",
-  tomorrow: "'demain à' LT'",
-  nextWeek: "dddd 'prochain à' LT",
-  other: 'L'
+  lastWeek: "eeee 'dernier à' p",
+  yesterday: "'hier à' p",
+  today: "'aujourd’hui à' p",
+  tomorrow: "'demain à' p'",
+  nextWeek: "eeee 'prochain à' p",
+  other: 'P'
 }
 
 export default function formatRelative (token, date, baseDate, options) {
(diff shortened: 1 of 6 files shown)


Lakston:
Thanks for the review ! Oh lord this made me realise I kinda commited some garbage files, sorry for that !

I looked into the AM / PM some more in French language and it appears that ruels are that we use it if necessary and always capitalize it so I reflected that in the last commit (as seen [here](https://www.unicode.org/cldr/charts/32/summary/fr.html#1928) )

Test is now un commented, same for 'minuit' which should be use in all cases.

Lakston:
Managed to narrow it down to the test errors that I can't understand where the `result` yields weird values such as:

```
assert(result === 'demain à 7:30')
               |      |                    
               |      false                
               "demain à 4513070200000"   
```

and the parsing tests which I have to work on because my regex game is weak.

src/locale/fr/_lib/formatRelative/index.js
```diff
@@ -1,9 +1,9 @@
 var formatRelativeLocale = {
-  lastWeek: '[hier] dddd [à] LT',
-  yesterday: '[hier à] LT',
-  today: '[aujourd’hui à] LT',
-  tomorrow: '[demain à] LT',
-  nextWeek: 'dddd [à] LT',
+  lastWeek: "dddd 'dernier à' LT",
```
leshakoss: Please use `eeee` instead of `dddd`; `p` instead of `LT`; `P` instead of `L`. Those are the `format` tokens: https://date-fns.org/v2.0.0-alpha.10/docs/format we've changed them in 8th alpha

leshakoss:
@Lakston I fixed some of the tests (see the last commit in your branch). Could you please verify that my changes make sense in French? Thanks

Lakston:
I started doing that yesterday :)

I'll have some modifications but I'll make another PR probably because I have some questions to ask first, I'll post those in the related issue.

kossnocorp:
The locale is released as `2.0.0-alpha.11`, thank you for your work!

Issue #776, opened 2018-06-09
Alias some of the more generic method names


cbovis:
Just started migrating from moment to date-fns and I'm loving the API so far. Fantastic job! One thing that's stood out however is that I need to alias some of my imports because the names are too generic. For example I'm aliasing `parse` and `format` as `parseDate` and `formatDate` respectively using `import { format as formatDate } from 'date-fns';`. I'd love to see some of these more generic names being given aliases out of the box so that I could simply write `import { formatDate } from 'date-fns';`

kossnocorp:
That is a great idea 👍

Lakston:
Is this just a matter of renaming the `format`  to `formatDate` in the `module.exports` and keep the `format` in the internals or rename it in the whole app ?

I could take a look at this.

jhpratt:
@kossnocorp Is there a list of possible aliases? I imagine this would be pretty simple to do — I'll give it a shot if there exists a list.

jatin33:
@kossnocorp Is there a list of possible aliases? I might be able to help.

wjhsf:
Here's a list of the current generically-named functions (don't have a date/time word in the name), and some suggested aliases:

Current | Alias | Comments
--- | --- | ---
format | formatDate |
parse | parseDate |
max | maxDate |
min | minDate |
add | addDate | Not sure if is super clear, since the second argument is a special object.
sub | subDate | 👆 
set | setDate | 👆 
closestIndexTo | closestIndexToDate | 
closestTo | closestToDate | closestDateTo is also an option, but the corresponding closestDateIndexTo doesn't sound as good.
formatRelative | formatRelativeDate |
isAfter | isAfterDate or isDateAfter |
isBefore | isBeforeDate or isDateBefore |
isEqual | isEqualDate or isDateEqual |
isExists | dateExists or isExistingDate | This one I'd do a bit different because "is exists" isn't grammatically correct.
isMatch | matchesFormat and matchesDateFormat | The first alias is because I don't think isMatch is super descriptive of what the function does. The second is derivative of the first.
isValid | isValidDate or isDateValid |
lightFormat | lightFormatDate |
parseJSON | parseJSONDate or parseDateJSON |

niharika412:
Hi, Is anybody working on this? I would like to give it a go.

with-shrey:
@kossnocorp Is anyone working on this issue ?

ryanolsonx:
[...]

PR #779, opened 2018-06-12
[refacto] renamed format to formatDate

Lakston:
WIP of #776

I chose to do only one at a time since I don't really know if what I did actually works and if what I did made sense. Basically I renamed all the occurrences of the `format` function to `formatDate` except in the build generated files.

I would have liked to be able to run the build script so I could check that the generated files reflect the renaming of the format function and so I could run the tests (I get an error because the fp files refer to a folder that does not exists anymore), how can I do that locally ?

No worries if that's not the right approach and it get scrapped, I learned to use rebase in the process so it's still a win for me !


leshakoss:
Sorry for not responding to you in that issue.

That is not the right approach :-) The plan was to make an alias, not fully rename the function, which would mean that the function is accessible by both names in any way.

Lakston:
Hey no worries, I was thinking that would be waaaay better, but would you do that in the build script so it creates 2 exports then ?

I wasn't sure modifying the build script to include something that specific was the right approach either.

leshakoss:
@Lakston I think that the function also should be accessible by `import ... from 'date-fns/formatDate'` for consistency, it should be searchable in the docs (and the docs should contain a note that it is an alias to what), and it has to be tested (at least some basic regular test + 2 FP tests: for `formatDate` and `formatDateWithOptions`). I am actually not sure what exact implementation would be the best in these case: either generating the whole functionality about the function during the build process, or treating it as a regular function (with content `import format from '../format'; export default format`), and doing a special case only for the documentation.

Lakston:
I'll see if I can come up with something, I'll close this PR in the mean time, thanks for your input !

PR #790, opened 2018-06-21
add locale zh-CN

cubicwork:
change locale file for zh-CN


cubicwork:
Add Simplified Chinese language support for date-fns v2.0

cubicwork:
Lack of test cases

PR #791, opened 2018-06-21
Add zh-CN support for v2.0

cubicwork:
Add Simplified Chinese language support for date-fns v2.0


cubicwork:
due to some lint errors

PR #793, opened 2018-06-22
add alias parseDate for parse (#776)

xkizer:
- added new function `parseDate`, an alias of `parse`
 - fixed bugs in the build scripts that causes functions without params or return or exceptions to fail

See #776


xkizer:
**Important to note:**

I simply aliased `parse`.

I initially provided no documentation of the `parseDate` "function" and simply referred to the `parse` function instead in the summary and description. However, typescript typings require the documentation to generate proper typings. I thus copied the relevant documentation from `parse` without modification.

I modified the build scripts to be able to parse functions without `@param`s (like this one) or`@throws`s. If we could be able to solve the TS typing issue, we will possibly be doing this a lot.

xkizer:
**Important to note:**

I subsequently modified the build script to pick the relevant documentation for aliases directly from the aliased functions. All that is needed are `@name` and `@alias` on the alias.

kossnocorp:
Sorry that after your great work, I have to close it. I appreciate it, but sadly, I can't merge or update it as the whole build system has changed.

Besides, I believe that we can do it more easily with exporting aliases like:

image

PR #807, opened 2018-07-11
Updated pt-BR locale for 2.0

YagoCarballo:
Updated the pt-BR locale to work with the 2.0.0 changes. (Portuguese is not my main language, but I tried to follow the same translations / as moment-js for `pt-BR` whenever possible)


hugows:
Any reason for not following en-US capitalization of days of the week eg "Mon" "Tue"?

PR #811, opened 2018-07-17
Create and test Dutch translations (nl)

curry684:
I happened to need it   😆


curry684:
Travis fail unrelated btw.

src/locale/nl/_lib/localize/index.js
```diff
@@ -1,21 +1,60 @@
 import buildLocalizeFn from '../../../_lib/buildLocalizeFn/index.js'
-import buildLocalizeArrayFn from '../../../_lib/buildLocalizeArrayFn/index.js'
 
-var weekdayValues = {
-  narrow: ['zo', 'ma', 'di', 'wo', 'do', 'vr', 'za'],
-  short: ['zon', 'maa', 'din', 'woe', 'don', 'vri', 'zat'],
-  long: ['zondag', 'maandag', 'dinsdag', 'woensdag', 'donderdag', 'vrijdag', 'zaterdag']
+var eraValues = {
+  narrow: ['BC', 'AD'],
```
stefanvermaas: I agree that this is the main difference now. I'm not quite sure what the official rules are on this, but in official texts I never see BC/AD. The only thing I could find about this is these links; https://onzetaal.nl/nieuws-en-dossiers/poll/voor-na-christus-vervangen/P50 and https://taaladvies.net/taal/advies/vraag/1559/vc_vchr/. 

Seen these links it seems that the correct version would be v. Chr.

leshakoss: @stefanvermaas [we recommend](https://github.com/date-fns/date-fns/blob/master/docs/i18nContributionGuide.md#localizeera-and-using-buildlocalizefn-function) to use the values [from CLDR chart](https://www.unicode.org/cldr/charts/32/summary/nl.html#1779)

curry684: If that's authoritative so it'll be. Just curious because literally nobody here ever says it like that, it's rather archaic.

leshakoss: @curry684 well then I would use the version that is used more often / sounds the least out of place :-)

curry684:
Fixed quarter/era notations to CLDR preference and added @stefanvermaas as co-author.

src/locale/nl/_lib/localize/index.js
curry684: Thing is - I'm not sure what "current local preference" would be, as it's not really a subject I discuss daily  😉    It's just that in these days the religious references are usually shunned, but I can't really say that's a "national sentiment". So I just changed it to the CLDR preference.

curry684:
Go 👍

src/locale/nl/_lib/match/index.js
```diff
@@ -1,47 +1,108 @@
...
+var matchOrdinalNumberPattern = /^(\d+)e?/i
+var parseOrdinalNumberPattern = /\d+/i
 
-var matchWeekdaysPatterns = {
-  narrow: /^(su|mo|tu|we|th|fr|sa)/i,
-  short: /^(sun|mon|tue|wed|thu|fri|sat)/i,
-  long: /^(sunday|monday|tuesday|wednesday|thursday|friday|saturday)/i
+var matchEraPatterns = {
+  narrow: /^(BC|AD)/,
+  abbreviated: /^(BC|AD)/,
```
stefanvermaas: I’m missing here the “v.Chr etc”, because now it can only parse the BC/AD and not the Dutch variations. Is this intended?

leshakoss: @stefanvermaas that's a good point. @curry684 generally `parse` should parse any input that `format` could have. Could you please change the regexps to match the “v.Chr” format as well as BC/AD format?

curry684: Doh, my bad. Weird that tests don't catch this?

curry684: Oh wait I didn't change the test cases either ofc.

curry684:
Fixed a bunch of matchers, also caught a missing translation on month matching that wasn't causing test failures as the parser was fixed correctly.

leshakoss:
@curry684 @stefanvermaas thank you for your contribution! ⭐️ The locale is released in [v2.0.0-alpha.14](https://gist.github.com/kossnocorp/a307a464760b405bb78ef5020a4ab136#v200-alpha14)

PR #812, opened 2018-07-17
Update for the Dutch locales

stefanvermaas:
Hi there 👋,

Thanks for this wonderful project. Hereby my contribution in the form of the updated Dutch translations for the 2.0.0 version of the project. The tests are 100% green. 

Looking forward to your comments!

Cheers, Stefan


msikma:
Nice work. I didn't test it myself but reading through the test case it looks good to me. Just mentioning this since I'm not sure any of the reviewers can speak Dutch.

src/locale/nl/_lib/formatRelative/index.js
```diff
@@ -1,10 +1,10 @@
 var formatRelativeLocale = {
-  lastWeek: '[vorige] dddd [om] LT',
-  yesterday: '[gisteren om] LT',
-  today: '[vandaag om] LT',
-  tomorrow: '[morgen om] LT',
-  nextWeek: 'dddd [om] LT',
-  other: 'L'
+  lastWeek: "'vorige week' eeee 'om' pp",
```
curry684: Correct is "afgelopen dinsdag om 12:34".

src/locale/nl/_lib/localize/index.js
```diff
@@ -1,38 +1,128 @@
...
+    noon: 'mid.',
+    morning: 'ochtend',
+    afternoon: 'mid.',
+    evening: 'avond',
+    night: 'nacht'
+  },
+
+  wide: {
+    midnight: 'middernacht',
+    noon: 'middag',
```
curry684: "Het middaguur".

src/locale/nl/_lib/localize/index.js
```diff
@@ -1,38 +1,128 @@
...
+    morning: 'ochtend',
+    afternoon: 'middag',
+    evening: 'avond',
```
curry684: 's ochtends, 's middags', 's avonds.

stefanvermaas:
Ah I see @curry684, let's work on this together then and improve the PR! I'll look into your comments, and see what I can change.

src/locale/nl/_lib/formatRelative/index.js
stefanvermaas: Hmm I think this depends on the context right? The variable says; "last week" and I interpreted this as "the Tuesday of the previous week".

When "last week" means this past Tuesday (when today is Wednesday) I would say you're correct. When "last week" means the Tuesday of the previous week, I would say this should be "vorige week dinsdag".

Maybe one of the maintainers can shed some light on this? What kind of "Tuesday" is meant with "last week"? The one that is still in the current week, or the one that is in the previous week before this week.

src/locale/nl/_lib/localize/index.js
stefanvermaas: I agree this is the literal translation, but I cannot recall anyone who ever used "het middaguur" in the Netherlands. I think the more used term is "middag". 

Also - according to the dictionary - the real correct translation would be "noen", which means "middag" and then as the other option "middaguur". Maybe I'm wrong, but feel free to correct me then.

src/locale/nl/_lib/formatRelative/index.js
curry684: From what I saw in the tests it's "in the past 7 days". So as today is Monday, it would be triggered for anything up to last Tuesday. Before that is just "other", so by absolute date.

src/locale/nl/_lib/localize/index.js
stefanvermaas: I think this also depend on context as you don't say; "in de 's ochtends", but "in de ochtend". I'll have to double check how these words are used in the tests, but I believe it's highly depending on the context.

src/locale/nl/_lib/formatRelative/index.js
leshakoss: @stefanvermaas @curry684 by default, two days ago is "last week" even if it is the same week. This behaviour could be changed to "last week"/"this week"/"next week" configuration in the locale like this: https://github.com/date-fns/date-fns/blob/master/src/locale/ru/_lib/formatRelative/index.js#L50-L57

src/locale/nl/_lib/localize/index.js
stefanvermaas: I agree ;)! @leshakoss is with "noon" also meant "exactly 12:00PM" or is "noon" more like afternoon (so between 12:00PM -  18:00PM)? If the first option is the case, it should indeed be "middaguur", otherwise I think it's safe to say it's "middag". Do you agree @curry684 ?

src/locale/nl/_lib/formatRelative/index.js
stefanvermaas: @curry684 do you agree we should implement this method for the Dutch translations too? As we've got quite some room for misunderstandings in Dutch when we mean "afgelopen dinsdag" instead of "vorige week dinsdag". 

We could implement it the same as in Russian as @leshakoss shows.

src/locale/nl/_lib/localize/index.js
leshakoss: `noon` is exactly 12 PM. Here is the description from the standard: https://www.unicode.org/reports/tr35/tr35-dates.html#Fixed_periods

src/locale/nl/_lib/formatRelative/index.js
curry684: The Dutch rules boil down to:
- Anything over 7 days: full date
- Previous full week: "vorige week <dag>"
- 7 days to 2 days ago: "afgelopen <dag>"
- 2 days ago: "eergisteren"
- 1 day ago: "gisteren"
- Today: "vandaag"
- 1 day ahead: "morgen"
- 2 days ahead: "overmorgen"
- 2 to 7 days ahead: "aanstaande <dag>"
- Next full week: "volgende week <dag>"
- Anything beyond that: full date

The library doesn't support the 2-day rules and they don't really matter for this purpose as they're also commonly used by day name as indicated.

src/locale/nl/_lib/localize/index.js
curry684: You also need the `het` prefix as it's used in the context of `rond middernacht` and `kort na middernacht` which would have to be `rond het middaguur` and `kort na het middaguur`.

curry684:
Can you do the reverse check on my PR? It needs less changes now    ;)   I think main discrepancy left is AD/BC vs v.Chr./na Chr.

stefanvermaas:
Hi everyone,

Thanks for looking into this so extensively. I think @curry684 is right about using his PR to actually merge it eventually into master. So I'll close this PR, and let's continue the discussion at #811

Issue #810, opened 2018-07-14
Three-letter week days (ddd) look weird in German


Philipp91:
I use date-fns through [material-ui-pickers](https://github.com/dmtrKovalenko/material-ui-pickers), and with the `de` locale. In the date picker dialog, it displays today's date as "Sam, Jul 14", which looks very strange to me, almost like a badly translated American date.

The overall format `ddd, MMM D` is hard-coded in material-ui-pickers. Side question: Is there a function or locale-specific format provided by date-fns that the material-ui-pickers could use? It's for display in a context like [this](https://i.stack.imgur.com/KDD8c.png).

What's most unfamiliar about this date format is the "Sam" -- or worse, yesterday it was "Fre". Although there is [some](https://www.google.ch/search?q="mon+die+mit+don+fre+sam+son") usage of these German three-letter abbreviations on the web, the far [more common](https://www.google.ch/search?q="mo+di+mi+do+fr+sa+so") variant seems to be the two-letter abbreviation. More evidence:
https://github.com/contao/core/issues/2874
https://www.drupalcenter.de/node/28207
https://www.experto.de/sprachen/englisch/the-days-of-the-week-die-wochentage-auf-englisch.html

I'm not sure if date-fns already has an answer to the question whether `dd` and `ddd` mean "short" and "abbreviated", respectively, or whether it's literally "two-letter" and "three-letter". In the former case, I suggest making `dd` the same as `ddd` for the German locale (I can send a pull request). In the latter case, see my question above for locale-specific date formats (including commas and such), which would never use `ddd` in German. Looking at the Japanese locale, I'm thinking it's probably the former definition, as their short and abbreviated day names are both a single character.

leshakoss:
@Philipp91 Thank you for using `date-fns`!

In the current stable release (aka v1) the definition of those format tokens are a bit ambiguous and were essentially the matter of preference of locale creators and the people who were reviewing their PRs. We didn't restrict `ddd` to three letter abbreviations, but the `de` locale as many others was based on `en-US` locale, which had lead to this unfortunate translation. If you want this to change, please send the PR to [`v1` branch](https://github.com/date-fns/date-fns/tree/v1). Though it could take some time to merge and release it because we are currently focusing on developing `v2`.

[...]

PR #817, opened 2018-07-25
#810 Correct short and abbreviated dayValues for de locale

Philipp91:
The values are taken from https://www.unicode.org/cldr/charts/32/summary/de.html#1844, using "Days - (abbreviated|short) - Formatting".
Alternatively, the short values could be left as is (without the period). According to the CLDR table, that's the Swiss locale, but it would provide a two-letter version matching the English locale.

### DIFF
--- a/src/locale/de/_lib/localize/index.js
+++ b/src/locale/de/_lib/localize/index.js
@@ -24,8 +24,8 @@ var monthValues = {
 
 var dayValues = {
   narrow: ['S', 'M', 'D', 'M', 'D', 'F', 'S'],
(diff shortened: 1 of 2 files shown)


Philipp91:
Done

leshakoss:
I've [changed the short values to be without the period](https://github.com/date-fns/date-fns/commit/ff33cc85afe31e129d8bd7dee3d7f7c6d2918390), and released this as [v2.0.0-alpha.14](https://gist.github.com/kossnocorp/a307a464760b405bb78ef5020a4ab136#v200-alpha14). Thank you for your contribution! ⭐️

Philipp91:
Thanks!

PR #821, opened 2018-07-30
In `parse`, return Invalid Date if there is any remaining non-whitespace input (closes #819)

### LINKED ISSUE #819: Options for strict parsing?
I'm using `date-fns` (latest 2.0 alpha) to validate that certain datetime strings are in a specific ISO8601 compliant format. It's critically important that they are in the exact format I'm testing for so that certain invariants hold.

However, this parses as valid:

```
> dateFns.parse("2012-12-12T12:30", "YYYY-MM-dd", new Date())
2012-12-11T22:00:00.000Z
```

and even this:

```
> dateFns.parse("2012-12-12asfasfasfT12:30", "YYYY-MM-dd", new Date())
2012-12-11T22:00:00.000Z
```

I propose the parser should check if there's any remaining input after parsing, as long as a format was provided. Best-effort parsing is okay for UI use, but for critical validation tasks this is unwanted behaviour.

There should also be an option for disabling the `Date` constructor fallback, for the same reasons.

leshakoss:
See #819 for more details

### DIFF
--- a/src/parse/_lib/parsers/index.js
+++ b/src/parse/_lib/parsers/index.js
@@ -517,7 +517,6 @@ var parsers = {
       }
     },
     validate: function (date, value, options) {
-      console.log({value})
       return value >= 0 && value <= 11
     },
     set: function (date, value, options) {
--- a/src/parse/index.js
+++ b/src/parse/index.js
@@ -23,6 +23,8 @@ var formattingTokensRegExp = /[yYQqMLwIdDecihHKkms]o|(\w)\1*|''|'(''|[^'])+('|$)
 var escapedStringRegExp = /^'(.*?)'?$/
 var doubleQuoteRegExp = /''/g
 
+var notWhitespaceRegExp = /\S/
+
 /**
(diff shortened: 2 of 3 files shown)


src/parse/_lib/parsers/index.js
```diff
@@ -517,7 +517,6 @@ var parsers = {
       }
     },
     validate: function (date, value, options) {
-      console.log({value})
       return value >= 0 && value <= 11
```
kossnocorp: Hah

curry684: That line just cost me an hour because it was crashing my webpack builds in CI   😢

leshakoss: @curry684 sorry :[ It's removed in v2.0.0-alpha.16

curry684: Hehe yeah already upgraded, they're succeeding again. My own fault for using alpha  ;)

Issue #524, opened 2017-07-07
Remote `docs.json` file from npm package


mrmlnc:
Hello, 

I think you can remove the `docs.json` file from the npm package. Just -204K.

kossnocorp:
Done: https://github.com/date-fns/date-fns/commit/d309d596b8d5ad0ede44a960d2fc098af8685a53

Thank you for the suggestion.

PR #837, opened 2018-08-11
Add docs.json to gitignore

hawkrives:
see also https://github.com/date-fns/date-fns/issues/524

I guess it crept back in 😅 

npm will use `.gitignore` if you don't provide an `.npmignore`, so I added it to .gitignore since it's a build artifact AFAICT.

```
npm notice
npm notice 📦  date-fns@1.29.0
npm notice === Tarball Contents ===
npm notice 3.6kB   package.json
npm notice 26.6kB  CHANGELOG.md
npm notice 291.3kB docs.json
...
```

Do you want a CHANGELOG entry for this?


kossnocorp:
The change was released as `date-fns@2.0.0-alpha.20`. See the change log: https://gist.github.com/kossnocorp/a307a464760b405bb78ef5020a4ab136#v200-alpha20

PR #870, opened 2018-09-14
add support for getUnixTime

### LINKED ISSUE #372: get unix timestamp of given date

Kingwl:
fix #372 again


Kingwl:
Is that mean we should rename the `toUnix ` to `toUnixTime`?🤫

kossnocorp:
@Kingwl 😅

kossnocorp:
The function was released as `date-fns@2.0.0-alpha.20`. See the change log: https://gist.github.com/kossnocorp/a307a464760b405bb78ef5020a4ab136#v200-alpha20 Thank you again for your work!

PR #874, opened 2018-09-17
tests: add Node.js 8 and 10

DanielRuf:


DanielRuf:
> This will make builds run 3x slower. I'd prefer sticking to LTS release

This should be an issue. Also 2 run always in parallel.
Well, there were breaking changes in 10 so yes, it makes sense to test 8 and 10. It is also about testing the devDependencies and all other dependencies.

DanielRuf:
> This will make builds run 3x slower. I'd prefer sticking to LTS release

This should not be an issue. Also 2 builds always run in parallel.
Well, there were breaking changes in 10 so yes, it makes sense to test 8 and 10. It is also about testing the devDependencies and all other dependencies.

DanielRuf:
6 is also still used on many clients even if it reached its EOL.

https://nodesource.com/node-by-numbers

DanielRuf:
What do you suggest, how should we proceed further? Just test 8 and 10 and make a new major release?

kossnocorp:
I've created an issue "Consider running tests in Node.js": https://github.com/date-fns/date-fns/issues/875. Until it's executed, I don't see a reason running tests in different Node environments.

In October v10 will enter LTS phase, so I think it's reasonable to upgrade 6 to 10 as it speed ups the build.

DanielRuf:
Done.

kossnocorp:
Thank you!

PR #880, opened 2018-09-20
Improve ja translation of distanceInWords

kudohamu:
I fixed a unnatural japanese translation of `distanceInWords` of v1.
Currently, `almostXYears` with a suffix is translated into Japanese `X年以下前`, but this expression is not used by Japanese.
So I changed it to be `X年ぐらい前` which is a natural Japanese translation.

`formatDistance` has a same problem, so if this PR is merged, I will also create PR for `formatDistance` to master branch :)

### DIFF
--- a/src/locale/ja/build_distance_in_words_locale/index.js
+++ b/src/locale/ja/build_distance_in_words_locale/index.js
@@ -64,7 +64,9 @@ function buildDistanceInWordsLocale () {
 
     almostXYears: {
       one: '1年以下',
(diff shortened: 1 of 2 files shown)


kossnocorp:
Thank you! I'll ship the change with a new v1 version.

kossnocorp:
The fix for the locale was released with `v1.30.1`. See: https://blog.date-fns.org/post/v130-is-out-m35nqh4m13r7

Thank you again!

PR #902, opened 2018-10-02
New Method: monthStringToNumber

ammuench:
Added a new method which takes the name or short-code of a month and returns the appropriate value (e.g. 'October' ==> '10', 'Apr.' ==> '04').  It's a very simple method but I've found this to be useful in various projects where date values are (unfortunately) not standardized and read from various strings.  Since I was already using `date-fns` to handle other date-related things in these projects I felt it could be a worthwhile addition.

Currently set to just support English names and short-codes, but could possibly be expanded to provide support for other languages.


kossnocorp:
Thank you for your work, but we can't accept this pull request, sorry:

1. The date-fns functions shouldn't depend on a particular language.
2. The implementation is very specific: it accepts months in a specific format (`SEPTEMBER`, `SEPT`, `SEP` but `JULY`, `JUL`), it returns number as strings, the strings are padded to 2-digits, the code performs some magic transformations (`normalizeMonthName.split('.')[0]`).
3. Because of the specificity, it's better to keep such helpers in the app codebase.

We could do it right and add the functionality to every locale, but it will take a lot of effort. Because of the narrow application, it's not worth it.

PR #913, opened 2018-10-05
V1: Fix monthsFull misspelling in TH locale.

ratchapol-an:
It's January in Thai locale.
From 'มกราคาม' to 'มกราคม'


kossnocorp:
Thank you for the fix! I'm sorry that it took so long, we've been focused on v2, and v1 branch got overlooked. I'm going to release the change soon along with a new v1 version.

kossnocorp:
The fix was released with `v1.30.1`. See: https://blog.date-fns.org/post/v130-is-out-m35nqh4m13r7

Thank you again!

PR #928, opened 2018-10-13
add new function roundToNearestMinutes

### LINKED ISSUE #787: Round to nearest X of interval
Sorry if this has already been answered, but I searched the issues and docs and didn't see anything like this.

Is there a way to round to the nearest X of an interval with `date-fns`? For example, is there anything like this?

``` js
roundToNearestMinute(new Date())
roundToNearestMinutes(new Date(), 5)
roundToNearest(new Date(), 5, "minutes")

startOfMinutes(new Date(), 5)
startOf(new Date(), 5, "minutes")

endOfMinutes(new Date(), 5)
endOf(new Date(), 5, "minutes")
```

xkizer:
Fixes #787


kossnocorp:
The PR was released as `v2.0.0-alpha.23`. See the change log: https://gist.github.com/kossnocorp/a307a464760b405bb78ef5020a4ab136#v200-alpha23

Thank you for your great work!

dcrystalj:
this is very limited just 1 to 30 minutes. What about 33 minutes ? 
Also there should be option to round to nearest seconds, hours. days, weeks...

kossnocorp:
@dcrystalj please elaborate your use case for rounding to 33 minutes. Regarding the seconds, hours, whatever, the function is called `roundToNearestMinutes` so it's should not.

dcrystalj:
@kossnocorp sorry my point was just that we need more generic function. I would expect of this fn to work if I give it roundToNearestMinutes(date, 120) same as roundToNearestHour(date, 2)

RIght now in my project I need function to round to nearest hour, 3 hours, day and 3 day... Maybe we should just open issues to implement `roundToNearestHour`, and `roundToNearestDay` ...

kossnocorp:
@dcrystalj yeah, that makes sense. Feel free to open those, but also send PRs.

xkizer:
@kossnocorp if you file a separate issue I can try to work on it. Thanks for the feedback.

Issue #933, opened 2018-10-17
ES6 syntax in the npm build


z4o4z:
After adding this lib to the project Webpack starts crashing for production build. After some investigation, I found that the problem connected with ES6 syntax in the `protectedTokens/index.js` file. UglifyJS can't process this file.

`protectedTokens/index.js`:
```
export const protectedTokens = ['D', 'DD', 'YY', 'YYYY']

export function isProtectedToken(token) {
  return protectedTokens.indexOf(token) !== -1
}

export function throwProtectedError(token) {
  throw new RangeError(
    '`options.awareOfUnicodeTokens` must be set to `true` to use `' +
      token +
      '` token; see: https://git.io/fxCyr'
  )
}

```

Error:
```
ERROR in js/vendors~main.8f103454cf43e9f26f43.chunk.js from UglifyJs
Unexpected token: keyword (const) [js/vendors~main.8f103454cf43e9f26f43.chunk.js:33464,0]
```

After removing ES6 code, everything works fine

endbay:
I have exactly the same issue. Changed Momentjs to date-fns today.
>ERROR in static/main.5183631e20bfc2d5b817.js from UglifyJs
Unexpected token: keyword (const) [static/main.5183631e20bfc2d5b817.js:12226,0]

janslu:
Same here. As a quick fix I changed `export const ...` to `export var ...`

kossnocorp:
Sorry, that's my bad. I'll fix the `var` and publish a new version.

kossnocorp:
The fix was released as `v2.0.0-alpha.23`. See the change log: https://gist.github.com/kossnocorp/a307a464760b405bb78ef5020a4ab136#v200-alpha23

Thank you for reporting the problem!

PR #934, opened 2018-10-18
Fix usage of const causing UglifyJS fail on date-fns

kossnocorp:
See: https://github.com/date-fns/date-fns/issues/933


kossnocorp:
That was released as `v2.0.0-alpha.23`. See the change log: https://gist.github.com/kossnocorp/a307a464760b405bb78ef5020a4ab136#v200-alpha23

PR #938, opened 2018-10-19
Fix error in documentation (example) of getUnixTime

### LINKED ISSUE #936: getUnixTime has wrong example
The function `getUnixTime` has a wrong usage example. See https://date-fns.org/v2.0.0-alpha.23/docs/getUnixTime#examples

xkizer:
Fixes #936


kossnocorp:

marcus-hiles:
Great @kossnocorp, couldn't find a documentation on the use of this new function, can you please share a link to that?

PR #939, opened 2018-10-19
Remove ES6 'const' from bn locale

charliedavison:
When using locales webpack/uglify fails with 

```
[webpack] errors:                                                               
app/6ec78c4b5b.js from UglifyJs
Unexpected token: keyword (const) [../wa-storybook/node_modules/date-fns/esm/locale/bn/_lib/localize/index.js:141,0][app/6ec78c4b5b.js:98206,2]
```

This patch changes const to var.


kossnocorp:

PR #940, opened 2018-10-19
Fix pt-BR translation

wregis:
I fixed some spanish translations on the pt-BR translation file and test.


kossnocorp:

Issue #950, opened 2018-10-23
Error in documentation of 'subDays' function


johnny-dash:
My issue is very straightforward, but it does confuse me when I look into it.
https://date-fns.org/v1.29.0/docs/subDays#examples

![screen shot 2018-10-23 at 11 41 23 am](https://user-images.githubusercontent.com/13574694/47327418-f191f080-d6b8-11e8-837a-ac6f22d960d3.png)

The picture above is from the example section in 'subDays' function in the documentation. The comment said the date should be 1st September while the date object is 1st August. 

It makes me feel that it will only work on the day while keeping the month as it is when I first time glanced the doc.

kossnocorp:
Months start at 0 in JavaScript, so `8` represents September:

```js
new Date(2014, 8, 1).toString()
//=> "Mon Sep 01 2014 00:00:00 GMT+0300 (Eastern European Summer Time)"
```

PR #951, opened 2018-10-23
fix the wrong date object in the documentation of subDays function

johnny-dash:
This pull request is corresponding to the open issues #950


kossnocorp:
See https://github.com/date-fns/date-fns/issues/950#issuecomment-432137752

PR #792, opened 2018-06-21
Add zh-CN locale

cubicwork:
Add Simplified Chinese language support for date-fns v2.0

PR #1013, opened 2018-12-12
change date pattern to meet the nation standard (GB/T 7408-2005)

cubicwork:
change date pattern to meet the nation standard (GB/T 7408-2005) which referenced by wikipedia [Date format by country](https://en.wikipedia.org/wiki/Date_format_by_country)
details in https://github.com/date-fns/date-fns/pull/792

### DIFF
--- a/src/locale/zh-CN/_lib/localize/index.js
+++ b/src/locale/zh-CN/_lib/localize/index.js
@@ -13,9 +13,48 @@ var quarterValues = {
 }
 
 var monthValues = {
-  narrow: ['一', '二', '三', '四', '五', '六', '七', '八', '九', '十', '十一', '十二'],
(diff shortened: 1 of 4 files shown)


src/locale/zh-CN/_lib/localize/index.js
```diff
@@ -90,7 +129,7 @@ var formattingDayPeriodValues = {
   }
 }
 
-function ordinalNumber (dirtyNumber, dirtyOptions) {
+function ordinalNumber(dirtyNumber, dirtyOptions) {
```
lc-soft: Is this coding style correct? the [JavaScript Standard Style](https://standardjs.com/#standardjs--the-rules) requires a space after function name.

cubicwork: @lc-soft there is no coding style problem, we use style check and auto style fix plugin in every build and commit .

PR #1015, opened 2018-12-13
Add `lightFormat` and optimize size of `toDate` (closes #1010)

### LINKED ISSUE #1010: `lightFormat` function
`format` is one of the most (if not the most) popular function in date-fns, but because it involves locales it's one the most (if not the most) heavyweight function in date-fns.

Given that v2 locales became even more heavyweight, I want to introduce `lightFormat` that will cover the most common cases and address the size problem.

The current goal is to be able to print ISO 8601 plus `hh` representing 12-hour clock and `a..aaa` representing am/pm.

kossnocorp:
From #1010:

>`format` is one of the most (if not the most) popular function in date-fns, but because it involves locales it's one the most (if not the most) heavyweight function in date-fns.
>
> Given that v2 locales became even more heavyweight, I want to introduce `lightFormat` that will cover the most common cases and address the size problem.
>
> The current goal is to be able to print ISO 8601 plus `hh` representing 12-hour clock and `a..aaa` representing am/pm.

This PR:

- Introduces `lightFormat` described in the quote above. The current function size is `1.73 KB`.
- Optimizes `toDate`size from `1.48 KB` to `1.27 KB`.

Unfortunately, it's not possible to bring the size significantly down. The bottleneck is `toDate` function and specifically the code responsible for parsing ISO 8601 strings.

### DIFF
--- a/src/lightFormat/index.js
+++ b/src/lightFormat/index.js
@@ -1,0 +1,134 @@
+import toDate from '../toDate/index.js'
+import formatters from '../_lib/format/lightFormatters/index.js'
+import getTimezoneOffsetInMilliseconds from '../_lib/getTimezoneOffsetInMilliseconds/index.js'
+import isValid from '../isValid/index.js'
+import subMilliseconds from '../subMilliseconds/index.js'
+
+// This RegExp consists of three parts separated by `|`:
+// - (\w)\1* matches any sequences of the same letter
+// - '' matches two quote characters in a row
+// - '(''|[^'])+('|$) matches anything surrounded by two quote characters ('),
+//   except a single quote symbol, which ends the sequence.
+//   Two quote characters do not end the sequence.
+//   If there is no matching single quote
+//   then the sequence will continue until the end of the string.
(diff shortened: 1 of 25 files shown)


src/lightFormat/index.js
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * Accepted patterns:
+ * | Unit                            | Pattern | Result examples                   |
+ * |---------------------------------|---------|-----------------------------------|
+ * | AM, PM                          | a..aaa  | AM, PM                            |
+ * |                                 | aaaa    | a.m., p.m.                        |
+ * |                                 | aaaaa   | a, p                              |
+ * | Calendar year                   | y       | 44, 1, 1900, 2017                 |
+ * |                                 | yy      | 44, 01, 00, 17                    |
+ * |                                 | yyy     | 044, 001, 000, 017                    |
```
leshakoss: A small table layout bug here

kossnocorp: Thanks!

PR #1019, opened 2018-12-13
Milliseconds aren't detected after seconds/Unix timestamp (closes #985)

### LINKED ISSUE #985: Milliseconds aren't detected after seconds/Unix timestamp
Here are my reproducing steps:

```
$ grep date-fns package.json 
    "date-fns": "2.0.0-alpha.25"

$ cat test.js 
var df = require('date-fns')
var now = new Date()
mydate = df.parse('1542289282.873', 't.SSS', now)
console.log(mydate)
mydate = df.parse('11:11:11.123', 'HH:mm:ss.SSS', now)
console.log(mydate)
mydate = df.parse('1542289282', 't', now)
console.log(mydate)

$ node test.js 
2018-11-15T13:41:22.000Z
2018-11-15T09:11:11.123Z
2018-11-15T13:41:22.000Z
```

Basically, if I add `t.SSS`, it loses the milliseconds. Timestamp is the same with just `t` and the second-precision timestamp. This doesn't happen with the `HH:mm:ss.SSS` format - I get the millisecond there.

I'm not sure if this expected or a bug, I was trying to parse an nginx `msec` timestamps: https://nginx.org/en/docs/http/ngx_http_core_module.html#var_msec via [Logagent](https://github.com/sematext/logagent-js).

leshakoss:
#985

### DIFF
--- a/src/parse/index.js
+++ b/src/parse/index.js
@@ -9,7 +9,7 @@ import {
   throwProtectedError
 } from '../_lib/protectedTokens/index.js'
 
-var TIMEZONE_UNIT_PRIORITY = 20
+var TIMEZONE_UNIT_PRIORITY = 10
 
 // This RegExp consists of three parts separated by `|`:
 // - [yYQqMLwIdDecihHKkms]o matches any available ordinal number token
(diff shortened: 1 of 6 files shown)


leshakoss:
This PR rearranges priorities of `parse`'s setters in order to add ability to set the fractions of seconds after the seconds/unix timestamp. It utilizes new argument in setter functions - `flags` which is supposed to store any modifiers to be used in later setters. Other possible use cases: set am/pm flag to use in 12-hour clock setter, set era flag to be used in years setter etc

src/parse/index.js
```diff
@@ -485,13 +486,23 @@ export default function parse(
       return new Date(NaN)
     }
 
-    utcDate = setter.set(utcDate, setter.value, subFnOptions)
+    var result = setter.set(utcDate, flags, setter.value, subFnOptions)
+    if (Array.isArray(result)) {
+      utcDate = result[0]
+      flags = Object.assign({}, flags, result[1])
```
kossnocorp: I'm afraid we can't use `Object.assign` as it doesn't support by IE (at all).

PR #1021, opened 2018-12-14
Update example in README

isnifer:
Just tried to run an example on date-fns.org and found an error:
image


kossnocorp:
The example in the README is for v2, for v1 see: https://github.com/date-fns/date-fns/tree/v1

PR #1044, opened 2019-01-19
Refactor Belarusian locale: add declensions and genitive cases

alyrik:
Recently added version of `be` locale has syntactical errors and doesn't reflect an actual language use. It was based on `en` locale structure but, in fact, Belarusian is much closer to Russian and Polish. So I have refactored it and added several improvements: declensions, genitive cases, a distinction between standalone and combined (with month) versions of ordinal day.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ This change log follows the format documented in [Keep a CHANGELOG].
 
 ## [Unreleased]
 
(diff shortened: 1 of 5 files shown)


kossnocorp:
Sorry, I have to freeze v1 and close all PRs to it. It became a very challenging task to release v1 because of differences between v1 and v2. Moving forward, we'll be more considerate while upgrading major versions. Sorry for the troubles.

PR #1045, opened 2019-01-19
fix(parseISO): Check for empty date after splitDateString

### LINKED ISSUE #1043: parseISO throws TypeError
When `splitDateString()` returns with `dateStrings.date = null`, `parseYear()` will throw `TypeError: Cannot read property 'match' of null`.
This is with 2.0.0-alpha.27.
See:
https://github.com/date-fns/date-fns/blob/e83c2d815d6363e33bf852b08a5860aed644ecda/src/parseISO/index.js#L154
and
https://github.com/date-fns/date-fns/blob/e83c2d815d6363e33bf852b08a5860aed644ecda/src/parseISO/index.js#L187

stefanprobst:
Don't crash when `splitDateString()` returns with `dateStrings.date = null`. This can happen when parsing a string containing a colon, e.g."00:00".

Fixes #1043

### DIFF
--- a/src/parseISO/index.js
+++ b/src/parseISO/index.js
@@ -103,8 +103,12 @@ export default function parseISO(argument, dirtyOptions) {
   }
 
   var dateStrings = splitDateString(argument)
-  var parseYearResult = parseYear(dateStrings.date, additionalDigits)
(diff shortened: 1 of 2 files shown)


kossnocorp:

Issue #1040, opened 2019-01-14
getWeekOfMonth() with options.weekStartsOn set to 1 doesn't work for Sundays


daandvs:
When executing `getWeekOfMonth(new Date(2019, 4, 5), { weekStartsOn: 1 })` it returns me `2` while I expect it to be `1`. 
It seems `getWeekOfMonth()` is always a week off when executed with a day set to Sunday and `options.weekStartsOn` set to > 0.

waseemahmad31:
This seems to an issue. I also have reproduced this. Can I work on this?

kossnocorp:
Thank for reporting it!

bitshadow:
@daandvs  which version of `date-fns` includes `getWeekOfMonth`?

daandvs:
It's in the alpha versions of v2 (https://date-fns.org/v2.0.0-alpha.27/docs/getWeekOfMonth)

________________________________
Van: Jignesh Kakadiya <notifications@github.com>
Verzonden: donderdag 31 januari 2019 14:50
Aan: date-fns/date-fns
CC: Daan Devos; Mention
Onderwerp: Re: [date-fns/date-fns] getWeekOfMonth() with options.weekStartsOn set to 1 doesn't work for Sundays (#1040)


@daandvs<https://github.com/daandvs> which version of date-fns includes getWeekOfMonth?

leshakoss:
Fixed in #1046 and released as [v2.0.1](https://github.com/date-fns/date-fns/releases/tag/v2.0.1)

PR #1046, opened 2019-01-19
fixed issue: options.weekStartsOn set to 1 doesn't work for Sundays

waseemahmad31:
fixed issue: #1040 

When executing `getWeekOfMonth(new Date(2019, 4, 5), { weekStartsOn: 1 })` it returns `2` while I expect it to be 1. It seems `getWeekOfMonth()` is always a week off when executed with a day set to `Sunday` and `options.weekStartsOn set to > 0`.


leshakoss:
The fix was released as [v2.0.1](https://github.com/date-fns/date-fns/releases/tag/v2.0.1). Thanks for your contribution! 🙏

PR #1050, opened 2019-01-21
Improve japanese translation of formatDistance with addSuffix option

kudohamu:
I fixed unnatural japanese translations of `formatDistance`.
`lessThanXXX` with a suffix is translated into Japanese `XXX未満前`, but this expression is not used by Japanese.


kossnocorp:

Issue #1042, opened 2019-01-15
Feature request: support fraction of seconds in lightFormat


alex3683:
Since I don't need all that locale specific stuff of format, I'd really like to use lightFormat. Alas, lightFormat doesn't support the fraction of seconds patterns (`S*`). I'm not really sure, if the fraction of seconds are an important part of ISO 8601, but for example in Chrome Console `new Date().toISOString()` prints them. So maybe adding them would make lightFormat more consistent / complete in regard to known date ISO formatting.

PR #1060, opened 2019-01-28
support fraction of seconds in lightFormat

waseemahmad31:
Improvement #1042 

Supported fraction of seconds patterns (S*) in lightFormat


kossnocorp:

PR #1070, opened 2019-01-30
Fix pt-BR translation

TatiTheFreaK:
Since weekdays' names end with a (as in segunda, terça, quarta, quinta, sexta), it makes more sense to use the character ª instead of º


kossnocorp:

saymondamasio (commented):
Can you please leave the first letter capitalized?

Issue #609, opened 2017-11-14
Add constants for max and min dates


graingert:
```js
const maxMillis = (10 ** 8) * 24 * 60 * 60 * 1000;
export const maxDate = new Date(maxMillis);
export const epochDate = new Date(0);
export const minDate =  new Date(-maxMillis);
```

ThePaulMcBride:
I'd be happy to do this if you could explain a little more about what the issue is.

graingert:
It's just for convenience

kossnocorp:
I've reworked the constants:

- `minTime` & `maxTime` instead of `minDate` & `maxDate` as having them as numbers makes more sense.
- Removed `epochDate` as it could be represented by simple `0`

PR #1076, opened 2019-02-01
Add useful constants

cmatiello:
Implements #609 

I wasn't able to run the build script because my file doesn't export a function. 
How can I do it?


kossnocorp:
I've reworked the PR and also integrated constants into the build system:

- `minTime` & `maxTime` instead of `minDate` & `maxDate` as having them as numbers makes more sense.
- Removed `epochDate` as it could be represented by simple `0`

PR #1101, opened 2019-02-24
Update zh-TW locale

jackhsu978:
Update zh-TW (traditional Chinese) locale based on zh-CN (simplified Chinese). The only difference between the two locales is that some characters are slightly different, so I basically copied and pasted everything from zh-CN locale and modified the characters that are different.


kossnocorp:

PR #1102, opened 2019-02-26
Fix document mistake

h-michael:


TwoDCube:
In JavaScript the months are numbered from 0 to 11 (January is 0, February is 1, and so on), so the comment is correct

PR #1105, opened 2019-02-26
Update formatLong for DE locale

saithis:
I updated the de locale to use the format defined in [DIN 5008](https://en.wikipedia.org/wiki/DIN_5008) which is a german standard for writing and layout rules. In the [german date format wikipedia entry](https://de.wikipedia.org/wiki/Datumsformat#DIN_5008) you can see some examples for DIN 5008 formated dates.


kossnocorp:

PR #1107, opened 2019-02-27
Turkish locale v2

dbtek:
This PR is built on top of #1009. So this one obsoletes it.

Apart from that PR by @berkaey:
- Added all tests.
- Improved tr locale for some failing tests on v2.
- Updated ordinal number formatting to keep consistency between format & parse operations.
- Updated parsers that were not working.

PR #1123, opened 2019-03-18
Add Turkish locale for v2 

hlthi:
- All test passed.


kossnocorp:
I'm sorry that I didn't address that in time and it caused you to spend your time. I've decided to merge #1107 over yours because it was based on the work that I already reviewed, published early and got a lot of support.

I'll try to find a way to avoid such conflicts in the future.

PR #1130, opened 2019-03-25
Treat invalid time as InvalidDate in parseISO

stefanprobst:
When a date string with an invalid time is provided to `parseISO` it should return `InvalidDate`.


stefanprobst:
@kossnocorp @leshakoss what do you think about this one?

kossnocorp:

Issue #829, opened 2018-08-03
Add option for using / publishing one package per function


mnquintana:
👋 Thanks so much for all your hard work on this awesome library!

It would be really helpful for applications where disk space is a constraint and a bundler isn't the most viable option (e.g. scripts for Electron apps) to be able to pull in individual packages for only the functions you need. For example, if you only needed to use `distanceInWords()`, you could run something like:

```
npm install @date-fns/distance-in-words
```

and only install the code needed for that one function. For a point of comparison, Lodash offers the option of [per-method packages](https://www.npmjs.com/search?q=keywords:lodash-modularized) in addition to function-per-file-in-one package style `date-fns` uses already.

Let me know if there's any more information I could provide that'd be helpful!

kossnocorp:
@mnquintana Thanks for your kind words! 

I like the idea but I have a concern: the people will start getting duplicated code because of dependencies using different versions, `date-fns` and `@date-fns/distance-in-words`, etc. 

@jdalton could you please tell me what's your experience of having `lodash.assign`, `lodash.clonedeep`, etc. with lodash?

jdalton:
If I could do it again I wouldn't produce `lodash.assign`, `lodash.clonedeep` and others. It's a maintenance pain. Folks end up reaching for them thinking they'll get a smaller bundle but the fact is that plain old `lodash` + `lodash-webpack-plugin` or `babel-plugin-lodash` end up creating smaller packages. If I could go back I'd focus folks on the primary modular package instead offering individual packages.

kossnocorp:
@jdalton good point, thanks! 

@mnquintana I think it's out-of-scope to provide individual packages in order to reduce disk space. It would be great to have a dedicated tool for reducing package size but it's definitely that we can't focus on.

PR #1144, opened 2019-04-08
Update README.md / Added function collection

guysaar223:
I’ve share date-fns as components using bit (https://github.com/teambit/bit), which turns libraries into components without refactoring. Having a function collection adds several benefits:

1. Each function can now be individually installed as a package with NPM and Yarn
2. People can use bit to develop each function right from their own project
3. People can visually explore the collection, search for functions and play with each in an online playground

I think this adds to the modular experience around the library as people can play and work with each function.
I’d be happy to hand over ownership and help update the collection if you like (it’s pretty simple).

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,8 @@ npm install date-fns --save
 yarn add date-fns
 ```
 
+Each function is available as a component in a [bit collection](https://bit.dev/date-fns/date-fns).
+
 ## Docs
 
 [See date-fns.org](https://date-fns.org/) for more details, API,
--- a/README.md
+++ b/README.md
@@ -36,8 +36,6 @@ npm install date-fns --save
 yarn add date-fns
 ```
 
-Each function is available as a component in a [bit collection](https://bit.dev/date-fns/date-fns).
-
 ## Docs
 
 [See date-fns.org](https://date-fns.org/) for more details, API,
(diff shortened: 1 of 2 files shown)


kossnocorp:
Hey Guy! 

Thank you for that, I like the direction, but right now it's not clear how to use date-fns via bit. I like the idea of your project and even tried to integrate it into my workflow (with no luck, tho) and willing to promote it.

However, I would like to avoid advising to use these package as dependencies in libraries. For reasoning, see #829. Given that, I'd like to avoid maintaining the collection myself.

I'll be happy to add bit collection to docs, but not in installation instructions and with disclaimer what it's not the bit packages aren't supposed to be used as library dependencies. I think the best place for that would be a new section "Try date-fns" below the "Installation" in [Getting Started](https://github.com/date-fns/date-fns/blob/master/docs/gettingStarted.md). Also, as I said it's not clear how to use these packages and the doc should contain an elaborate example of how to get started without reading bit docs.

guysaar223:
Hi @kossnocorp 
Thanks  for the feedback!
I’ve edited the PR as requested.

Thanks 🙂

kossnocorp:
Sorry that it took long for me to get back to you and also sorry for closing this PR after forcing you to make changes in it. I understand that I would help you to promote the business, but after trying the proposed flow myself, I didn't find the value in it. It's not easier a bit than installing date-fns from npm, and it would only confuse users. Anyhow, good luck with bit, I hope it will take off!

Issue #584, opened 2017-10-19
Feature Request: Business Days functionality


serle:
Is there any plan to support business day add/subtract calculate from a start and end date. This would need to exclude non-workdays e.g. public holidays and weekdays array.

kossnocorp:
@serle it sounds like a great idea, but I can't promise that we'll work on that in the near future. If you want to help with that, I'll be glad to help you with the guidance.

serle:
@kossnocorp I'm a bit bogged down at the moment, but I have implemented an algorithm on top of the moment.js library, which I'm happy to contribute. I found the standard approach of looping through each day to do a business day test is way too slow for extended durations. i.e. 

1) calculate the number of business days between date1 and date2.
2) add.subtract N business days to dateY

so it required a bit of a specialized approach.

rfoel:
Help still needed?

klompenrunner:
Just like me you might have ended here in search of the additional functionality.

`dateFns.differenceInDays(date1, date2) - dateFns.differenceInCalendarWeeks(date1, date2) * 2`

is meeting my needs for now.

It doesn't exclude holidays days; but excluding those will be such a localization nightmare. Business apps would typically need to follow the businesses holidays anyway.

dkozickis:
This has landed with `date-fns@2.x`. Closing.

younes-io:
Is there a function that sets up the banking holidays ? 
I would expect something like `addBusinessHolidaysToCalendar(holidaysArray)`

BenStirrup:
A feature taking into account holidays would be great !

thedamon:
> This has landed with date-fns@2.x. Closing.

Google brought me here lol.. a name of the new function would be handy. In case anyone else is wondering:

`addBusinessDays`, `subBusinessDays`, `differenceInBusinessDays`

It appears to only use M-F for determining whether something is a business day though so the name "business day" is perhaps misleading.

tonmoydebnath100:
//Here 'form' is a date type variable and 'Days' is numerical variable. 
//subtract only BusinessDays, you can choose any days of a week using isMonday(), isSunday() and etc.
function SubtractBussinessdays(from,Days){
        let i=0,newDate=from;
        Days++;
        while(i<=Days){
          if(isFriday(newDate)){
            Days++;
          }
          newDate=subDays(newDate,1);
          i++;
        }
        return newDate; //it will return the expected date
      }
[...]

PR #1154, opened 2019-04-26
Add new feature addBusinessDays

ThorrStevens:
Should partly solve [#584](https://github.com/date-fns/date-fns/issues/584). Allthough it does not take holidays into consideration, I recon it might still have some value for end users.


ThorrStevens:
PS: Also planning on making a PR for a `getWeekDays` function somewhere in the near future.

This is how I've solved the lack thereof so far:

```javascript
// Get a range or number of weekdays
export const weekDays = (start: DirtyDateType, end: DirtyDateType, asNumber: boolean = false) => {
    if (!start || !end) return 0;
    const businessDays = eachDay(toDate(start), toDate(end)).filter(d => !isWeekend(d));
    return asNumber ? businessDays.length : businessDays;
};
```

Is a function like this something that could be handy in your opinion?

kossnocorp:
Update: I've figured that "business day" is more common term, so please name it `addBusinessDays`.

ThorrStevens:
> Update: I've figured that "business day" is more common term, so please name it `addBusinessDays`.

Updated to `addBusinessDays` as requested 🙌

kossnocorp:
Thanks for the update! I see that the tests fail in `America/Santiago` time zone: https://travis-ci.org/date-fns/date-fns/jobs/543469770#L691

The problem that your tests operate on the DST edge (which is a lucky coincidence) and it causes an hour shift. Look how `eachDayOfInterval` solves this problem: https://github.com/date-fns/date-fns/blob/master/src/eachDayOfInterval/index.js#L94. However, in your case, instead of setting time to 00:00, you have to save hours and reapply in each iteration. I think this will work.

Since your function, is affected by DST, please add a DST test. See:

- https://github.com/date-fns/date-fns/blob/master/scripts/test/dst.sh
- https://github.com/date-fns/date-fns/blob/master/test/dst/eachDayOfInterval/basic.js

Also, please rebase your branch with the fresh master.

Please tell me if you need assistance!

ThorrStevens:
I'm not 100% sure I follow.
I get that the DST is giving problems in some regions, but I'm still unsure of how to fix it.

More specifically, I'm unsure of:
- How to "save" hours in the function
- If I should move the failing test from `./src/addBusinessDays/test.js` to `./test/dst/addBusinessDays/basic.js`

PS: I'll likely only have time to get back to this in a few days.
Any help would be appreciated though.

kossnocorp:
Alrighty, I'll proceed from here 👌

kossnocorp:
You can see the solution for the DST problem: https://github.com/date-fns/date-fns/pull/1154/commits/03bbf4d3609bfe58d038df8e5ba6307e91c8c90c

kossnocorp:

ThorrStevens:
Thanks a bunch for the help and finalisation! 🙌 I’ll try to take some time this weekend to check out how the final pieces actually connect with fixing the tests

hugows:
Its very unexpected that this function doesn't accept negative arguments.

PR #1155, opened 2019-04-30
Correct import example in webpack.md

masives:
The correct way of dynamically load modules with webpack is using import(), not require.https://webpack.js.org/guides/code-splitting/#dynamic-imports


kossnocorp:
Thank you for the PR! Unlike `require`, `import` is async and it requires more complex code example.

If you want, I'll gladly accept a new PR with an elaborate example of how to include locales asynchronously. Also, I'll fix the typo in `master`.

PR #1178, opened 2019-06-03
Incorrect argument in the example

leoafarias:
The current argument causes a weird error when rendering the date


kossnocorp:
Hey, Leo! I'm sorry that the docs have confused you, but this doc is for v2. You can see Getting Started for v1 on the website: https://date-fns.org/v1.30.1/docs/Getting-Started

PR #1186, opened 2019-06-05
Remove zh-TW from outdatedLocales.json

hibearpanda:
Noticed that zh-TW was recently updated.


kossnocorp:
Dang! Well, I suppose I need to ship a new release.

PR #1188, opened 2019-06-06
[tests] update readme

nanomosfet:


kossnocorp:
Thank you for that, but I'm not a fan of badges. A coverage report is a tool for maintainers and contributors not for library users.

Issue #896, opened 2018-09-29
Make p+ tokens consistent


kossnocorp:
Right now p+ tokens (`p`, `pp`, `Pp`, etc.) aren't' consistent and also have incorrect docs.

kossnocorp:
@marnusw, to be honest, I'm not sure why is that, those tokens are merely aliases to more complex ones, so it should be working. I'll check it out.

ericreis:
@kossnocorp I am having the same problem here with `parse` and `p+` tokens. Is there any news regarding this issue?

kossnocorp:
I missed the moment when we fixed it, but according to the snapshots now the docs and output in sync and also corresponds to the standard.

PR #1189, opened 2019-06-06
Add p+ token support to parse function

ericreis:
There's currently no way to parse date strings using long localized date/time formats. This PR aims to add support to `p+` tokens to the `parse` function.

Added patterns:

| Unit                            |Prior| Pattern | Result examples                   | Notes |
|---------------------------------|-----|---------|-----------------------------------|-------|
| Long localized date             |   ? | P           | 05/29/1453                        | 7     |
|                                               |      | PP        | May 29, 1453                      | 7     |
|                                               |      | PPP      | May 29th, 1453                    | 7     |
|                                               |      | PPPP    | Sunday, May 29th, 1453            | 2,5   |
| Long localized time             |   ? | p       | 12:00 AM                          | 7     |
|                                               |      | pp      | 12:00:00 AM                       | 7     |

and their combinations.

_PS: Timezone time parses `ppp` and `pppp` were left out of this PR because the `z+` parsers are not yet implemented in the locales._


ericreis:
@kossnocorp Can you take a look and check if this makes sense? This should close #896, or at least make it more consistent. Also, it should fix https://github.com/Hacker0x01/react-datepicker/issues/1774.

src/parse/index.js
```diff
@@ -397,7 +414,19 @@ export default function parse(
...
   var i
 
-  var tokens = formatString.match(formattingTokensRegExp)
+  var tokens = formatString
+    .match(longFormattingTokensRegExp)
+    .map(function(substring) {
+      var firstCharacter = substring[0]
+      if (firstCharacter === 'p' || firstCharacter === 'P') {
+        var longFormatter = longFormatters[firstCharacter]
+        const a = longFormatter(substring, locale.formatLong, subFnOptions)
```
kossnocorp: Please use `var` instead of `const`, we don't transpile code and this will break UglifyJS.

src/parse/index.js
```diff
@@ -186,6 +191,16 @@ var unescapedLatinCharacterRegExp = /[a-zA-Z]/
...
+ * | Long localized date             |   ? | P       | 05/29/1453                        | 7     |
+ * |                                 |     | PP      | May 29, 1453                      | 7     |
+ * |                                 |     | PPP     | May 29th, 1453                    | 7     |
+ * |                                 |     | PPPP    | Sunday, May 29th, 1453            | 2,5   |
+ * | Long localized time             |   ? | p       | 12:00 AM                          | 7     |
+ * |                                 |     | pp      | 12:00:00 AM                       | 7     |
+ * | Combination of date and time    |   ? | Pp      | 05/29/1453, 12:00 AM              | 7     |
+ * |                                 |     | PPpp    | May 29, 1453, 12:00:00 AM         | 7     |
+ * |                                 |     | PPPpp   | May 29th, 1453 at ...             | 7     |
+ * |                                 |     | PPPPpp  | Sunday, May 29th, 1453 at ...     | 2,5   |
```
kossnocorp: Please update the priority and notes columns:

- Put `NA` instead of `?` as priority is defined by the expanded tokens.
- Add extra note that explains the point above (about priority).
- All new tokens should have `5` note but not `7`.
- `p` and `pp` notes might be joined as in table rows above.

kossnocorp:

PR #1194, opened 2019-06-10
Add new feature differenceInBusinessDays

ThorrStevens:
Sister PR to #1154 

Should partly solve [#584](https://github.com/date-fns/date-fns/issues/584). Allthough it does not take holidays into consideration, it does exclude weekends and should have *some* value for end users.

All in all, it is just a thin wrapper around `eachDayOfInterval`, filtering out the weekends though.

Example usage:

```javascript
var amountOfBusinessDays = differenceInBusinessDays(
    new Date(2018, 0, 1),
    new Date(2019, 0, 1)
)
// => 261 (52 weekends excluded = 104 days skipped)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,19 +13,6 @@ This change log follows the format documented in [Keep a CHANGELOG].
 [See v2 Pre-Releases Change Log](https://gist.github.com/kossnocorp/a307a464760b405bb78ef5020a4ab136)
 for the list of changes made since `v2.0.0-alpha.1`.
 
-### Added
-
-- Added new feature `differenceInBusinessDays`,
-  which counts all days in a range, except for weekends.
-
-  ```javascript
-  var amountOfBusinessDays = differenceInBusinessDays(
-    new Date(2018, 0, 1),
-    new Date(2019, 0, 1)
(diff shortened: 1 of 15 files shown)


kossnocorp:
That sounds great, thank you! Please rebase with master so I can review the diff.

kossnocorp:
I'll finish the PR.

kossnocorp:
Please note that I've changed the function behavior:

- I've swapped the order of arguments, so the later date comes first. Now it's consistent with `differenceInDays` and `differenceInCalendarDays`
- For the sake of consistency, the function now also returns a negative number if the earlier date comes first.
- The function returns `Invalid Date` if any of the arguments is invalid.
- I subtract `1` from the length of the interval because it contains both starting and ending dates. The difference between two dates should not be `2`

kossnocorp:

ro-savage:
@kossnocorp - great work release this. We've been using a custom function with quite similar code in our code base. However, we also take into account 'holidays' by allowing an array of excluded dates to be passed.

Would you be interested in a PR that allows people to add in an array of dates (holidays) to be excluded from differenceInBusinessDays?

PR #1197, opened 2019-06-11
Restore now-dependent functions (closes #1184)

### LINKED ISSUE #1184: Restore now-dependent functions
These functions were removed so we can have "pure" FP. It was done with good intention but caused some users pain. Now I see that we can have both FP and now-dependent functions.

What needs to be done:

- Restore now-dependent functions.
- Adjust the FP generation and ignore these functions.
- Adjust docs to reflect the previous point.

See this gist for the full list of the functions: https://gist.github.com/kitze/b0e72300cee1fb7194e2bd8ccf4091f3

kossnocorp:
See #1184


kossnocorp:

PR #1212, opened 2019-06-19
Fix package name

TatiTheFreaK:


kossnocorp:
You've changed the example for `date-fns-tz`. `date-fns-timezone` example is below.

TatiTheFreaK:
Sorry I didn't explain on the description what I did. (rs)
On the documentation, both examples shows the imported package name as `date-fns-timezone`, but the example I changed, as you said, was for the `date-fns-tz`, so I believe the imported package name on the example is wrong.
Again, sorry I didn't explain at first

kossnocorp:
Oh, I see it know. You're right, sorry I got confused!

kossnocorp:
Thank you for spotting it!

Issue #1221, opened 2019-06-25
Remove locale-independent tokens from locale snapshots


kossnocorp:
Not all tokens depend on locales, and those create unnecessary noise. Also, the "Equals" columns don't make sense and should be removed.

Issue #1220, opened 2019-06-25
Add tokens validation to parse


kossnocorp:
Throw `RangeError` on formats that don't make sense, such as `YYYY-MM-dd`, `YYYY-yyyy`, `HH a`, etc.

Also, see #1210.

Issue #1219, opened 2019-06-25
Rename parse's baseDate argument to backupDate


kossnocorp:
See #1210 for the reasoning.

PR #1224, opened 2019-06-26
Parse function improvements

kossnocorp:
- Stand-alone era token now parses as `1 BC` and `1 AD` instead of `10 BC` and `10 AD`.
-  Fix bug when week year was used to parse Gregorian year and the Gregorian year was used to parse week year.
- Validate parse tokens (see #1220)
- Rename baseDate to backupDate in parse (see #1219)
- Remove locale-independent tokens from locale snapshots (see #1221)

Also, @babel/plugin-transform-parameters and @babel/plugin-transform-destructuring were added to the Babel config.


src/parse/index.js
```diff
@@ -454,6 +466,25 @@ export default function parse(
     var firstCharacter = token[0]
     var parser = parsers[firstCharacter]
     if (parser) {
+      const { incompatibleTokens } = parser
```
leshakoss: Are we using `const` now?

src/parse/index.js
```diff
@@ -454,6 +466,25 @@ export default function parse(
...
+      if (Array.isArray(incompatibleTokens)) {
+        const incompatibleToken = usedTokens.find(
```
leshakoss: `find` is not available for IE, we have to rewrite it with `for` or provide a polyfill

src/parse/index.js
```diff
@@ -454,6 +466,25 @@ export default function parse(
...
+          ({ token }) =>
+            incompatibleTokens.indexOf(token) !== -1 || token === firstCharacter
```
leshakoss: Same for `indexOf`, only available since IE9

src/parse/_lib/parsers/index.js
```diff
@@ -282,12 +283,16 @@ var parsers = {
           )
       }
     },
-    set: function(date, _flags, value, _options) {
+
+    set: function(date, flags, value, _options) {
       // Sets year 10 BC if BC, or 10 AC if AC
-      date.setUTCFullYear(value === 1 ? 10 : -9, 0, 1)
+      flags.era = value
+      date.setUTCFullYear(value === 1 ? 1 : 0, 0, 1)
```
leshakoss: A nitpick: interestingly value here could only be 0 or 1 so the statement could be simplified to `date.setUTCFullYear(value, 0, 1)`

src/parse/index.js
kossnocorp: Yup, several times I've shipped `const`, template literals and arrow functions so I handpicked Babel plugins to transpile those: https://github.com/date-fns/date-fns/blob/master/.babelrc.js#L2-L6

Also, I've added a few in the PR: https://github.com/date-fns/date-fns/pull/1224/commits/e244d27d76a954cdc899f85e9932b0032c6686c7

src/parse/_lib/parsers/index.js
kossnocorp: Wow, I didn't realize that. Good catch!

src/parse/index.js
kossnocorp: God damn, IE!

src/parse/index.js
kossnocorp: Let's raise the bar up to IE11.

src/parse/_lib/parsers/index.js
```diff
@@ -282,12 +283,16 @@ var parsers = {
           )
       }
     },
-    set: function(date, _flags, value, _options) {
+
+    set: function(date, flags, value, _options) {
       // Sets year 10 BC if BC, or 10 AC if AC
```
leshakoss: That comment is outdated :)

kossnocorp: Right!

Issue #1201, opened 2019-06-13
Produce clean CommonJS


kossnocorp:
Right now ESM is transpired with Babel which produces a lot of noise that could be avoided: https://twitter.com/kossnocorp/status/1139113006695092224

One of the options is Rollup. I've experimented with it a bit and seems like it solves the problem with functions but the index files are still noisy: https://gist.github.com/kossnocorp/5257fb234a50413dfb83d4aa67f1b891

The config and the command that produced the desired result:

```js
export default {
  preserveModules: true,
  external: () => true
}
```

```sh
./node_modules/.bin/rollup tmp/package/esm/addDays/index.js --dir tmp/package/addDays --format cjs -c rollup.config.js --no-interop --no-strict
```

kossnocorp:
It's pretty clean now

Issue #1190, opened 2019-06-07
Rework TypeScript definitions


kossnocorp:
- Define modules in the corresponding type definition instead of re-exporting from the root:

  ```ts
  import { closestTo } from 'date-fns'
  export default closestTo
  ```

  →

  ```ts
  export default function closestTo(...): ...
  ```

- Make index definitions re-export from modules:

  ```ts
  interface dateFns {
    addDays(date: Date | number, amount: number): Date

    addHours(date: Date | number, amount: number): Date

    ...
  }
  ```

  →

  ```ts
  export { default as addDays } from './addDays'
  export { default as addHours } from './addHours'
  ...
  ```

nvuhung:
I also have the problem with typescript definitions of date-fns.  I'm working with locale and same like it is missing some locales (Eg: vi)

kossnocorp:
@nvuhung that is an unrelated problem. Please open an issue with a detailed description of the problem, the version you use and steps to reproduce.

yacinehmito:
I'm up for solving this while doing https://github.com/date-fns/date-fns/issues/1226
It might be necessary anyway.

kossnocorp:
I love the idea!

Issue #1117, opened 2019-03-12
Will there be an UMD or AMD version published to npm?


taoqf:
I did not find a browser version in npm? Did I miss something?

kajmagnus:
And I want to generate my own UMD or [no-bundle, just a global object] bundles, one per language and as custom built, as small as possible for one's use case — I asked about that here: https://github.com/date-fns/date-fns/issues/1484

PR #1234, opened 2019-06-29
Migrate to TypeScript

### LINKED ISSUE #1226: Get rid of Flow type checking; consider TypeScript
Every time Flow fails a build, and it's always a random failure. We never got any value from it. 

For example https://github.com/date-fns/date-fns/pull/1224/commits/a89e77ddf69faffa2b36defa946c89d210c6b2a9

yacinehmito:
_Disclaimer: This PR is very much WIP. I'll probably restart from scratch once everything is ironed out. Indeed, the current setup has many caveats. See at the end._

This PR should close #1226 and #1190.

The idea is to typecheck the codebase using TypeScript instead of Flowtype. It poses several challenges as the build process of this library is quite involved. There are many moving parts so TypeScript, as typically used, is hard to fit in.

Things done in the PR:
- **Converted all JS lib files to TypeScript.** By lib files I mean actual library code, not the tests nor the benchmarks. I tried to sensibly type most of the public API but I didn't go all the way through so there are a lot of `any` types. It compiles in TypeScript's strictest settings.
- **Configured TypeScript so that it emits JS files in the same place where there are TS files.** This is the default behaviour. It allows for minimal changes to the current build process (although we might want to change it further). It has the big caveat of doubling the amount of checked out source code.
- **Ran prettier on all files.** I did a lot of regex-based refactoring and wanted to make sure that I could go back to a consistent style. Therefore I ran prettier on all files, both before and after the changes.
- **Introduced @typescript-eslint.** ESLint does not support TypeScript by default so I had to change the config. I went for the recommended stuff, but you might want to customize it to your liking as it adds new rules.
- **Commended out TS typings generation.** This used to create declaration files calling to `typings.d.ts`. Now we have the proper declaration files generated in place. This goes a long way towards #1190. What is left to do is the opposite: generate `typings.d.ts` from these declaration files. Microsoft's [API extractor](https://api-extractor.com/) does this very well.

I don't think this PR deserves to be reviewed yet because the approach is probably the wrong one. I tried to minimize the changes to perform on the build process but I only made it more complicated. Even worse, `scripts/build/package.sh` has stopped working so more work would be needed.

I suggest instead that we do an overhaul. This is a proposed architecture:
- Code and tests all in `src/`, as today. Difference is that it would be all TypeScript, without any `.d.ts` nor `.flow` files.
- A first build step dumps in `build/` the content of `src/` compiled to JavaScript, with the declaration files. We could have two TypeScript projects: one for the library code, that generates JS files with declaration files, and one for the test code, that only generates JS files. We can leverage project references to make the test code depend on the library code. An added benefit is that it becomes impossible to mistakenly import test code in the library code.
- With all the JS code in `build/`, we can run tests and benchmarks from `build/` instead of `src/`, as we do today.
- A second build step copies the library code and the declaration files from `build/` to `dist/` and does the following:
  - it generates `typings.d.ts` using [API extractor](https://api-extractor.com/)
  - it generates `.flow` files from the `.d.ts` files (the "how" is left to be figured out)
  - it builds various bundles using Rollup (this would also solve solve #1117 and #1201)

The big change is that it we wouldn't check out JavaScript, Flow and declaration files in source control. It would only exist while building. This may affect some other things I didn't notice yet; if so, let me know.

What do you think of this proposition? Of course I'm willing to do the work and document it. Let me know if you have any remarks.


NaridaL:
Any progress on this?

It seems like it is difficult to handle/review all in one go. Maybe we could split it up in multiple PRs; for example.

1. apply prettier to the entire code base
2. set up TS build step compiling (not the tests at first) JS files. We could then gradually migrate the files to TS.
3. ...

yacinehmito:
@NaridaL I did not carry on with this. Indeed this must be "split", but better than that I think we should have steps that follow the proposed architecture. I'm closing the PR.

PR #1237, opened 2019-07-02
providing way to disable logging in postinstall script 

joeycozza:
core-js provides a way to quiet the logging in its post install script. It seems like a good idea.

https://github.com/zloirock/core-js/blob/master/packages/core-js/scripts/postinstall.js

### DIFF
--- a/docs/printV2Notice.js
+++ b/docs/printV2Notice.js
@@ -1,4 +1,14 @@
-console.log(`
+const env = process.env
+const CI = is(env.CI)
+const DISABLE_OPENCOLLECTIVE = is(env.DISABLE_OPENCOLLECTIVE)
+const SILENT = ['silent', 'error', 'warn'].includes(env.npm_config_loglevel)
+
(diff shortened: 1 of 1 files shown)


AdamBrodzinski:
@kossnocorp @joeycozza  This commit breaks on the post install on older legacy versions of node (by using `.includes`). I would suggest using a different method. React Datepicker is (wrongly) using the alpha/beta versions and a recent deploy to Heroku crashes the build (node v4.9.0).

bduff9:
Yep, just ran into this exact issue where it broke a production deploy because includes is not a function.  Rolling back to beta.1 resolves it but it would be awesome if we could utilize the trusty indexOf instead for this.

kossnocorp:
Thank you for reporting, I'll simply remove the notice: https://github.com/date-fns/date-fns/issues/1267

kossnocorp:
I removed the notice in `v2.0.0-beta.5`

PR #1241, opened 2019-07-03
Fixed days of the week and capitalization of months

hugows:


kossnocorp:
I asked around and seems like in Portuguese (at least in Brazilian) names of months and days of weeks aren't capitalized in the middle of a sentence like in English "we'll meet in January at a Sunday".

PR #1233, opened 2019-06-29
fix(typings): export interface

ricardogobbosouza:

PR #1256, opened 2019-07-18
Revert "fix(typings): export interface"

kossnocorp:
Reverts date-fns/date-fns#1233

The change makes TypeScript fail with an obscure error: https://travis-ci.org/date-fns/date-fns/jobs/559946272#L689

@ricardogobbosouza if you want to see the global interface exported, could you please redo it again, but this time run `./scripts/build/build.sh` to update the typings and then make sure that the smoke tests pass (`./scripts/test/smoke.sh` or `cd examples/typescript && yarn build && yarn test`)


ricardogobbosouza:
@kossnocorp is working here https://travis-ci.org/date-fns/date-fns/builds/552215342

kossnocorp:
@ricardogobbosouza yup, because that is just script change, but the result of the script wasn't committed.

PR #1257, opened 2019-07-18
Add a default formatStr of 'PP'

Satyam:
Add a default `formatStr` of `'PP'` (or any of its variants) to the example custom formatting function so that it defaults to whichever format the locale has for dates.  Show the example output without `formatStr` set.


kossnocorp:
Thank you a lot!

PR #1259, opened 2019-07-22
Fix formatters for "D MMM", "DD MMM", "Do MMM"

yepstepz:
see #1260


kossnocorp:
Sorry, I have to freeze v1 and close all PRs to it. It became a very challenging task to release v1 because of differences between v1 and v2. Moving forward, we'll be more considerate while upgrading major versions. Sorry for the troubles.

PR #1262, opened 2019-07-22
replaced lodash by Ramda

6ewis:


stunaz:
lol... why?

6ewis:
@stunaz i felt it was more accurate. Ramda is functional as opposed to lodash, lodash was inspired and created lodash-fp. https://twitter.com/kyleshevlin/status/1004610055721779200?lang=en

kossnocorp:
Even though it might be accurate, Lodash is far more popular. The purpose of the comparison is to give an idea of what to expect from the library, so Lodash is more appropriate.

Issue #1268, opened 2019-07-29
differenceInBusinessDays returns incorrect value when a given date falls on a weekend


RudyRed:
When the given dates fall on a weekend the differenceInBusinessDays can be incorrect

eg. 
```
differenceInBusinessDays(new Date(2019, 6, 28), new Date(2019, 6, 20));
//                   Sun July, 28th 2019        Sat July 20th 2019
// expected: 5
// actual: 4
```

also: 
```
differenceInBusinessDays(new Date(2019, 6, 28), new Date(2019, 6, 27));
//                   Sun July, 28th 2019        Sat July 27th 2019
// expected: 0
// actual: -1
```

There is also a discrepancy between the docs and the test for this. in the docs the following is stated (from the [differenceInBusinessDays/index.js](https://github.com/date-fns/date-fns/blob/master/src/differenceInBusinessDays/index.js))
```
 * var result = differenceInBusinessDays(
 *   new Date(2014, 6, 20),
 *   new Date(2014, 0, 10)
 * )
 * //=> 136
```
however under the [differenceInBusinessDays/test.js](https://github.com/date-fns/date-fns/blob/master/src/differenceInBusinessDays/test.js) it shows the following for the same input:
```
describe('differenceInBusinessDays', function() {
  it('returns the number of business days between the given dates, excluding weekends', function() {
    var result = differenceInBusinessDays(
      new Date(2014, 6 /* Jul */, 20),
      new Date(2014, 0 /* Jan */, 10)
    )
    assert(result === 135)
  })
```
fyi the docs are correct and the value should be 136

The reason for the error is that the code is assuming that that both the start and end dates are always included in the array eachDayOfInterval ([differenceInBusinessDays/index.js](https://github.com/date-fns/date-fns/blob/master/src/differenceInBusinessDays/index.js)))
```
  var daysOfInterval = eachDayOfInterval(interval)
  var difference = daysOfInterval.filter(function(day) {
    return !isWeekend(day)
  })
  // Substruct 1 because interval contains both starting and ending dates
  var result = sign * (difference.length - 1)
```

however weekends are already filtered out therefore both the start and ending dates might not be included.

I am working on a PR for this, sorry for the info overload, this is my first issue/PR to a package so I'm not sure what is standard 😃

kossnocorp:
Thank you for reporting that, good catch!

IgnatG:
I'm still experiencing this issue in 2.28.0.
Do you have any updates on this?

[...]

PR #1269, opened 2019-07-29
fix: incorrect response for dates falling on weekends (differenceInBusinessDays)

RudyRed:
#1268


RudyRed:
- Fixes the issue by checking if the left date (which should be the larger and the end of the range) fall on a weekend, only if it falls on a weekend do you subtract it from the total work days.

- fixed old tests to only use input dates that arent weekends

- created new test for inputs that fall on weekends

Issue #1276, opened 2019-08-01
isBefore and isAfter wrong way around according to documentation 


Jordan-Hall:
const as = isBefore(new Date(1989, 6, 10), new Date(1987, 1, 11));
Expected: //=> false
Actual: //=> true

const a = isAfter(new Date(1989, 6, 10), new Date(1987, 1, 11));
Expected: //=> true
Actual: //=> false

kossnocorp:
Hi! Are you sure you did not import `isBefore` from 'date-fns/isAfter` and vice versa? I'm getting the right results for your examples:

```js
import { isAfter, isBefore } from 'date-fns'

isBefore(new Date(1989, 6, 10), new Date(1987, 1, 11))
//=> false

isAfter(new Date(1989, 6, 10), new Date(1987, 1, 11))
//=> true
```

Also, check the tests:

- https://github.com/date-fns/date-fns/blob/master/src/isBefore/test.js
- https://github.com/date-fns/date-fns/blob/master/src/isAfter/test.js

PR #1277, opened 2019-08-01
Before and after logic was back to front

Jordan-Hall:
Issue: #1276 

Documentation compared with the actual result is incorrect. isBefore is actually isAfter and isAfter was isBefore

### DIFF
--- a/src/isAfter/index.js
+++ b/src/isAfter/index.js
@@ -31,5 +31,5 @@ export default function isAfter(dirtyDate, dirtyDateToCompare) {
 
   var date = toDate(dirtyDate)
   var dateToCompare = toDate(dirtyDateToCompare)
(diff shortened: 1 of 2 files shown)


kossnocorp:
See my comment: https://github.com/date-fns/date-fns/issues/1276#issuecomment-517743428

Jordan-Hall:
@kossnocorp  yes

import {
  subYears,
  isBefore,
  isAfter,
  getYear,
  differenceInYears
} from 'date-fns/fp';

kossnocorp:
@Jordan-Hall you use FP submodule where the order of arguments is reversed, please see the doc: https://date-fns.org/v2.0.0-beta.3/docs/FP-Guide

PR #1286, opened 2019-08-09
addBusinessDays negative amount fix, performance fixes for addBusinessDays and differenceInBusinessDays

RudyRed:
Unlike the addDays function, addBusinessDays does not handle a negative amount and same given date.

```
addBusinessDays(new Date(2014, 8 /* Sep */, 15), -10)
// expected:  Sept 1st 2014
// actual: Sept 15 2014
```

addBusinessDays also loops over the entire given amount linearly e.g. is 1,000,000 is given as the amount than the loop runs 1,000,00 times. This behaviour is fixed using division in the PR.
```
  date.setDate(date.getDate() + toInteger(amount / 5) * 7)
```

differenceInBusinessDays also had the same performance issue where an array of all the dates in a range was being created and filtered through due to the [following line](https://github.com/date-fns/date-fns/blob/master/src/differenceInBusinessDays/index.js#L53:L56)
```
  var daysOfInterval = eachDayOfInterval(interval)
  var difference = daysOfInterval.filter(function(day) {
    return !isWeekend(day)
  })
```
this performance issue is fixed below
```
  var weeks = toInteger(calendarDifference / 7)

  var result = weeks * 5
  dateRight = addDays(dateRight, weeks * 7)
```
remaining days not part of a full week are accounted for afterwards

### DIFF
--- a/src/addBusinessDays/index.js
+++ b/src/addBusinessDays/index.js
@@ -33,12 +33,17 @@ export default function addBusinessDays(dirtyDate, dirtyAmount) {
   if (isNaN(amount)) return new Date(NaN)
 
   var hours = date.getHours()
-  var numWeekDays = 0
-  while (numWeekDays < amount) {
-    date.setDate(date.getDate() + 1)
-    date.setHours(hours)
-    if (!isWeekend(date)) numWeekDays++
+  var sign = amount < 0 ? -1 : 1
+
+  date.setDate(date.getDate() + toInteger(amount / 5) * 7)
+  amount %= 5 // to get remaining days not part of a full week
+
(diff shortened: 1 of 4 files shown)


RudyRed:
`this.timeout(500 /* 500 ms test timeout */)` is returning a not a function error. It works properly in the local test suite ( all `yarn test` pass), any advice on the differences of the 2 testing environments and how I would be able to fix that test and ensure it will work with TravisCI without just pushing it up and hoping that it works?

kossnocorp:
Thank you again, I've shipped this PR as `v2.0.0-beta.5`

Issue #1290, opened 2019-08-15
Fix exports in TS typings


kibertoad:
2.0.0-beta.4 with TypeScript 3.5.2 throws 12 TS errors when trying to use date-fns types during compilation:

```
../node_modules/date-fns/typings.d.ts:1131:3 - error TS2309: An export assignment cannot be used in a module with other exported elements.

1131   export = format
       ~~~~~~~~~~~~~~~

../node_modules/date-fns/typings.d.ts:1461:3 - error TS2309: An export assignment cannot be used in a module with other exported elements.

1461   export = max
       ~~~~~~~~~~~~

../node_modules/date-fns/typings.d.ts:1466:3 - error TS2309: An export assignment cannot be used in a module with other exported elements.

1466   export = min
       ~~~~~~~~~~~~

../node_modules/date-fns/typings.d.ts:1471:3 - error TS2309: An export assignment cannot be used in a module with other exported elements.

1471   export = parse
       ~~~~~~~~~~~~~~

../node_modules/date-fns/typings.d.ts:1901:3 - error TS2309: An export assignment cannot be used in a module with other exported elements.

1901   export = format
       ~~~~~~~~~~~~~~~

../node_modules/date-fns/typings.d.ts:2231:3 - error TS2309: An export assignment cannot be used in a module with other exported elements.

2231   export = max
       ~~~~~~~~~~~~

../node_modules/date-fns/typings.d.ts:2236:3 - error TS2309: An export assignment cannot be used in a module with other exported elements.

2236   export = min
       ~~~~~~~~~~~~

../node_modules/date-fns/typings.d.ts:2241:3 - error TS2309: An export assignment cannot be used in a module with other exported elements.

2241   export = parse
       ~~~~~~~~~~~~~~

../node_modules/date-fns/typings.d.ts:2671:3 - error TS2309: An export assignment cannot be used in a module with other exported elements.

2671   export = format
       ~~~~~~~~~~~~~~~

../node_modules/date-fns/typings.d.ts:3001:3 - error TS2309: An export assignment cannot be used in a module with other exported elements.

3001   export = max
       ~~~~~~~~~~~~

../node_modules/date-fns/typings.d.ts:3006:3 - error TS2309: An export assignment cannot be used in a module with other exported elements.

3006   export = min
       ~~~~~~~~~~~~

../node_modules/date-fns/typings.d.ts:3011:3 - error TS2309: An export assignment cannot be used in a module with other exported elements.

3011   export = parse
[...]

PR #1291, opened 2019-08-15
Add TypeScript definition file linting and building scripts

kibertoad:
Add two new scripts: one will run ESLint TypeScript linting on typing file (all failing rules are disabled for now, can be enabled if typing generation script is adjusted in the future), and another one attempts to build dummy TS file with dependency on date-fns typings (which currently fail due to #1290)

Note that this is not automatically executed in CI and will not fail builds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ docs.json
 .idea
 .vscode
 coverage
+test/ts/*.js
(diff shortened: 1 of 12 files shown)


kossnocorp:
Thank you for the PR! I've moved build test to examples as they are part of smoke tests and also enabled them.

kibertoad:
@kossnocorp Appreciate it! Note that I believe some obsolete changes ended up being merged: please see https://github.com/date-fns/date-fns/pull/1299

PR #1293, opened 2019-08-15
fix(format): convert null date to empty string

fkm:
Currently, the function returns the 1st of January 1970.


kossnocorp:
Thank you for raising the concern regarding the behavior, but your proposal isn't consistent with the rest of date-fns. I've opened a PR that makes `format` and `lightFormat` throw an error: https://github.com/date-fns/date-fns/pull/1300

fkm:
Sounds reasonable. Thank you for your reply!

PR #1297, opened 2019-08-19
fix(typings): export interface

ricardogobbosouza:
@kossnocorp  now it works #1256


leshakoss:
As we are going TS-first in v3 soon, this change will be overridden by new typing generator

PR #1298, opened 2019-08-20
Fix format example in README

bertdeblock:
I think the format `yyyy-MM-dd` in the README example is incorrect and should be `YYYY-MM-DD`.


kossnocorp:
The example in README is for v2 that now is out.

bertdeblock:
Woops, ma bad!

PR #1299, opened 2019-08-20
Cleanup after #1291

kibertoad:
After some of the changes in the PR, I believe these lines are obsolete.


kossnocorp:
Thank you!

PR #1300, opened 2019-08-20
Throw TypeError if date argument in format functions is invalid

kossnocorp:
Now `format` and `lightFormat` would throw `TypeError` if the first argument is not a number or `Date`.

See PR with raised concern regarding the behavior: https://github.com/date-fns/date-fns/pull/1293

### DIFF
--- a/src/format/index.js
+++ b/src/format/index.js
@@ -1,4 +1,5 @@
 import isValid from '../isValid/index.js'
+import isDate from '../isDate/index.js'
 import defaultLocale from '../locale/en-US/index.js'
 import subMilliseconds from '../subMilliseconds/index.js'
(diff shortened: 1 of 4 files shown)


kossnocorp:
After discussion with @leshakoss we decided to keep the current behavior where `null` results in the 1970 date. 

First of all, we believe that type-safety should be provided by TypeScript or Flow, so it will ensure that the build size is minimal. If we going to add type checking it would blow the size up dramatically. There's a lot of corner cases where you can't just use `instanceof` (i.e. when object is coming from another iframe). 

Secondly, it consistent with the JavaScript behavior:

```js
new Intl.DateTimeFormat('en-US').format(null)
//=> "1/1/1970"
```

cc @fkm

PR #1311, opened 2019-08-24
Fix format after migrating V2.

Dallas62:
The translation was:

"Samedi 24ième août"

When we expect "Samedi 24 août"

I removed "Do" because I'm not sure if we should apply it in this case, no sense for me with V2.


Dallas62:
This PR is not working, but there is still a bug, formatters are not loaded right :/

PR #1316, opened 2019-08-27
Fix `isDate` to return false for Invalid Dates

jprodrigues70:
Currently if I use

```javascript
isDate (new Date ('aaaaaaa'))
``` 
or

```javascript
isDate(new Date('22/22/2222'))
```
it will return `true`, because when `new Date(someArgument)` returns `Invalid Date` it still a `instance of` `Date`. So I check if the argument passed to `isDate` contains an valid miliseconds value.

### DIFF
--- a/src/isDate/index.js
+++ b/src/isDate/index.js
@@ -42,8 +42,8 @@ export default function isDate(value) {
   }
 
   return (
-    value instanceof Date ||
+    (value instanceof Date ||
     (typeof value === 'object' &&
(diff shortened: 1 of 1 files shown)


leshakoss:
Since `Invalid date` is an instance of `Date` I think it makes perfect sense that `isDate(new Date(NaN))`. It doesn't say that it is a valid date, just that it is a date.

leshakoss:
Anyway, it is a breaking change so we can't accept it 🙂 But thanks for the contribution! ⭐️

PR #1395, opened 2019-08-30
Fix comments for date comparison examples

siakaramalegos:


leshakoss:
[In JS](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Date/getMonth) months are 0 indexed, so `0` is January, `1` is February and `11` is December :)

Issue #1388, opened 2019-08-28
Feat: multi set function to set multiple parts of a date


benneq:
I think it would be nice to have a function, that can set multiple values at once.

So instead of doing this:
```
setSeconds(setMinutes(setHours(myDate, 12), 12), 12);
```

Something like this would be nice:
```
set(myDate, { hours: 12, minutes: 12, seconds: 12 });
```

PR #1398, opened 2019-08-31
Add new set function

SneakyFish5:
Fix https://github.com/date-fns/date-fns/issues/1388

Moment.js also has a function like this one so I thought it'd be good to add, and may be easier for some people to use.
This is my first contribution so do let me know if I did anything wrong and I'll be happy to fix it.


SneakyFish5:
I keep getting flow errors if I try to implement tests to check for the types. Are these tests/RangeErrors unneeded then?

src/set/index.js
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * @name set
+ * @category Common Helpers
+ * @summary Set the object to a given date
+ *
+ * @description
+ * Set the object to a given date
+ *
+ * @param {Date|Number} date - the date to be changed
+ * @param {Object} [options] - an object with options
```
leshakoss: `options` is a special argument in date-fns and is used to store all optional parameters. It has special behaviour for FP functions: if a function `x` [has this argument](https://github.com/date-fns/date-fns/blob/master/scripts/build/docs.js#L283) then it recieves two corresponding FP functions: `fp/x` and `fp/xWithOptions` (reflecting `options`' optional nature)

So, please rename it to something else like `values`

leshakoss: Also, it definitely shouldn't be an optional argument.

src/set/index.js
```diff
@@ -0,0 +1,72 @@
...
+import toInteger from '../_lib/toInteger/index.js'
+import setMonth from '../setMonth/index.js'
+
+/**
+ * @name set
+ * @category Common Helpers
+ * @summary Set the object to a given date
+ *
+ * @description
+ * Set the object to a given date
```
leshakoss: Needs a more detailed description. Set what exactly? What happens if some values are not provided?

Also, probably needs some note about bundle size: that this function internally does not use `setX` functions from date-fns so you probably want to include either `set` or `setX` but not both at the same time if you're concerned about the bundle size.

src/set/index.js
```diff
@@ -0,0 +1,72 @@
+import toDate from '../toDate/index.js'
+import toInteger from '../_lib/toInteger/index.js'
+import setMonth from '../setMonth/index.js'
+
+/**
+ * @name set
+ * @category Common Helpers
+ * @summary Set the object to a given date
```
leshakoss: `Set date values to a given date`

src/set/index.js
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * @description
+ * Set the object to a given date
+ *
+ * @param {Date|Number} date - the date to be changed
+ * @param {Object} [options] - an object with options
+ * @param {Number} [options.years=0] - the number of years to be set
```
leshakoss: I think property names should be the following: `year`, `month`, `date`, `hours` (etc, plural from now on). This is the convention used by JS standard lib: [`year` and `month`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setMonth) in the singular, then collective noun [`date`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setDate) instead of "days", then [everything else in the plural](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setMinutes)

src/set/index.js
```diff
@@ -0,0 +1,72 @@
...
+
+  var date = toDate(dirtyDate)
+
+  // Check if date is Invalid Date because Date.prototype.setFullYear ignores the value of Invalid Date
+  if (isNaN(date)) {
+    return new Date(NaN)
+  }
+
+  if (years > 0) date.setFullYear(years)
+  if (months > 0) date = setMonth(date, months)
```
leshakoss: Month indexing in JS starts with 0 (= January) so this function fails in that case. The correct solution would be to check for nullability and setting the value at the same time:

```js
if (hours != null) {
  date.setHours(toInteger(hours))
}
```

src/set/index.js
```diff
@@ -0,0 +1,72 @@
...
+  if (arguments.length < 2) {
+    throw new TypeError(
+      '2 arguments required, but only ' + arguments.length + ' present'
+    )
+  }
+
+  var options = dirtyOptions || {}
+
+  // check if properties of object are numbers and not NaN
+  var isNumbers = Object.values(options).every(x => typeof x === 'number')
```
leshakoss: `Object.values` is not implemented in IE so we need to either transpile it during the compilation and I am not sure if we're doing this or not. @kossnocorp?

src/set/index.js
```diff
@@ -0,0 +1,72 @@
...
+  var hasNaN = Object.values(options).every(x => isNaN(x))
+
+  if (!isNumbers) {
+    throw new RangeError('The options object must only contain numbers')
```
leshakoss: Everywhere else in date-fns we treat number-like objects (like string `"5"` for example) like numbers, that is the point of that `toInteger` function. Please remove this exception

src/set/index.js
```diff
@@ -0,0 +1,72 @@
...
+  } else if (hasNaN) {
+    throw new RangeError('The options object must not contain a NaN')
```
leshakoss: JS std lib usually returns NaN in math operations if any of the parameters were NaN and we'd like to preserve that behaviour:

```js
const date = new Date()
date.setHours(NaN)
date //=> Invalid Date
```

Please remove this exception

src/set/index.js
leshakoss: (also notice how momentjs follows the same convention)

src/set/index.js
```diff
@@ -0,0 +1,72 @@
+import toDate from '../toDate/index.js'
+import toInteger from '../_lib/toInteger/index.js'
+import setMonth from '../setMonth/index.js'
```
leshakoss: Why do we need to import `setMonth` for that?

leshakoss:
> I keep getting flow errors if I try to implement tests to check for the types. Are these tests/RangeErrors unneeded then?

You need to let flow know about the intended type mismatch: https://github.com/date-fns/date-fns/blob/master/src/areIntervalsOverlapping/test.js#L166

src/set/index.js
SneakyFish5: Done 👍

src/set/index.js
SneakyFish5: Done 👍

src/set/index.js
SneakyFish5: It was a bug I had but I worked on it a bit to get fixed so I've removed this in the latest commit.

src/set/index.js
SneakyFish5: I updated the description as best as I could but I'm sure it could still use improvement. If you have any other ideas I am certainly open to help with it.

src/set/index.js
SneakyFish5: Ah yes that is my bad, I forgot that brackets meant the argument was optional. I've changed the parameter to values and removed the brackets, which should've fixed the `fp` problem.

src/set/index.js
SneakyFish5: Done 👍

src/set/index.js
SneakyFish5: I removed the exceptions anyways due to comments down the line, so this is no longer needed.

src/set/index.js
SneakyFish5: Done 👍

src/set/index.js
SneakyFish5: Done 👍

src/set/index.js
SneakyFish5: Done 👍

src/set/index.js
```diff
@@ -0,0 +1,83 @@
...
+ * @param {Number} values.hours - the number of hours to be set
+ * @param {Number} values.minutes - the number of minutes to be set
+ * @param {Number} values.seconds - the number of seconds to be set
+ * @param {Number} values.milliseconds - the number of milliseconds to be set
+ * @returns {Date} the new date with options set
+ * @throws {TypeError} 2 arguments required
+ * @throws {RangeError} The options object must only contain number properties
+ * @throws {RangeError} The options object must not contain a NaN
+ *
+ * @example
```
leshakoss: :+1:

src/set/index.js
```diff
@@ -0,0 +1,83 @@
...
+  var milliseconds =
+    values.milliseconds == null ? 0 : toInteger(values.milliseconds)
+
+  // Check if date is Invalid Date because Date.prototype.setFullYear ignores the value of Invalid Date
+  if (isNaN(date)) {
+    return new Date(NaN)
+  }
+
+  if (year > 0) date.setFullYear(year)
+  if (month != null) date.setMonth(toInteger(month), days)
```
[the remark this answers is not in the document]
SneakyFish5: Fixed, however I needed to keep days for the setMonth otherwise it would create an error where the resulting month is a month higher than it is supposed to be. Maybe my implementation is wrong, however I saw this happening date-fns `setMonth` function as well, so I opted to leave it in there unless you have any objection.

src/set/index.js
```diff
@@ -0,0 +1,83 @@
...
+ * @param {Number} values.year - the number of years to be set
+ * @param {Number} values.month - the number of months to be set
+ * @param {Number} values.date - the number of days to be set
+ * @param {Number} values.hours - the number of hours to be set
+ * @param {Number} values.minutes - the number of minutes to be set
+ * @param {Number} values.seconds - the number of seconds to be set
+ * @param {Number} values.milliseconds - the number of milliseconds to be set
+ * @returns {Date} the new date with options set
+ * @throws {TypeError} 2 arguments required
+ * @throws {RangeError} The options object must only contain number properties
```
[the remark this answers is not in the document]
SneakyFish5: Done 👍

src/set/test.js
```diff
@@ -0,0 +1,79 @@
...
+    var result = set(new Date(2014, 8, 20, 12, 58), { seconds: 60 })
+    assert.deepEqual(result, new Date(2014, 8, 20, 12, 59))
+  })
+  it('milliseconds turns into seconds', function() {
+    var result = set(new Date(2014, 8, 20, 12, 58, 30), { milliseconds: 1000 })
+    assert.deepEqual(result, new Date(2014, 8, 20, 12, 58, 31))
+  })
+  it('throws TypeError exception if passed less than 2 arguments', function() {
+    assert.throws(set.bind(null), TypeError)
+  })
```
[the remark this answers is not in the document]
SneakyFish5: I did all of these except for the first one. The test I wrote is like this:

```
it('if any value in values is NaN', function () {
    var result = set(new Date(2014, 8), { year: NaN })
    assert.deepEqual(result, new Date(NaN))
  })
```

However, when I run the tests this is my result.
```
 assert.deepEqual(result, new Date(NaN))
                       |       |        |    
                       |       |        NaN  
                       |       new Date(null)
                       new Date(null) 
```

I'm not quite sure why this happens but maybe you could give me some insight?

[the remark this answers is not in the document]
leshakoss: It's a property of NaN :)

```js
NaN === NaN
//=> false
```

You can check for Invalid Date like this:
```js
assert(isNaN(date))
```

src/set/index.js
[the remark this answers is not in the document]
leshakoss: Not sure what do you mean. I checked the branch out locally and removed `daysForMonth` and tests pass, also in the console it looks normal:

```
> const set = require('./src/set')
> set(new Date(), {month: 0}).toString()
'Sun Jan 06 2019 13:24:19 GMT+0100 (Central European Standard Time)'
```

What am I missing?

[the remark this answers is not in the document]
SneakyFish5: Huh, I'm not too sure. When I run the tests on my local with `daysForMonth` removed I end up failing this test, and using `daysForMonth` as the extra parameter seemed to fix it.

```
    × set the date with every option
      Chrome 76.0.3809 (Windows 10.0.0)
    AssertionError:   # src\set\test.js:18
      
      assert.deepEqual(result, new Date(2014, 8, 20, 12, 12, 12, 12))
                       |       |
                       |       new Date("2014-09-20T17:12:12.012Z")  
                       new Date("2014-10-20T17:12:12.012Z")
```
Maybe I've messed up something on my end?

src/set/test.js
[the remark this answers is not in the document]
SneakyFish5: Ah I see, thank you very much. Just fixed that.

leshakoss:
@SneakyFish5 I removed `daysForMonth` from your branch just to let travis do the full test run through the time zones, let's see what the bug is

SneakyFish5:
It failed, but I'm not quite sure the reason behind it failing or not/why it's only for that test.

leshakoss:
@SneakyFish5 ok, I added an import of `setMonth` like you did before and seems like the bug is gone. Let's see what will travis say about this :)

Also done some minor stylistic changes to make the code a bit more uniform with the rest of date-fns.

SneakyFish5:
Awesome, thank you @leshakoss for all the help. I squashed the commits for cleanup.

PR #1420, opened 2019-09-12
(fix) Wrong translation of 'yesterday' for sl locale

VesterDe:
@mzgajner did a great job with the translation, but made 1 small mistake.

I'm surprised there weren't more, given how strange our language is sometimes :)


mzgajner:
Can confirm that this fix is correct as a native speaker :)

@VesterDe kudos for spotting this! Judging by the CI output I think you'll have to run `yarn locale-snapshots` and commit the updated `snapshot.md` file as well.

PR #1430, opened 2019-09-18
Fix French parsing issues with june and august

lovelovedokidoki:
There are two typos in the parseMonthPatterns for French locale causing the parser to erroneously parse july as june and not parse august at all.

The pre-commit script wants to add newlines in between the list items, so changes might be hard to spot, but they are very few: It's just `/^jui/i` -> `/^juin/i` and `/^au/i` -> `/^ao/i`.


Lakston:
Thanks for your contribution !
Merci pour ta contribution :)

PR #1451, opened 2019-09-26
little improvement

anatoliy8493:


kossnocorp:
I appreciate the effort and like the direction, but the typing file is automatically generated so your changes will disappear on the very next build.

PR #1459, opened 2019-10-05
Introduce property-based tests

Djaler:
More info about property based tests: https://github.com/dubzzz/fast-check/blob/master/documentation/1-Guides/HandsOnPropertyBased.md

They can be run with other tests by ` ./scripts/test/node.sh`

And there are already sereral failed cases, for example:
```
const date = new Date("1919-07-01T00:31:40.999Z")

const isoPattern = `yyyy-MM-dd'T'HH:mm:ss.SSSXXX`

const formatted = format(date, isoPattern)
console.log(formatted) // '1919-07-01T04:31:59.999+04:00'
const formattedAfterParse = format(parse(formatted, isoPattern, new Date()), isoPattern) 
console.log(formattedAfterParse) // '1919-07-01T04:32:18.999+04:00'
```

As you can see, results are different

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "coveralls": "^3.0.6",
     "eslint": "^5.16.0",
     "eslint-config-prettier": "^4.3.0",
+    "fast-check": "^1.17.0",
     "firebase": "^3.7.1",
     "flow-bin": "0.84.0",
     "fs-promise": "^1.0.0",
--- a/test/properties.test.js
+++ b/test/properties.test.js
@@ -1,0 +1,20 @@
+// @flow
+/* eslint-env mocha */
+
+import fc from 'fast-check'
+import format from '../src/format'
+import parse from '../src/parse'
+
+const isoPattern = `yyyy-MM-dd'T'HH:mm:ss.SSSXXX`
+
+it('format should return the same string after parse', () => {
+  fc.assert(
+    fc.property(fc.date(), date => {
+      const formatted = format(date, isoPattern)
+      return (
+        formatted ===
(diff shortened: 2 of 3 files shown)


dkozickis:
Hi @Djaler!

The idea is interesting, but I'm not entirely sure where do we go from here?
Re example test - you are passing different dates to functions.
If we pass same date - comparison is equal.

```javascript
it('format should return the same string after parse', () => {
  fc.assert(
    fc.property(fc.date({
      min: new Date(1990, 0, 1),
      max: new Date(2100, 0, 1)
    }), date => {
      const formatted = format(date, isoPattern)
      return (
        formatted ===
        format(parse(formatted, isoPattern, date), isoPattern)
      )
    })
  )
})
```

Djaler:
The main idea is that `parse` and `format` functions must be mutually reversible. format(X) === (format(parse(format(x))).

Also, as you can see in the results, `format` works incorrect even without `parse`:
```
const date = new Date("1919-07-01T00:31:40.999Z")

const isoPattern = `yyyy-MM-dd'T'HH:mm:ss.SSSXXX`

const formatted = format(date, isoPattern)
console.log(formatted) // '1919-07-01T04:31:59.999+04:00'
```

dkozickis:
As confusing as it may seem - your example is correct and expected result.

You're parsing from UTC time (`Z` at the end) and then formatting to local time. 
Difference in seconds comes due to the fact that timezones used to have differences in hours/minutes/seconds back in 1919. 

Timezone support is one of the milestones for `date-fns`.

Djaler:
Do you have any ideas how to modify this test?

PR #1125, opened 2019-03-20
CS locale

JozefBiros:
Hello, I tried to create locale for 'cs' language. It's far from perfect but it should work.

PR #1462, opened 2019-10-05
[CS] Change to use 24H time format

FredyC:
As per comment https://github.com/date-fns/date-fns/pull/1125#issuecomment-538044354 (_czech language ahead_) in Czech, we do not use AM/PM suffixes, but 24H format instead.

I've also verified it with Intl browser support.

image

cc @JozefBiros 

_Note: It's a tiny change, I am not going to add myself as contributor_


kossnocorp:
Hey, thank you for the change, it makes total sense. 

Please address the CI Error: 

> The snapshot on the disk doesn't match the generated snapshot: /home/travis/build/date-fns/date-fns/src/locale/cs/snapshot.md. Please run yarn locale-snapshots and commit the results.

You need to run `yarn locale-snapshots` and commit the result.

FredyC:
Shoot, I was hoping to get away by editing directly on GH 😊 

Anyway, there is no such script as `locale-snapshots` configure in `package.json`. No mention of how to update snapshot in `CONTRIBUTING.md` either. I am not sure how to proceed, please advice.

kossnocorp:
You're probably in a wrong directory 🤔 Check out https://github.com/date-fns/date-fns/blob/master/package.json#L22

FredyC:
Yea sorry about that, that was premature. I haven't realized my fork is so old, I switched to that patch branch and it's ok there, working on it :)

FredyC:
Btw, that command is not Windows-friendly, it's a good thing we have WSL for such cases, but not everyone knows about it.

image

Snapshot pushed, let's hope it passes :)

src/locale/cs/snapshot.md
```diff
@@ -293,12 +293,12 @@ If now is January 1st, 2000, 00:00.
...
-| 1999-12-31T00:00:00.000Z | včera v 12:00 dop.             |
-| 1999-12-27T00:00:00.000Z | poslední pondělí ve 12:00 dop. |
-| 1999-12-21T00:00:00.000Z | 21.12.99                       |
+| Date                     | Result                   |
+| ------------------------ | ------------------------ |
+| 2000-01-10T00:00:00.000Z | 10.1.00                  |
+| 2000-01-05T00:00:00.000Z | v středu o 0:00          |
+| 2000-01-02T00:00:00.000Z | zítra v 0:00             |
+| 2000-01-01T00:00:00.000Z | dnes v 0:00              |
+| 1999-12-31T00:00:00.000Z | včera v 0:00             |
```
kossnocorp: We need to use a padded string token, so it would be `00:00`, see `en-GB` as an example:

```

## `formatRelative`

If now is January 1st, 2000, 00:00.

| Date                     | Result               |
| ------------------------ | -------------------- |
| 2000-01-10T00:00:00.000Z | 10/01/2000           |
| 2000-01-05T00:00:00.000Z | Wednesday at 00:00   |
| 2000-01-02T00:00:00.000Z | tomorrow at 00:00    |
| 2000-01-01T00:00:00.000Z | today at 00:00       |
| 1999-12-31T00:00:00.000Z | yesterday at 00:00   |
| 1999-12-27T00:00:00.000Z | last Monday at 00:00 |
| 1999-12-21T00:00:00.000Z | 21/12/1999           |
```

FredyC:
Well, if you look at the image in the first comment, we don't use padded hours in CS. I am not sure how much can be Intl trusted if you have a better source, I can verify, but I don't think it should be like that really.

kossnocorp:
Good point, I'll check how it works for other locales.

kossnocorp:
I work with dates for so many years and didn't realize that some ppl use `9:00` instead of `09:00`.

kossnocorp:
So you're right, I'll be checking `Intl` for the reference from now on 👍

Image 2019-10-06 at 9 29 27 AM

FredyC:
Yea, those folks at Babylon tower must have had a really hard time to have even such small differences in formatting 😆

kossnocorp:
Yeah, right 🤣

Issue #1016, opened 2018-12-13
Extract `parseISO` from `toDate` and disallow passing strings into functions


kossnocorp:
See #1015 for the background.

`toDate` is the function that converts passed argument to a date object and used virtually in every function in the library. The thing is that it weights `1.27 KB` (after heavy size optimization), so the minimal build size couldn't go below the number.

For example, `addDays` size is `1.31 KB` (it was `983 B` in v1).

The size bottleneck is in ISO 8601 string parsing code, so even if you don't use strings as arguments, you're going to pay the price.

The solution to the problem is to extract this code into `parseISO` and make date-fns functions accept only dates and numbers as arguments. It will make the library slightly more difficult to use for those who use strings to store dates but more lightweight for others.

This is a breaking change and will cause a backlash, but I believe it's worth it. Many choose date-fns because it's light and without this change, we won't be able to achieve the best possible result.

leshakoss:
I agree. A lot of people choose date-fns because it is lightweight and we should try to be the best library possible for this type of users — while maintaining the same flexibility and cover all possible date use cases. Our main intent was always to work with `Date` objects, and while many users enjoyed that we treat ISO-8601 strings as first class citizens as well, I believe isolating the parsing code in its own separate function is the best solution because:
1. You only need to parse a date once, and invocation of parser in every function is overkill. I feel like every function should only be doing one thing.
2. Working with `Date` objects is our primary use case and many users just don't need to work with ISO-8601 string dates. 
3. [debatable] `parseISO` being a separate function forces users to describe their data flow more explicitly, which could reduce the number of bugs caused by the type mix up (why doesn't `date.getTime()` work here? Oh, because I haven't converted the string to date yet).

leshakoss:
And BTW I think that this change would mean that most of the functions wouldn't need `options` argument, so we won't need to generate `fp/{functionName}WithOptions` for them anymore, which would make FP module much easier to maintain

stof:
Shouldn't this issue be closed as https://github.com/date-fns/date-fns/pull/1023 was merged ?

marnusw:
I've been thinking about this change quite a bit because, to be honest, for me it has been a blocker upgrading many of my projects from `alpha.26` to `alpha.27`. My concern is not so much that arbitrary date strings (e.g. partial ISO dates) are no longer supported, but rather that the fully formed ISO date output from `JSON.stringify(new Date())` can no longer be used.

[...]

PR #1463, opened 2019-10-07
parseJSON - minimal implementation for parsing JSON dates from an API

marnusw:
The ISO date parsing previously done in `toDate` was moved into `parseISO` to keep the minimal build size low when string date manipulation is not required (#1016). The problem is that any application that retrieves a date from an API will parse some form of serialized date string before passing it to other functions. The following is typical example:

```js
const output = format(addDays(parseISO(dateFromAPI), 1), 'yyyy-MM-dd')
```

As soon as `parseISO` is used in this way the benefit of removing `toDate` has been negated. The `parseISO` and `parse` functions are still around 7000 and 8000 bytes respectively. The proposed `parseUTC` function is only ~520 bytes in size and sufficient to parse the standard JSON serialized date string formats of most server-side languages. 

Aside from complete ISO dates and the input types accepted by `toDate` the function returns an `Invalid Date`. The intention of this function will not be to parse user input, for this `parse` would still be needed, but it permits convenient parsing of date strings from an API with a known ISO format.

This is a proposed compromise/solution for the issue I raised [here](https://github.com/date-fns/date-fns/issues/1016#issuecomment-498315179), and something I have used in a couple of projects recently to good effect. 

I am still undecided about the best name for this function. My initial thought was `parseJSONDate`, but then I realized only UTC dates are supported and calling it `parseUTC` might be more generic and describe the function more specifically.

I would like to propose this function for inclusion in the core `date-fns` library, however, if it is deemed not a good fit I can also add it to `date-fns-tz`.

### DIFF
--- a/src/parseJSON/index.js
+++ b/src/parseJSON/index.js
@@ -30,7 +30,7 @@ import toDate from '../toDate/index.js'
  * @returns {Date} the parsed date in the local time zone
  * @throws {TypeError} 1 argument required
  */
-export default function index(argument) {
+export default function parseJSON(argument) {
   if (arguments.length < 1) {
     throw new TypeError(
       '1 argument required, but only ' + arguments.length + ' present'
--- a/src/parseJSON/index.js
+++ b/src/parseJSON/index.js
@@ -1,0 +1,59 @@
+import toDate from '../toDate/index.js'
+
+/**
+ * @name parseJSON
+ * @category Common Helpers
+ * @summary Parse a JSON date string
+ *
+ * @description
+ * Converts a complete ISO date string in UTC time, the typical format for transmitting
+ * a date in JSON, to a JavaScript `Date` instance.
+ *
+ * This is a minimal implementation for converting dates retrieved from a JSON API to
+ * a `Date` instance which can be used with other functions in the `date-fns` library.
+ * The following formats are supported:
+ *
+ *     - `2000-03-15T05:20:10.123Z`: The output of `.toISOString()` and `JSON.stringify(new Date())`
+ *     - `2000-03-15T05:20:10Z`: Without milliseconds
+ *     - `2000-03-15T05:20:10+00:00`: With a zero offset, the default JSON encoded format in some other languages
+ *     - `2000-03-15T05:20:10+0000`: With a zero offset without a colon
+ *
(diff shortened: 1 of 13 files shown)


marnusw:
Thanks for the feedback. I agree calling it `parseJSON` makes sense after all and it has been renamed.

src/parseJSON/index.js
```diff
@@ -0,0 +1,59 @@
...
+ *     - A `Date` instance will be cloned
+ *     - A `number` will be treated as a timestamp
+ *
+ * Any other input type or invalid date strings will return an `Invalid Date`.
+ *
+ * @param {String|Number|Date} argument A fully formed ISO1806 date string to convert
+ * @returns {Date} the parsed date in the local time zone
+ * @throws {TypeError} 1 argument required
+ */
+export default function index(argument) {
```
kossnocorp: Please replace `index` with `parseJSON`

kossnocorp:
Please add `/* eslint-env mocha */` to the top of the `test.js` to make ESLint happy.

marnusw:
Thank you for reviewing and accepting this @kossnocorp!

kossnocorp:
Thank **you**! I'll ship it with the next release.

src/parseJSON/index.js
```diff
@@ -0,0 +1,59 @@
...
+    if (parts) {
+      return new Date(
+        Date.UTC(
+          +parts[1],
+          parts[2] - 1,
+          +parts[3],
+          +parts[4],
+          +parts[5],
+          +parts[6],
+          +(parts[7] || 0)
```
TrySound: Yay, we have this in a few places in our app

kossnocorp:
I've shipped the PR with `2.6.0`: https://github.com/date-fns/date-fns/releases/tag/v2.6.0

alexdilley:
Is there an argument against simply using the parsing capabilities built into JS's `Date` type? i.e.

```js
new Date('2000-03-15T05:20:10.123Z')
```

Is it a matter of trust in the format supplied by the (commonly) API that would necessitate using a library-supplied function such as this?

marnusw:
@alexdilley yes it is simply about trust in the underlying implementation of different browsers. These days that might not be as relevant as it once was, but using this function leaves no room for doubt.

Issue #1468, opened 2019-10-09
Please provide a way to import some part of locale object


WarisR:
Now when I need to use locale object for formatRelative translation I do as below
```javascript
import { subDays, formatRelative } from 'date-fns';
import enUs from 'date-fns/locale/en-US'

return formatRelative(subDays(new Date(), 3), new Date(), { locale: enUs })
```

But enUs locale object has too many functionalities than what we need 
```javascript
var locale = {
  formatDistance: formatDistance,
  formatLong: formatLong,
  formatRelative: formatRelative,
  localize: localize,
  match: match,
  options: {
    weekStartsOn: 0 /* Sunday */,
    firstWeekContainsDate: 1
  }
};
```

Could you please provide a way to get only formatRelative from the locale object as below?
```javascript
import { subDays, formatRelative } from 'date-fns';
import { formatRelative as enFormatRelative } from 'date-fns/locale/en-US'

return formatRelative(
  subDays(new Date(), 3),
  new Date(),
  { locale: { formatRelative: enFormatRelative  }
)
```

aravind1078:
hey, I have submitted a pr on this #1469 can someone guide me further?

galabra:
I just opened a PR with an implementation to this feature
https://github.com/date-fns/date-fns/pull/2936

EnricoPf:
Starting work on this right away!

mohitsingh1930:
Is this still needed?
I am new to open source, can I start with this issue?

aparajitaks:
Hi! 👋 I'd like to work on this issue.

I’ll explore how partial locale object imports can be supported and propose an implementation that keeps the current API behavior intact.
If there are any specific guidelines, files, or edge cases I should look into before starting, please let me know.

Thanks!

RudraKo:
Hi! I'd like to take this issue.
I'll look into how partial locale imports can be supported and share a proposal soon.

PR #1469, opened 2019-10-09
export each key seperately along with complete local export  in en-us #1468

aravind1078:


kossnocorp:
Unfortunately, it's a breaking change that we can't accept right now. Please tell more about why do you need it, and we'll consider it for the next version if it makes sense.

WarisR:
@kossnocorp 
Now when I need to use locale object for formatRelative translation I do as below
```javascript
import { subDays, formatRelative } from 'date-fns';
import enUs from 'date-fns/locale/en-US'

return formatRelative(subDays(new Date(), 3), new Date(), { locale: enUs })
```

But enUs locale object has too many functionalities than what we need.
So I have to import this big bundle, but what I need is only formatRelative.
That make my js bundle bigger by import unused code.
```javascript
var locale = {
  formatDistance: formatDistance,
  formatLong: formatLong,
  formatRelative: formatRelative,
  localize: localize,
  match: match,
  options: {
    weekStartsOn: 0 /* Sunday */,
    firstWeekContainsDate: 1
  }
};
```

Could you please provide a way to get only formatRelative from the locale object as below?
```javascript
import { subDays, formatRelative } from 'date-fns';
import { formatRelative as enFormatRelative } from 'date-fns/locale/en-US'

return formatRelative(
  subDays(new Date(), 3),
  new Date(),
  { locale: { formatRelative: enFormatRelative  }
)
```

kossnocorp:
Thank you for your input, that makes sense, but we can't accept the PR as is. 

It will require changes across the board (update all locales and pending PRs, update docs, work with the community). 

If you want to champion the change, I'll be glad to offer guidance. Otherwise, I can't tell you when we'll get there as right now we have other priorities (time zones support).

kossnocorp:
Right now, you can simply copy-paste locales into your source code and apply the change there.

aravind1078:
@kossnocorp as @WarisR explained it was meant to solve #1468 . But I thought it was meant only for en-us.

PR #1470, opened 2019-10-09
Add en-AU

### LINKED ISSUE #1103: Add en-AU locale
please add more en-AU language
version: 2.0.0-alpha.27

JulienMalige:
Fix #1103

inspire by the en-AU of moment.js

https://github.com/moment/moment/blob/develop/locale/en-au.js


JulienMalige:
The snapshot.md is not ok. I have some issue to generated it. 
Is not just about to run `❯ ./scripts/build/build.sh`

src/locale/en-AU/_lib/formatLong/index.js
```diff
@@ -0,0 +1,41 @@
+import buildFormatLongFn from '../../../_lib/buildFormatLongFn/index.js'
+
+var dateFormats = {
+  full: 'EEEE, d MMMM yyyy',
+  long: 'd MMMM yyyy',
+  medium: 'd MMM yyyy',
+  short: 'dd/MM/yyyy'
+}
+
+var timeFormats = {
```
dkozickis: I believe 12-hour clock is used in AUS. Using `HH` makes it a 24-hour clock. 
I think it's also implemented that way in [moment](https://github.com/moment/moment/blob/develop/src/locale/en-au.js#L14)

kossnocorp: That's right, but as I'm working on shipping a new release I'm going to fix it myself in the `master`.

Issue #1461, opened 2019-10-05
isValid returns false for ISO8601 formatted date


jelovac:
Hi,

The isValid method returns false for the following ISO8601 formatted date string:

> 2019-10-25T13:59:54Z

According to the [Wikipedia article](https://en.wikipedia.org/wiki/ISO_8601) this is valid UTC date.

My usage example:

```
  dateToLocal(date?: string): string {

    if(date === undefined || !dateFns.isValid(date)) {
      return '';
    }

    let dateObject: Date = dateFns.parseISO(date);
    return dateObject.toLocaleString();
  }
```

Removing the validation part the parsing finished successfully.

dkozickis:
Hi @jelovac,

If you're on date-fns `2.x` then `isValid` does not accept `ISO8601` string as input.
Please see [docs](https://date-fns.org/v2.4.1/docs/isValid)

jelovac:
Hi @dkozickis,

Yes, I am using date-fns 2.x. I thought that it does because in the docs there was an example using the string input.

![Annotation 2019-10-08 072135](https://user-images.githubusercontent.com/3627139/66369433-aee62880-e99c-11e9-8f99-7202407c6f15.png)

I am guessing that toDate method doesn't support this.

dkozickis:
Yeah, that looks like a doc update is in order, thank you for reporting!
Would you care to submit a PR so that future generations can avoid same confusion?

jelovac:
Done.

Also I have found out that validation can be performed after parseISO, since parseISO will return instance of date object even for invalid date.

So the solution to my question is:

```
 let dateObject: Date = dateFns.parseISO(dateString);

  if (dateFns.isValid(dateObject) === false) {
    return '';
  }
```

politicsruinseverything:
I was going to open a new issue for this very thing, that isValid is returning false for ISO 8601. I won't do that but I would love to know why one of the most universal date formats in the JS world isn't considered valid by date-fns.

PR #1471, opened 2019-10-09
Updated isValid() example, refs #1461

jelovac:
Passing date as a string to isValid method will always result in false result on v2.x.


kossnocorp:
Thank you!

PR #1475, opened 2019-10-12
Export Interval and Locale types in flow typings

### LINKED ISSUE #1474: export Interval flowtype
The `Interval` flowtype isn't exported by date-fns, so I have to redefine it in my own code.

102:
fixes #1474

### DIFF
--- a/scripts/build/_lib/typings/flow.js
+++ b/scripts/build/_lib/typings/flow.js
@@ -35,7 +35,7 @@ const getFlowFPTypeAliases = (arity = 4) =>
 
 function getFlowTypeAlias(type) {
   const { title, properties } = type
-  return `type ${title} = ${getParams(properties)}`
+  return `export type ${title} = ${getParams(properties)}`
 }
 
 function generateFlowFnTyping(fn, aliasDeclarations) {
--- a/src/addBusinessDays/index.js.flow
+++ b/src/addBusinessDays/index.js.flow
@@ -1,12 +1,12 @@
 // @flow
 // This file is generated automatically by `scripts/build/typings.js`. Please, don't change it.
 
-type Interval = {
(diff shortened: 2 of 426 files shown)


dkozickis:
@leshakoss any reason we can’t reexport function definition from index.js.flow (same way it is done for TS)?

Maybe flow didn’t support reexport before, but it seems to support it now - https://github.com/facebook/flow/issues/1379#issuecomment-447357914

102:
> @leshakoss any reason we can’t reexport function definition from index.js.flow (same way it is done for TS)?
> 
> Maybe flow didn’t support reexport before, but it seems to support it now - [facebook/flow#1379 (comment)](https://github.com/facebook/flow/issues/1379#issuecomment-447357914)

I'm not sure that flow files for all the function folders required, I think `index.flow.js` with content like this example should work

```
declare module "date-fns" {
  declare export function format ...
}

declare module "date-fns/format" {
    declare export default $PropertyType<$Exports<"date-fns">, "format">;
}

...
```

kossnocorp:
>@leshakoss any reason we can’t reexport function definition from index.js.flow (same way it is done for TS)?
>
> Maybe flow didn’t support reexport before, but it seems to support it now - facebook/flow#1379 (comment)

I like the Flow approach better and hope we'll adapt it for TS in the future. First of all, the index TS file is an approachable mess (it's hard to comprehend, editors have a terrible time parsing and displaying it), secondly, with Flow definitions next to the source makes functions self-containing.

Imagine that you want to see a definition of a function. With Flow, you go to the function directory and inspect it.

Although it would be cool to keep common exports, such as interfaces, enums, etc in a single place and reexport them, and I hope we'll approach that in the future as well.

Issue #1386, opened 2019-08-27
Add date format constants


strarsis:
`format(...)` function requires a date format string as 2nd parameter.
For parsing dates in HTTP header (e.g. when using [request](https://github.com/request/request)) the date format would be [RFC 7231](https://tools.ietf.org/html/rfc7231#section-7.1.1.2).
For other purposes, the date format may be [ISO 8601](https://www.ietf.org/rfc/rfc3339.txt) instead.

It would be much better to use a constant from `date-fns` with the proper date format string instead of defining these date formats as constants oneself – or worse, just copying + pasting (without even the guarantee that the copied date format string is really complete and correct and defined under that name) the date format string into the method invocation. 
Also, as the [date-fns parse 2.x documentation states](https://date-fns.org/v2.0.1/docs/parse#description), its date format string conventions differ from other libraries. Having predefined constants, properly tuned and tested would be much better than trying to adopt an existing format string.

Example:
`date-fns.dateFormats.RFC_7231`
`date-fns.dateFormats.ISO_8601`

Some related issue in another repository: https://github.com/briannesbitt/Carbon/issues/897

Imballinst:
Hi! Is this feature still needed? I'd like to give it a go. If yes, any other formats that you'd like to see other than these 2?

strarsis:
https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_get-format

> ISO format refers to ISO 9075, not ISO 8601.
So ISO 9075 is also useful for MySQL.
Postgres documentation mentions the ISO 8601 format as well supported.

Imballinst:
Noted. Thanks for the further information! I can't guarantee this to be finished fast (because I'm quite busy this month) -- but my target is _hopefully_, at maximum, two weeks.

Imballinst:
@strarsis PTAL: https://github.com/date-fns/date-fns/pull/1477

Imballinst:
This feature is now live with the merging of https://github.com/date-fns/date-fns/pull/1536. The docs is up there with the latest release, as well! 🎉 

![image](https://user-images.githubusercontent.com/7077157/69151154-33949c80-0b2e-11ea-9fef-f20a873a5b25.png)

PR #1477, opened 2019-10-13
implement formatISO8601, formatISO9075, formatRFC3339, and formatRFC7231

Imballinst:
Fixes https://github.com/date-fns/date-fns/issues/1386.

## Checklist

- [x] `./scripts/build/build.sh`
- [x] `yarn test` passes
- [x] `yarn lint`

## Description

This PR adds 4 functions to format dates to certain standards, such as:

1. ISO 8601: `formatISO8601`
2. ISO 9075: `formatISO9075`
3. RFC 3339: `formatRFC3339`
4. RFC 7231: `formatRFC7231`


giautm:
Hi @Imballinst , can you add RFC3339 in this PR?
https://tools.ietf.org/html/rfc3339#section-5.6

kossnocorp:
I like the idea but I'm not sure about the implementation.

First of all, dependence on `format` makes the function unreasonable heavyweight. None of the formats require locales (RFC uses `MMM` but exclusively English, so it's easier to inline the months' array right into the function) so I'd prefer to have them self-containing.

Secondly, the RFC formats print GMT but in fact outputs the time in the local time zone.

Also, the naming is misleading.

I would like to see a set of focused functions instead. I'm thinking of `formatRFC7231`, `formatISO8601` and `formatISO9075` with an option that controls the completeness.

Imballinst:
> Hi @Imballinst , can you add RFC3339 in this PR?
> https://tools.ietf.org/html/rfc3339#section-5.6

@giautm  After I address the requested changes, sure!

re: @kossnocorp

> I like the idea but I'm not sure about the implementation.
> 
> First of all, dependence on `format` makes the function unreasonable heavyweight. None of the formats require locales (RFC uses `MMM` but exclusively English, so it's easier to inline the months' array right into the function) so I'd prefer to have them self-containing.

Hmm, that's actually true. So in that case, is it safe to say that we only need this part of the `format` function?

https://github.com/date-fns/date-fns/blob/ceb5615fea6b09f6eb6f9b31fc03b6596839ac53/src/format/index.js#L395-L466

> Secondly, the RFC formats print GMT but in fact outputs the time in the local time zone.

My bad, I missed this part. I should have subtracted it first using `Date.getTimezoneOffset()`

> Also, the naming is misleading.
> 
> I would like to see a set of focused functions instead. I'm thinking of `formatRFC7231`, `formatISO8601` and `formatISO9075` with an option that controls the completeness.

Agree, I think that is way better than `formatConstants`. Do we need to group them in one folder? e.g.

```
...
endOfYesterday
format
<name of folder>
  formatRFC7231
  formatISO8601
  formatISO9075
  ...
formatDistance
formatDistanceStrict
...  
```

because if the number of the formats increases, then it will add "noise" to the `src` folder. I'm not sure whether that is a good thing or not. What do you think?

kossnocorp:
> Hmm, that's actually true. So in that case, is it safe to say that we only need this part of the format function?

I would say we don't need anything from `format`, for example, ISO-9075 could look something like (it's a simplified example):

```js
function formatISO9075(date) {
  const year = date.getFullYear()
  const month = date.getMonth() + 1
  const day = date.getDate()
  const hours = date.getHours()
  const minutes = date.getMinutes()
  const seconds = date.getSeconds()
  return `${year}-${month}-${day} ${hours}:${minutes}:${seconds}`
}
```

> Agree, I think that is way better than formatConstants. Do we need to group them in one folder? e.g.

I don't bother with the noise in `src`, there're almost two hundred of functions, and a few more won't make any difference. Let's keep the flat structure.

Imballinst:
Gotcha. Thanks for the guide and clarification! Will definitely patch this PR.

kossnocorp:
Thank you for your work; it's an excellent addition to the library!

Imballinst:
@kossnocorp do you have any idea how to test the `formatRFC7231` function?

Since the test is running in the browser, the timezone may vary and therefore, the result may not always be the same. Should I force "normalize" the date in the test? e.g.

```js
const date = new Date(2019, 4, 4);
const tzOffset = date.getTimezoneOffset();
const testDate = addMinutes(date, tzOffset);

...
```

Imballinst:
Would you like to give another shot of feedback @kossnocorp? Thanks!

kossnocorp:
@Imballinst it's next on my virtual date-fns todo list 👌

leshakoss:
Our parse ISO 8601 is just called `parseISO`, maybe we should name `formatISO8601` as `formatISO` too, with `formatISO8601` as an alias? WDYT @Imballinst @kossnocorp

Imballinst:
I think it's good -- so that it [the ISO 8601 convention] is consistent across functions.

It _should_ be quite common in the perspective of the users as well, e.g. `.toISOString()` does not have "8601" in its function body.

kossnocorp:
@leshakoss @Imballinst you're right, that's a good idea. Too bad we don't have aliases mechanism right now (I'll look into it), so let's simply rename `formatISO8601` to `formatISO`.

Imballinst:
Done! Anyway, I'm curious. When a function file/folder gets renamed/removed, does `./scripts/build/build.sh` not remove the old/non-existent typings?

e.g. just now I renamed `formatISO8601` to `formatISO`, but the typings for `formatISO8601` was still there -- so I had to delete them manually (perhaps there is a better approach to this that I don't know (?))

Necmttn:
looking forward to this

leshakoss:
Hi @Imballinst ! Sorry for the huge delay. I decided to continue your work in a new PR: #1536
I've made a few API changes aimed to make the functions look internally consistent with other `formatX` functions.

Thanks a lot for your contribution! ⭐️

PR #1489, opened 2019-10-19
Add the language code to each locale spec

### LINKED ISSUE #1476: Request: Add the language code to the locale object spec for use with date-fns-tz
When the extended `format` function provided by `date-fns-tz` is used to ouptut the `z...zzzz` tokens the `locale` option can be provided and is used for the locale passed to `Intl.DateTimeFormat`. This is important because, for example, the `en-US` locale renders `Europe/London` ans `GMT+1` instead of `BST`, whereas the `en-GB` locale gives the preferred `BST` output.

The problem is that there is no way to determine which locale a specific locale spec is once it has been imported. This requires users to set the code on each imported locale explicitly somewhere in their application. From the `date-fns-tz` docs:

```js
// The time zone name is generated by the Intl API which works best 
// when a locale is also provided
import enGB from 'date-fns/locale/en-GB'

enGB.code = 'en-GB' // <-- This is not ideal

format(parisDate, 'yyyy-MM-dd HH:mm:ss zzz', { timeZone: 'Europe/Paris' })
// 2014-10-25 10:46:20 GMT+2
format(parisDate, 'yyyy-MM-dd HH:mm:ss zzz', { timeZone: 'Europe/Paris', locale: enGB }) 
// 2014-10-25 10:46:20 CEST
format(parisDate, 'yyyy-MM-dd HH:mm:ss zzzz', { timeZone: 'Europe/Paris', locale: enGB }) 
// 2014-10-25 10:46:20 Central European Summer Time
```

I would like to propose adding a new `code` field to all `date-fns@2` locales as part of the spec to resolve this.

Would it be a problem doing so? I will happily do the work if it will be accepted.

marnusw:
Closes #1476

### DIFF
--- a/docs/i18nContributionGuide.md
+++ b/docs/i18nContributionGuide.md
@@ -116,6 +116,7 @@ import match from './_lib/match/index.js'
  * @author Lesha Koss [@leshakoss]{@link https://github.com/leshakoss}
  */
 var locale = {
+  code: 'en',
   formatDistance: formatDistance,
   formatLong: formatLong,
   formatRelative: formatRelative,
--- a/src/locale/af/index.js
+++ b/src/locale/af/index.js
@@ -13,6 +13,7 @@ import match from './_lib/match/index.js'
  * @author Marnus Weststrate [@marnusw]{@link https://github.com/marnusw}
  */
 var locale = {
+  code: 'af',
   formatDistance: formatDistance,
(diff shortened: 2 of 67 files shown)


kossnocorp:
I've shipped the PR with `2.6.0`: https://github.com/date-fns/date-fns/releases/tag/v2.6.0

PR #1490, opened 2019-10-20
use const / let instead of var

anatoliy8493:


src/_lib/assign/index.js
```diff
@@ -7,7 +7,7 @@ export default function assign(target, dirtyObject) {
 
   dirtyObject = dirtyObject || {}
 
-  for (var property in dirtyObject) {
+  for (let property in dirtyObject) {
```
TrySound: I think this can be const

anatoliy8493: fixed

kossnocorp:
I appreciate the effort, but there's no way I'm going to go through every changed line, and I can't merge the PR without doing it.

Please tell me what tools you used to transform the code so I can do it myself.

Thanks!

anatoliy8493:
> I appreciate the effort, but there's no way I'm going to go through every changed line, and I can't merge the PR without doing it.
> 
> Please tell me what tools you used to transform the code so I can do it myself.
> 
> Thanks!

I did it manually

kossnocorp:
@anatoliy8493 but how?

PR #1497, opened 2019-10-23
Added Type Guard to isDate Function

lichtsteinerdev:
Used Type Guard instead of simple boolean for is Date function.

### DIFF
--- a/typings.d.ts
+++ b/typings.d.ts
@@ -490,7 +490,7 @@ declare module 'date-fns' {
   function isBefore(date: Date | number, dateToCompare: Date | number): boolean
   namespace isBefore {}
 
-  function isDate(value: any): boolean
+  function isDate(value: any): value is Date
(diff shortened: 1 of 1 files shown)


leshakoss:
@lichtsteinerdev this file is generated automatically by `scripts/build/typings.js`, your change will be erased the next time the script runs

kossnocorp:
As @leshakoss said the PR won't work. It would be cool to have it, but it's, unfortunately, it's difficult to implement.

mchambaud:
I know this is old.

I'm replacing home made date utility functions in our project with date-fns and have this exact issue. I was wondering if it was in the plans to have typings implementing the type guard? 

In the meantime and as as workaround for interested parties, I'm trying to extend date-fns typings locally in my project to have to type guard but I'm not having any luck, perhaps someone has done something similar in the past?

Issue #1496, opened 2019-10-23
Feature request: parseJSON, make trailing Z optional + allow more digits in milliseconds fiedl


dmytro-gokun:
"parseJSON" function released with 2.6.0 is very useful. However, i think it would be even more useful if it allowed to omit the trailing 'Z' (or '+00') and have more than 3 digits in the milliseconds field. The thing is that some server-side frameworks produces dates in that format by default. For example, I've got the case with .NET Core producing string like "2019-10-23T09:11:12.969108" when DateTimeKind is Unspecified. I know, this might be fixed by explicitly specifying DateTimeKind on all dates in my .NET app + setting JsonSerializer settings, but that's whole lot of work for a very small reason. And i think i'm not alone here with this issue.

If this feature request is accepted, I may work on a PR.

kossnocorp:
I like the idea. Please go on. Since you on it, could you also check other languages that you have access to and add a list of known supported languages to the function? Thanks!

dmytro-gokun:
Yep, will do.

kossnocorp:
Closed by #1499

PR #1499, opened 2019-10-23
Allow `parseJSON` to accept strings without a trailing 'Z' symbold and with up to 6 digits in the milliseconds field (#1496)

dmytro-gokun:
Here's one thing to think about though.. JS Date does not allow to have fractional milliseconds. So, the way i've implemented it is that extra digits are simply truncated. E.g.:

2019-12-31T23:59:59.999999Z becomes 2019-12-31T23:59:59.999Z

The question arises: do we need implemented a proper rounding here instead of truncation? E.g., with proper rounding:

2019-12-31T23:59:59.999999Z becomes 2020-01-01T00:00:00Z

In my opinion, this is not really needed and we should keep truncating it for the sake of simplicity. But I'd like to hear different opinions.

### DIFF
--- a/src/parseJSON/index.js
+++ b/src/parseJSON/index.js
@@ -17,6 +17,8 @@ import toDate from '../toDate/index.js'
  *     - `2000-03-15T05:20:10Z`: Without milliseconds
  *     - `2000-03-15T05:20:10+00:00`: With a zero offset, the default JSON encoded format in some other languages
  *     - `2000-03-15T05:20:10+0000`: With a zero offset without a colon
+ *     - `2000-03-15T05:20:10`: Without a trailing 'Z' symbol
+ *     - `2000-03-15T05:20:10.134566`: Up to 6 digits in milliseconds field. Only first 3 are taken into account since JS does now allow fractional milliseconds
  *
  * For convenience and ease of use these other input types are also supported
  * via [toDate]{@link https://date-fns.org/docs/toDate}:
@@ -39,7 +41,7 @@ export default function parseJSON(argument) {
 
   if (typeof argument === 'string') {
     var parts = argument.match(
-      /(\d{4})-(\d{2})-(\d{2})T(\d{2}):(\d{2}):(\d{2})(?:\.(\d{3}))?(?:Z|\+00:?00)/
+      /(\d{4})-(\d{2})-(\d{2})T(\d{2}):(\d{2}):(\d{2})(?:\.(\d{0,6}))?(?:Z|\+00:?00)?/
     )
     if (parts) {
       return new Date(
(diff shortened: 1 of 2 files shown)


src/parseJSON/index.js
```diff
@@ -17,6 +17,9 @@ import toDate from '../toDate/index.js'
  *     - `2000-03-15T05:20:10Z`: Without milliseconds
  *     - `2000-03-15T05:20:10+00:00`: With a zero offset, the default JSON encoded format in some other languages
  *     - `2000-03-15T05:20:10+0000`: With a zero offset without a colon
+ *     - `2000-03-15T05:20:10`: Without a trailing 'Z' symbol
+ *     - `2000-03-15T05:20:10.134566`: Up to 6 digits in milliseconds field. Only first 3 are taken into account
```
dkozickis: Not sure how JSDoc parser will treat this, but maybe leave `since JS does now allow fractional milliseconds` in line 21 instead of new line

dkozickis: It is not `up to` it's `between 3 and 6` :)

dkozickis:
Thank you @dmytro-gokun.

I think the big bad Oracle truncated milliseconds, so we can just follow the lead on this one.

src/parseJSON/index.js
dmytro-gokun: Fixed

dmytro-gokun: I've fixed the comment. But now i'm thinking... may be we should really allow "up to 6". Do you see any harm in that?

dmytro-gokun:
@dkozickis You are right. There is no real need to over-complicate this. If someone really cares about fractional milliseconds and rounding, they should probably do it server-side.

src/parseJSON/index.js
dkozickis: Not at all. Milliseconds come in different kinds and shapes :D

dmytro-gokun: Okay, i will fix it and add appropriate test cases

dmytro-gokun: Done

src/parseJSON/index.js
```diff
@@ -50,7 +52,7 @@ export default function parseJSON(argument) {
           +parts[4],
           +parts[5],
           +parts[6],
-          +(parts[7] || 0)
+          +(parts[7] || '0').substr(0, 3).padEnd(3, '0')
```
dkozickis: Using `padEnd` required shipping a polyfill for IE11 and even some older versions of Chrome and Firefox.
Is padding needed at all?

dmytro-gokun: Oops. I'll try to re-implement this without padEnd.

dmytro-gokun: Fixed in the last commit

dkozickis:
Thank you @dmytro-gokun !

PR #1505, opened 2019-10-27
areIntervalsOverlapping: improve doc

NaridaL:
make clear that adjacent intervals do not count as overlapping


dkozickis:
Thank you for your contribution @NaridaL !
That will make Docs more clear.

PR #1512, opened 2019-10-29
update day values for Kazakh language

talgautb:
I think that's right for day values.
and yes i am native speaker


dkozickis:
@date-fns/i18n-kk

src/locale/kk/_lib/localize/index.js
```diff
@@ -77,8 +77,8 @@ var formattingMonthValues = {
 
 var dayValues = {
   narrow: ['Ж', 'Д', 'С', 'С', 'Б', 'Ж', 'С'],
-  short: ['жс', 'дс', 'сс', 'ср', 'бс', 'жм', 'сб'],
-  abbreviated: ['жс', 'дс', 'сс', 'ср', 'бс', 'жм', 'сб'],
+  short: ['жк', 'дй', 'сй', 'ср', 'бй', 'жм', 'сн'],
```
jarjan: Can you cite where did you see such short usage of weekdays?

At [Astana Akimat official website](http://astana.gov.kz/) I could find `жс, дс, сс...` usage.

src/locale/kk/_lib/localize/index.js
```diff
@@ -77,8 +77,8 @@ var formattingMonthValues = {
...
+  abbreviated: ['жек', 'дүй', 'сей', 'сәр', 'бей', 'жұм', 'сен'],
```
jarjan: This is good catch!

src/locale/kk/_lib/localize/index.js
talgautb: a lot of site use like here https://github.com/moment/moment/blob/develop/src/locale/kk.js

kossnocorp: I see that [CLDR uses](https://www.unicode.org/cldr/charts/32/summary/kk.html#1544) `жс`. Unless there's better standard, we should stick to it.

src/locale/kk/_lib/localize/index.js
```diff
@@ -117,11 +117,11 @@ var formattingDayPeriodValues = {
...
-    noon: 'түс.',
-    morning: 'таң.',
-    afternoon: 'күн.',
-    evening: 'кеш.',
-    night: 'түн.'
+    noon: 'түс',
+    morning: 'таң',
+    afternoon: 'күн',
+    evening: 'кеш',
+    night: 'түн'
```
kossnocorp: @jarjan what do you think about this change?

kossnocorp: cc @drugoi

talgautb:
i have some trouble with windows 10, want to check in mac

src/locale/kk/_lib/localize/index.js
drugoi: I agree with @kossnocorp, and I also asked many native kazakh speakers and they confirm, that `жс, дс, сй, etc.` are commonly used format for short date notation.
Please, change it back, @talgautb.

src/locale/kk/_lib/localize/index.js
drugoi: Look's good 👍🏼

drugoi:
@talgautb what kind of problem do you have with snapshot generation on Windows 10?

talgautb:
@drugoi i think win problem need another issues,
ok i will check values

PR #1518, opened 2019-10-31
Don't export unused type aliases in all flow files

leshakoss:
Only export those that are used in arguments of a function.

👕👕


kossnocorp:
Let's delay this one until v3 because it introduces breaking change.

kossnocorp:
I reckon we don't need it given the TypeScript transition

PR #1529, opened 2019-11-02
Fix mistake in Slovenian translation

mzgajner:
Noticed a mistake I made when I initially translated the Slovenian locale (I guess I don't need to explain that Slavic languages can get complicated :blush:).


dkozickis:
Does this need a Snapshot rebuild @leshakoss ?

kossnocorp:
@dkozickis that's fine, I rebuild it before every release.

mzgajner:
I think the Slovenian translation has so many different combinations that they're not all exposed by the default snapshot - I rebuilt it in my commit but there were no changes.

PR #1530, opened 2019-11-02
isValid: use type predicate

NaridaL:
See https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates

Allows one to write `if (isValid(foo)) ...` instead of `if (foo && isValid(foo))` to check if foo is not undefined.

### DIFF
--- a/src/isValid/index.js
+++ b/src/isValid/index.js
@@ -38,8 +38,8 @@ import toDate from '../toDate/index.js'
  *   that try to coerce arguments to the expected type
  *   (which is also the case with other *date-fns* functions).
  *
- * @param {*} date - the date to check
- * @returns {Boolean} the date is valid
+ * @param {*} dirtyDate - the date to check
+ * @returns {dirtyDate is number | Date} the date is valid
  * @throws {TypeError} 1 argument required
  *
  * @example


src/isValid/index.js
```diff
@@ -38,8 +38,8 @@ import toDate from '../toDate/index.js'
  *   that try to coerce arguments to the expected type
  *   (which is also the case with other *date-fns* functions).
  *
- * @param {*} date - the date to check
- * @returns {Boolean} the date is valid
+ * @param {*} dirtyDate - the date to check
```
dkozickis: No need to change he name here - this is the param name that is exposed to used, `dirtyDate` is the internal name :)

src/isValid/index.js
```diff
@@ -38,8 +38,8 @@ import toDate from '../toDate/index.js'
...
+ * @returns {dirtyDate is number | Date} the date is valid
```
dkozickis: `returns` type is used for both Typescript and Flow types generation. Not to mention, that `dirtyDate is number | Date` is not a valid JSDoc/JS return type.
 
A change like this actually requires change to how our Typescript definitions are generated, I think.

* Add a custom tag to JSDoc (like `tsReturn`)
* Make our JSDoc pipeline understand it
* Use this specific tag in place of `returns` tag during TS typings generation.

Are you up for that?

kossnocorp:
As @dkozickis said, to make it work as you expected, the PR should be approached from another side. If you're willing to put your time into it, we'll offer you guidance. For now, I'm closing the PR.

PR #1535, opened 2019-11-07
differenceInMinutes: improve JSdoc

NaridaL:


NaridaL:
Might add them for the other differenceIn* functions later.

As an aside, it would be great if the doc was also in the .d.ts files. That way, you could see it directly in VS Code.

PR #1536, opened 2019-11-07
implement formatISO, formatISO9075, formatRFC3339, and formatRFC7231 (cont.)

leshakoss:
Continuing @Imballinst's work from #1477


leshakoss:
Motivation for naming options `representation` and `format`: apparently this is what those things are called in ISO-8601:

2.3 Representations and formats
2.3.1
date and time representation
expression indicating a time point, time interval or recurring time interval
2.3.2
date and time format representation
expression describing the format of a group of date and time representations
2.3.3
**basic format**
format of a date and time representation or date and time format representation comprising the minimum number of time elements necessary for the accuracy required
NOTE The basic format should be avoided in plain text.
2.3.4
**extended format**
extension of the basic format that includes additional separators
2.3.5
**complete representation**
representation that includes all the date and time components associated with the expression; limited, if applicable, for time elements of representations expressing a calendar year to four digits
2.3.6
decimal representation
expansion of a representation by addition of a decimal fraction to the lowest order component of the expression
2.3.7
representation with reduced accuracy
abbreviation of a representation by omission of lower order components
2.3.8
expanded representation
expansion of a representation to allow identification of dates in calendar years outside the range [0000] till [9999]

src/formatISO/test.js
```diff
@@ -0,0 +1,89 @@
...
+    assert(
+      formatISO(date, { representation: 'time', format: 'basic' }) === '190052'
+    )
+  })
+
+  describe('implicitly converts options', function() {
+    it('`format`', function() {
+      // eslint-disable-next-line no-new-wrappers
+      var format = new String('basic')
+      var date = new Date(2019, 9 /* Oct */, 4, 12, 30, 13, 456)
```
Imballinst: The `/* Oct */` comment is a nice touch. I didn't think of it in my previous PR.

src/formatISO/test.js
```diff
@@ -0,0 +1,89 @@
...
+      // $ExpectedMistake
```
Imballinst: What is being handled with this comment (?)

src/formatISO/index.js
```diff
@@ -0,0 +1,111 @@
...
+ * @name formatISO
+ * @category Common Helpers
+ * @summary Format the date according to the ISO 8601 standard (http://support.sas.com/documentation/cdl/en/lrdict/64316/HTML/default/viewer.htm#a003169814.htm).
+ *
+ * @description
+ * Return the formatted date string in ISO 8601 format. Options may be passed to control the parts and notations of the date.
+ *
+ * @param {Date|Number} date - the original date
+ * @param {Object} [options] - an object with options.
+ * @param {'extended'|'basic'} [options.format='extended'] - if 'basic', doesn't show delimeters between date and time values.
```
Imballinst:
```suggestion
 * @param {'extended'|'basic'} [options.format='extended'] - if 'basic', hide delimiters between date and time values.
```
sorry for nitpicking here, just a suggestion for changes:

1. "doesn't show" --> "hide"
2. "delimeter" --> "delimiter"

src/formatISO9075/index.js
```diff
@@ -0,0 +1,111 @@
...
+ * @name formatISO9075
+ * @category Common Helpers
+ * @summary Format the date according to the ISO 9075 standard (https://dev.mysql.com/doc/refman/5.7/en/date-and-time-functions.html#function_get-format).
+ *
+ * @description
+ * Return the formatted date string in ISO 9075 format. Options may be passed to control the parts and notations of the date.
+ *
+ * @param {Date|Number} date - the original date
+ * @param {Object} [options] - an object with options.
+ * @param {'extended'|'basic'} [options.format='extended'] - if 'basic', doesn't show delimeters between date and time values.
```
Imballinst:
```suggestion
 * @param {'extended'|'basic'} [options.format='extended'] - if 'basic', hide delimiters between date and time values.
```
same thing like the previous one in `formatISO`

src/formatRFC3339/index.js
```diff
@@ -0,0 +1,95 @@
...
+    throw new RangeError('Invalid time value')
+  }
+
+  const options = dirtyOptions || {}
+  const secondFractionDigits =
+    options.secondFractionDigits == null
+      ? 0
+      : toInteger(options.secondFractionDigits)
+
+  // Test if weekStartsOn is between 0 and 6 _and_ is not NaN
```
Imballinst:
```suggestion
  // Test if secondFractionDigits is between 0 and 3 inclusively
```
is this a typo (?)

src/formatRFC7231/index.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  const dayName = days[originalDate.getUTCDay()]
+  const dayOfMonth = addLeadingZeros(originalDate.getUTCDate(), 2)
+  const monthName = months[originalDate.getUTCMonth()]
+  const year = originalDate.getUTCFullYear()
+
+  const hour = addLeadingZeros(originalDate.getUTCHours(), 2)
+  const minute = addLeadingZeros(originalDate.getUTCMinutes(), 2)
+  const second = addLeadingZeros(originalDate.getUTCSeconds(), 2)
```
Imballinst: oh wow. I actually didn't know that these `UTC` functions exist. Good to know! 👍

src/formatISO/test.js
leshakoss: This is checked by flow at some point during CI tests. `format` the line below is not the type it's supposed to be (`String` instance instead of plain `string`) and this comment makes flow ignore it.

src/formatISO/test.js
```diff
@@ -0,0 +1,89 @@
...
+    )
+    assert(
+      formatISO(date, { representation: 'time', format: 'basic' }) === '190052'
+    )
+  })
+
+  describe('implicitly converts options', function() {
+    it('`format`', function() {
+      // eslint-disable-next-line no-new-wrappers
+      var format = new String('basic')
```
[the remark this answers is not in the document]
leshakoss: In this test we check that the function implicitly converts its parameters the same way JS standard library does. This should work with any object with `toString` method that returns `'basic'`, and `new String('basic')` is just the easiest way to create one.

src/formatISO/index.js
```diff
@@ -0,0 +1,111 @@
...
+
+  const hour = addLeadingZeros(originalDate.getHours(), 2)
+  const minute = addLeadingZeros(originalDate.getMinutes(), 2)
+  const second = addLeadingZeros(originalDate.getSeconds(), 2)
+
+  const options = dirtyOptions || {}
+  const format = options.format == null ? 'extended' : String(options.format)
+  const representation =
+    options.representation == null ? 'complete' : String(options.representation)
+
```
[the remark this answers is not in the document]
leshakoss: `x == null` is basically a shorthand for `x === null || x === undefined`

src/formatISO/index.js
leshakoss: Thanks a lot! 🙏

src/formatISO/test.js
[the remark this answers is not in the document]
Imballinst: Interesting. good to know!

src/formatISO/test.js
Imballinst: Cool -- thanks for clarifying!

src/formatISO/index.js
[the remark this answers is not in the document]
Imballinst: oooh! I didn't know about that. Alright -- seems good to me, then!

src/formatISO/index.js
```diff
@@ -0,0 +1,111 @@
...
+  const showTime = representation !== 'date'
+  const extended = format === 'extended'
+
+  // Result variables.
+  let result = ''
+  let dateDelimiter = ''
+  let timeDelimiter = ''
+
+  // Set delimiter, if extended. Otherwise, keep it empty.
+  if (extended) {
```
dkozickis: Personal opinion only - don't see the need to assign `extended` at line 83 just to check for it once

src/formatISO/index.js
```diff
@@ -0,0 +1,111 @@
...
+    dateDelimiter = '-'
+    timeDelimiter = ':'
+  }
+
+  // Write result if date options is true.
+  if (showDate) {
```
dkozickis: Same re. assigning `const` to check it only once - in this case it would be more clear (IMO) to inline the check here, and leave a comment above it, e.g.:

```javascript
// Representation is not limited only to time
// Write date portion
if (representation !== 'time') {
...
}
...
// Representation is not limited only to date
// Write time portion
if (representation !== 'date') {
...
}

```

src/formatRFC7231/index.js
```diff
@@ -0,0 +1,64 @@
+import toDate from '../toDate/index.js'
+import isValid from '../isValid/index.js'
+import addLeadingZeros from '../_lib/addLeadingZeros/index.js'
+
+const days = ['Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat']
```
dkozickis: These two arrays look like they have to be stored in some `superStandardStuff.js` :)

src/formatISO/index.js
dkozickis: Comment valid for `src/formatISO9075/index.js` as well

src/formatISO/index.js
kossnocorp: Either that or move it upper to simplify conditions on lines 69, 74-76.

src/formatISO9075/index.js
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  // Write result if date options is true.
+  if (showDate) {
+    // yyyyMMdd or yyyy-MM-dd.
+    result = `${year}${dateDelimiter}${month}${dateDelimiter}${day}`
+  }
+
+  if (showTime) {
+    const separator = result !== '' ? ' ' : ''
```
kossnocorp: A nitpick: `!== ''` could be removed.

leshakoss:
Cleaned it up a bit :)

src/formatRFC7231/index.js
leshakoss: I think next time :) So far it is used just here and in the locale

PR #1538, opened 2019-11-12
Remove next week preposition for Swedish locale

Neorth:
Removes formatRelative's nextWeek preposition to support wider use of the function. Can be compared to English 'on Monday at ...' vs just 'Monday at ...'.


dkozickis:
Thank you @alexandernanberg

PR #1540, opened 2019-11-14
Fix typo in Czech locale (formatDistance)

filipjuza:
Fixes typo in formatDistance for the Czech locale. A `y` character should be used instead of `i` in `xDays.few.regular`.


kossnocorp:
@date-fns/i18n-cs please take a look

kossnocorp:
Note to myself: the change is valid, I just need to push the branch with the updated snapshot.

PR #1553, opened 2019-12-01
Added an argurment in getDate() and getISODate()

geekMukh:
I am using this package in one of my project. I need to show the name of the day instead of the date number, for this I have to create a custom function to return data according to getDate().

### DIFF
--- a/src/getDay/index.js
+++ b/src/getDay/index.js
@@ -13,6 +13,7 @@ import toDate from '../toDate/index.js'
  * - [Changes that are common for the whole library](https://github.com/date-fns/date-fns/blob/master/docs/upgradeGuide.md#Common-Changes).
  *
  * @param {Date|Number} date - the given date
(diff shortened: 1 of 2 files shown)


kossnocorp:
I appreciate the effort, but I can't merge this PR. First of all, `day` in JavaScript means the index, not the name and it would appropriate to add a function `getDayName`. Secondly, it's not a good API to return different types depending on the input. Last but not least, even tho we use English as default, it's coming from locales and if you want to see this the change you should integrate the function with locales.

I understand that you need it and suggest you keep it in your codebase instead of adding it to date-fns.

kossnocorp:
Also, I appreciated the review game, that's clever!

geekMukh:
I didn't feel good after the reject, but the meme really uplifted my mood.
😁. I will try to contribute in future.

Thank You.

PR #1506, opened 2019-10-27
Locale and Interval types are not global anymore

hydrognomik:
Tyring to fix issue #1472.

PR #1555, opened 2019-12-02
Locale and Interval types are not global anymore

hydrognomik:
Reopen #1506.


kossnocorp:
I just realized that this is a breaking change, and I can't merge it right away. I can't increase the major version because of such a minor change.  We've planned to migrate to TypeScript in the next main release, so the problem will naturally go away. Given that I neither can merge it right nor it would be helpful with the next version, I have to close the PR.

I suggest you use another global type name, e.g., `LocaleCode`. I know that's not ideal but I can't help another way.

PR #1563, opened 2019-12-11
Move argument length requirement checks to a shared function to dedupe code

levibuzolic:
I noticed in a compiled bundle that there were a lot of duplicated strings which were used for argument length validation, so I was interested to see what effect deduplicating these would have on bundle size.

Bundle | Bytes
-------|--------
Before | `25474`
After   | `24528`

That's a reduction of **946 bytes** (3.71%) from a complete minified bundle.

### DIFF
--- a/src/_lib/getUTCDayOfYear/index.js
+++ b/src/_lib/getUTCDayOfYear/index.js
@@ -1,15 +1,12 @@
 import toDate from '../../toDate/index.js'
+import requiredArgs from '../requiredArgs/index.js'
 
 var MILLISECONDS_IN_DAY = 86400000
 
 // This function will be a part of public API when UTC function will be implemented.
 // See issue: https://github.com/date-fns/date-fns/issues/376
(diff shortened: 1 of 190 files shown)


kossnocorp:
I like it very much. Thank you for such great work! I considered making this change but was afraid that arguments leaking could affect the performance. I've run benchmarks, and it turned out that it's not a problem anymore.

I have a question, why did you choose to create a function before calling it inside of `requiredArgs`? I think that's redundant and unless there's a reason for this, I would ask you to change the API to simple `requiredArgs(2, arguments)`

levibuzolic:
@kossnocorp haha good point, originally I was planning to reuse instances but never came across for a need for it. 

Will simplify and update!

kossnocorp:
Thank you so much! I'll ship it with the next release once you update it.

levibuzolic:
@kossnocorp updated, thanks for the feedback. :+1:

PR #1567, opened 2019-12-12
Add the missing property code in the Locale type

doberkofler:


kossnocorp:
I appreciate the effort but you did update a generated file and the change will be overwritten. I'll open a PR with proper fix.

kossnocorp:
See: https://github.com/date-fns/date-fns/pull/1580

Issue #1559, opened 2019-12-09
formatISO omitting final `Z` character in dates


aryzing:
An ISO date should have either a time offset or `Z` appended to the end. Omitting this `Z` character causes the date as being relative to the host machine's timezone. When performing operations across time zones, this is problematic.

jpodpro:
just noticing this too. isn't the whole point of iso8601 that it must have a time-zone specified also?

Imballinst:
Hmm, back then we only took the date, time, and datetime as the scope (see http://support.sas.com/documentation/cdl/en/lrdict/64316/HTML/default/viewer.htm#a003169814.htm). We can add timezone if it is needed.

I think the reason we thought that it was okay to omit timezone because ISO8601 is mostly in UTC (?) cc: @leshakoss

kossnocorp:
I believe the full output should contain the timezone. I didn't think about that when I was looking at the PR. Not sure how it fits the API though.

Imballinst:
I can work on this issue, this weekend.

cyberwombat:
Perhaps along the same lines - doesn't seem possible to use format with UTC either. See this SO post https://stackoverflow.com/questions/58561169/date-fns-how-do-i-format-to-utc

PR #1576, opened 2019-12-22
Add timezone to formatISO

Imballinst:
Fixes https://github.com/date-fns/date-fns/issues/1559.

### DIFF
--- a/src/formatISO/index.js
+++ b/src/formatISO/index.js
@@ -71,6 +71,7 @@ export default function formatISO(dirtyDate, dirtyOptions) {
   }
 
   let result = ''
+  let tzOffset = ''
 
   const dateDelimiter = format === 'extended' ? '-' : ''
   const timeDelimiter = format === 'extended' ? ':' : ''
@@ -87,6 +88,22 @@ export default function formatISO(dirtyDate, dirtyOptions) {
 
   // Representation is either 'time' or 'complete'
   if (representation !== 'date') {
+    // Add the timezone.
+    const offset = originalDate.getTimezoneOffset()
+
+    if (offset !== 0 || format === 'extended') {
+      const absoluteOffset = Math.abs(offset)
+      const hourOffset = addLeadingZeros(absoluteOffset / 60, 2)
+      const minuteOffset = addLeadingZeros(absoluteOffset % 60, 2)
+      // If less than 0, the sign is +, because it is ahead of time.
+      const sign = offset < 0 ? '+' : '-'
+
+      tzOffset = `${sign}${hourOffset}:${minuteOffset}`
+    } else {
+      // The notation "Z" only applies for basic format AND if the timezone offset is 0.
+      tzOffset = 'Z'
+    }
+
     const hour = addLeadingZeros(originalDate.getHours(), 2)
     const minute = addLeadingZeros(originalDate.getMinutes(), 2)
     const second = addLeadingZeros(originalDate.getSeconds(), 2)
@@ -95,7 +112,7 @@ export default function formatISO(dirtyDate, dirtyOptions) {
     const separator = result === '' ? '' : 'T'
 
     // HHmmss or HH:mm:ss.
-    result = `${result}${separator}${hour}${timeDelimiter}${minute}${timeDelimiter}${second}`
+    result = `${result}${separator}${hour}${timeDelimiter}${minute}${timeDelimiter}${second}${tzOffset}`
   }
 
   return result
--- a/src/formatISO/index.js
+++ b/src/formatISO/index.js
@@ -12,7 +12,7 @@ import addLeadingZeros from '../_lib/addLeadingZeros/index.js'
  *
  * @param {Date|Number} date - the original date
  * @param {Object} [options] - an object with options.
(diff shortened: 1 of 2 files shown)


src/formatISO/test.js
```diff
@@ -3,25 +3,54 @@
...
+    const sign = offset < 0 ? '+' : '-'
+
+    tzOffset = `${sign}${hourOffset}:${minuteOffset}`
+  } else {
+    // The notation "Z" only applies for basic format AND if the timezone offset is 0.
+    tzOffset = 'Z'
+  }
+
+  return tzOffset
+}
```
Imballinst: This is similar with https://github.com/date-fns/date-fns/blob/fa96c90128e87024f7e8f344e4f764924ebcc542/src/formatRFC3339/test.js#L8-L26, but different logic in the first `if` branch.

In the RFC format, there is no `extended` format, while ISO formats have it.

src/formatISO/index.js
```diff
@@ -95,7 +112,7 @@ export default function formatISO(dirtyDate, dirtyOptions) {
     const separator = result === '' ? '' : 'T'
 
     // HHmmss or HH:mm:ss.
-    result = `${result}${separator}${hour}${timeDelimiter}${minute}${timeDelimiter}${second}`
+    result = `${result}${separator}${hour}${timeDelimiter}${minute}${timeDelimiter}${second}${tzOffset}`
```
Imballinst: now that I think about it, is it better to do `[hour, minute, second].join(timeDelimiter)` instead for readability purposes? 😅

src/formatISO/test.js
```diff
@@ -3,25 +3,54 @@
...
+    const sign = tzOffset < 0 ? '+' : '-'
+
+    offset = `${sign}${hourOffset}:${minuteOffset}`
+  } else {
+    // The notation "Z" only applies for basic format AND if the timezone offset is 0.
+    offset = 'Z'
+  }
+
+  return offset
+}
```
Imballinst: This is similar with https://github.com/date-fns/date-fns/blob/fa96c90128e87024f7e8f344e4f764924ebcc542/src/formatRFC3339/test.js#L8-L26, but different logic in the first `if` branch.

In the RFC format, there is no `extended` format, while ISO formats have it.

Imballinst:
I'll fix the CI test slightly later™

src/formatISO/index.js
kossnocorp: @Imballinst yes, that makes sense 👍

src/formatISO/test.js
```diff
@@ -34,12 +62,16 @@ describe('formatISO', () => {
 
   it('formats only time', () => {
     const date = new Date(2019, 2 /* Mar */, 3, 19, 0, 52, 123)
+    const tzOffsetBasic = generateOffset(date)
+    const tzOffsetExtended = generateOffset(date)
```
kossnocorp: I suppose these two are equal, so why don't we use a single variable?

Imballinst: ...right. I'll update it.

src/formatISO/index.js
```diff
@@ -12,7 +12,7 @@ import addLeadingZeros from '../_lib/addLeadingZeros/index.js'
  *
  * @param {Date|Number} date - the original date
  * @param {Object} [options] - an object with options.
- * @param {'extended'|'basic'} [options.format='extended'] - if 'basic', hide delimiters between date and time values.
+ * @param {'extended'|'basic'} [options.format='extended'] - if 'basic', hide delimiters between date and time values and hide time zone.
```
Imballinst: @kossnocorp I have updated the JSDoc comments. Please take a look!

kossnocorp:
@Imballinst please hold on, the review is coming. I don't like the fact that timezone presence is tied to delimiters, because they aren't connected. But I can't find a solution yet, and without a solution, the review is pointless.

I hope @leshakoss can help me here because it was he who developed the current argument schema.

Imballinst:
> @Imballinst please hold on, the review is coming. I don't like the fact that timezone presence is tied to delimiters, because they aren't connected. But I can't find a solution yet, and without a solution, the review is pointless.
> 
> I hope @leshakoss can help me here because it was he who developed the current argument schema.

Sorry for that! I thought you missed this. Thanks for clarifying!

Could you tell more about "timezone tied to delimiters"? Isn't the timezone only tied to the `time` and `complete` representation (regardless of `basic` or `extended` format)?

Alternatively, we could add something like, `includeTimezone` field in the `options` object. But yeah, I agree, this needs to be discussed. Perhaps there is a best way to do that.

kossnocorp:
> Could you tell more about "timezone tied to delimiters"? Isn't the timezone only tied to the time and complete representation (regardless of basic or extended format)?

You're right I completely missed it. I'm not even sure what I was talking about lol. Sorry about that.

I took a second look at the code and the current behavior is what we need.

Sorry again.

src/formatISO/index.js
kossnocorp: Oh god damn, that's the line that confused me. The docs are just misleading. I'll fix that in the master.

PR #1581, opened 2020-01-03
Feature/implement generic add

### LINKED ISSUE #1427: Feat: multi add function to add multiple parts of a date
Similar to #1388 (which was implemented by @SneakyFish5 in #1398 ), it would be nice to have a function, that can add multiple values at once.

So instead of doing this:
```
addSeconds(addMinutes(addHours(myDate, 12), 12), 12);
```

Something like this would be nice:
```
add(myDate, { hours: 12, minutes: 12, seconds: 12 });
```

Anshuman71:
Implement the add(Object) function from Moment and fix #1427

Add 2 hours, 15 minutes and 30 seconds to 10 July 2014 12:45:00;

var result = add(new Date(2014, 6, 10, 12, 45, 0), { minutes: 15, hours: 2, seconds: 30 });
result: Thu Jul 10 2014 15:00:30

Supports adding

{
    hours: number,
    years: number,
    days: number,
    quarters: number,
    weeks: number,
    months: number,
    minutes: number,
    seconds: number,
    milliseconds: number
} 
combined in a single call

### DIFF
--- a/src/add/index.js
+++ b/src/add/index.js
@@ -54,10 +54,8 @@ export default function add(dirtyDate, givenAmount) {
     )
   }
   const givenDate = toDate(dirtyDate)
-  const dirtyAmount = {
-    ...DEFAULT,
-    ...givenAmount
-  }
+  const dirtyAmount = Object.assign({}, DEFAULT)
+  Object.assign(dirtyAmount, givenAmount)
   const {
     hours,
     years,
--- a/src/add/index.js
+++ b/src/add/index.js
@@ -7,6 +7,18 @@ import addMonths from '../addMonths/index.js'
 import addQuarters from '../addQuarters/index.js'
 import addYears from '../addYears/index.js'
 import toDate from '../toDate/index'
+import toInteger from '../_lib/toInteger/index.js'
+
+const DEFAULT = {
+  hours: 0,
+  years: 0,
+  days: 0,
+  quarters: 0,
+  weeks: 0,
+  months: 0,
+  minutes: 0,
+  seconds: 0
+}
 
 /**
  * @name add
@@ -30,16 +42,22 @@ import toDate from '../toDate/index'
  * var result = add(new Date(2014, 6, 10, 12, 45, 0), {minutes: 15, hours: 2, seconds: 30})
  * //=> Thu Jul 10 2014 15:00:30
  */
-export default function add(dirtyDate, dirtyAmount) {
+export default function add(dirtyDate, givenAmount) {
   if (arguments.length < 2) {
     throw new TypeError(
       '2 arguments required, but only ' + arguments.length + ' present'
     )
   }
-  if (typeof dirtyAmount !== 'object' || !Object.keys(dirtyAmount).length) {
-    return new Date(NaN)
+  if (typeof givenAmount !== 'object' || !Object.keys(givenAmount).length) {
+    throw new TypeError(
+      '2nd argument expected Object, found ' + typeof givenAmount
+    )
   }
   const givenDate = toDate(dirtyDate)
+  const dirtyAmount = {
+    ...DEFAULT,
+    ...givenAmount
+  }
   const {
     hours,
     years,
@@ -56,17 +74,20 @@ export default function add(dirtyDate, dirtyAmount) {
         addWeeks(
           addDays(
             addHours(
-              addMinutes(addSeconds(givenDate, seconds || 0), minutes || 0),
-              hours || 0
+              addMinutes(
+                addSeconds(givenDate, toInteger(seconds)),
+                toInteger(minutes)
+              ),
+              toInteger(hours)
             ),
-            days || 0
+            toInteger(days)
           ),
-          weeks || 0
+          toInteger(weeks)
         ),
-        months || 0
+        toInteger(months)
       ),
-      quarters || 0
+      toInteger(quarters)
     ),
-    years || 0
+    toInteger(years)
   )
 }
--- a/src/add/index.js
+++ b/src/add/index.js
@@ -1,46 +1,51 @@
-import addSeconds from '../addSeconds/index.js'
-import addMinutes from '../addMinutes/index.js'
-import addHours from '../addHours/index.js'
-import addDays from '../addDays/index.js'
-import addWeeks from '../addWeeks/index.js'
-import addMonths from '../addMonths/index.js'
-import addQuarters from '../addQuarters/index.js'
-import addYears from '../addYears/index.js'
 import toDate from '../toDate/index.js'
+import getDaysInMonth from '../getDaysInMonth/index.js'
 import toInteger from '../_lib/toInteger/index.js'
-
-const DEFAULT = {
-  hours: 0,
-  years: 0,
-  days: 0,
-  quarters: 0,
-  weeks: 0,
-  months: 0,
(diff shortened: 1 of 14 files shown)


kossnocorp:
@Anshuman71 thank you a lot!

src/add/index.js
```diff
@@ -0,0 +1,92 @@
...
+ * // Add followind amount to 'Mon Sep 01 2014 10:20:00 GMT+0530 (India Standard Time)':
+ * var result = add(new Date(2014, 8 , 1, 10, 19, 50, 10000), {
+ *     milliseconds: 30000,
+ *     seconds: 30,
+ *     minutes: 9,
+ *     hours: 5,
+ *     days: 7,
+ *     weeks: 1,
+ *     months: 9,
+ *     quarters: 5,
```
kossnocorp: Let's remove the support of quarters. First of all, it's ambiguous and overlaps with months and also it's missing in Temporal's Duration and ISO 8601.

kossnocorp: Btw, I suppose the result of the example will change, so please update it as well.

Anshuman71:
> @Anshuman71 thank you a lot!

Can you please help me with this failing test?

kossnocorp:
@Anshuman71 sure, I'm working on a review atm, and tests will pass once you address my comments. Please stay put.

src/add/index.js
```diff
@@ -0,0 +1,92 @@
...
+export default function add(dirtyDate, givenAmount) {
+  if (arguments.length < 2) {
+    throw new TypeError(
+      '2 arguments required, but only ' + arguments.length + ' present'
+    )
+  }
+
+  if (!givenAmount) {
+    return new Date(NaN)
+  }
```
kossnocorp: JavaScript throws `RangeError` whenever option values are wrong. Please do a simple object check instead:

```js
if (!duration || typeof duration !== 'object') throw new RangeError('duration must be an object')
```

src/add/index.js
```diff
@@ -0,0 +1,92 @@
...
+  let milliseconds = toInteger(givenAmount.milliseconds) || 0
+  let seconds = toInteger(givenAmount.seconds) || 0
+  let minutes = toInteger(givenAmount.minutes) || 0
+  let hours = toInteger(givenAmount.hours) || 0
+  let days = toInteger(givenAmount.days) || 0
+  let months = toInteger(givenAmount.months) || 0
+  const quarters = toInteger(givenAmount.quarters) || 0
+  const weeks = toInteger(givenAmount.weeks) || 0
+  const years = toInteger(givenAmount.years) || 0
```
kossnocorp: Please don't default values to 0 after `toInteger` because it will suppress any `NaN` in here, which would be behavior that is hard to debug. Instead, check if the key is present and default it to 0 only if it's not:

```js
let milliseconds = 'milliseconds' in duration ? toInteger(duration.milliseconds) : 0
```

This way `NaN` will stay `NaN` even if it's passed as the value, so the resulting date would be `Invalid Date`.

src/add/index.js
```diff
@@ -0,0 +1,92 @@
...
+ * @category Combined time Helpers
+ * @summary Add the specified milliseconds, seconds, minutes, hours, days, months, quarters and years to the given date.
+ *
+ * @description
+ * Add the specified milliseconds ,seconds, minutes, hours, days, months, quarters and years to the given date.
+ *
+ * - [Changes that are common for the whole library](https://github.com/date-fns/date-fns/blob/master/docs/upgradeGuide.md#Common-Changes).
+ *
+ * @param {Date|Number} date - the date to be changed
+ * @param {Object|Number} amount - the object with milliseconds, seconds, minutes, hours, days, months, quarters and years to be added
```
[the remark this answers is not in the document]
Anshuman71: Sounds good to me

src/add/index.js
```diff
@@ -0,0 +1,92 @@
...
+import getDaysInMonth from '../getDaysInMonth/index.js'
+import toInteger from '../_lib/toInteger/index.js'
+
+/**
+ * @name add
+ * @category Combined time Helpers
+ * @summary Add the specified milliseconds, seconds, minutes, hours, days, months, quarters and years to the given date.
+ *
+ * @description
+ * Add the specified milliseconds ,seconds, minutes, hours, days, months, quarters and years to the given date.
```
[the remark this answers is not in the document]
Anshuman71: done

src/add/index.js
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * - [Changes that are common for the whole library](https://github.com/date-fns/date-fns/blob/master/docs/upgradeGuide.md#Common-Changes).
+ *
+ * @param {Date|Number} date - the date to be changed
+ * @param {Object|Number} amount - the object with milliseconds, seconds, minutes, hours, days, months, quarters and years to be added
+ *
+ * | Key                 |            Significance              |
+ * |---------------------|------------------------------------- |
+ * | milliseconds        |   Amount of milliseconds to be added |
```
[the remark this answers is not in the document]
Anshuman71: Sure

src/add/index.js
kossnocorp: @Anshuman71 this is still actual, please address it.

src/add/index.js
kossnocorp: @Anshuman71 this also still actual.

src/add/index.js
Anshuman71: @kossnocorp Sorry I think I missed those while going through review

src/add/index.js
Anshuman71: @kossnocorp Sorry I think I missed those while going through the review

src/add/index.js
Anshuman71: > Btw, I suppose the result of the example will change, so please update it as well.

I think it shouldn't because example doesn't use quarters

src/add/index.js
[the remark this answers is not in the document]
Anshuman71: > Let's remove the `Number` support. There's `addMilliseconds` for this.
Is it related to this?

Cannot call add.bind with new Date(...) bound to duration because number [1] is
incompatible with object type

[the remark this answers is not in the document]
Anshuman71: I suspect Flow types `Object` for duration doesn't match Number in tests

kossnocorp:
@Anshuman71 I'm going to push some changes in the PR. I reevaluated how other functions handle incorrect arguments, and now I think `RangeError` is a wrong approach. I also want to adjust tests a bit. Please don't push anything.

Anshuman71:
> @Anshuman71 I'm going to push some changes in the PR. I reevaluated how other functions handle incorrect arguments, and now I think `RangeError` is a wrong approach. I also want to adjust tests a bit. Please don't push anything.

Sure, go ahead

kossnocorp:
I've pushed some changes. Most notable changes are:

- I've rolled back the Invalid Error on invalid duration. You were right returning it.
- I've removed weeks and milliseconds to make the duration object correspond to ISO 8601's duration. This will help us down the road when we are going to add support to it the library.
- I reused `addMonths` to avoid copy-paste of the code with quirks.

Anshuman71:
> Great work!

@kossnocorp thanks for your help in shipping the feature

kossnocorp:
You're welcome!

Issue #869, opened 2018-09-13
Granularity


Christilut:
I couldn't find it in the docs, but is there a granularity option like Moment has?

For example:
```
date.isAfter(Moment(), 'day') 
```

I want my `date` to be after today (regardless of time).

jemerald:
Could compose with `startOfDay`: `isAfter(startOfDay(date1), startOfDay(day2))`

dkozickis:
Closing as workaround was provided. 
@Christilut - please re-open if you have further questions.

enzoferey:
@dkozickis Hi ! Do you think this is a feature we want inside the library ? I have myself implemented it in my projects as it's straightforward to due with all the existing `startOf*UNIT*` functions. I would be glad to send a PR, and if you provide me some indications about how you invision it in the codebase it would be perfect 👍

kossnocorp:
@enzoferey could you please tell first how do you use this function? My concern is that such function adds extra bytes to your build size (this function would bundle 6 other functions) and that people would start using it without an actual need for it.

enzoferey:
Thanks for pointing it out.

I started using this syntax because of a migration from `moment`, which uses this last argument granularity API. I do think having an API like `isBeforeByDay`, `isBeforeByHour`, etc. would make more sense inside the library. It's more explicit, you don't rely on magic strings constants and you would add to the bundle what is required (`isBefore` + `isStartOf*unit*`).

That said, it's only a proposal, I haven't thought in detail about it and I don't know the vision you guys share on how the library should evolve. I do think that people expect this kind of APIs as first class APIs and not as some clever reuse of functions around (proof of that is this issue, myself googling it, and potentially all the people that arrive here but didn't comment).

kossnocorp:
I like the idea of `isBeforeByDay` and similar (but `isBeforeDay`, `isAfterWeek` names will be more fitting), so you or anyone else feel adventurous, feel free to open PR which I'll gladly accept.

enzoferey:
I will open a PR the latest next weekend covering all `isBefore` and `isAfter` variations 👍

kossnocorp:
Nice, thank you a lot!

PR #1595, opened 2020-01-11
Implement date comparison granularity

enzoferey:
Following up https://github.com/date-fns/date-fns/issues/869.

I have implemented `isBeforeDay` and I would like to know if this format suits you before doing all the others as it's quite mechanical from here. I have taken `isBefore` as a reference and tweaked the docs and test cases accordingly.

Also, I can do one PR for all the `isBefore` variations and another one for the `isAfter` variations if that makes it easier to review (or even one per function if you prefer). Let me know 👍 

Pinging @kossnocorp as he was the one I talked with 😁

### DIFF
--- a/src/isBeforeDay/index.js
+++ b/src/isBeforeDay/index.js
@@ -21,7 +21,7 @@ import requiredArgs from '../_lib/requiredArgs/index.js'
  * //=> true
  *
  * // Is 11 July 1989 09:00 before 11 July 1989 10:00 by day?
- * var result = isBeforeDay(new Date(1989, 6, 10, 9), new Date(1989, 6, 11, 10))
+ * var result = isBeforeDay(new Date(1989, 6, 11, 9), new Date(1989, 6, 11, 10))
  * //=> false
  */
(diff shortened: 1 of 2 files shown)


enzoferey:
Anybody ?

dcousens:
When I do these operations myself,  I typically write them out for readability:

For example
```
isBefore(startOfDay(a), startOfDay(b))
```

I don't think `isBeforeDay` is easier to read than ^.

dcousens:
The only scenario where `isBeforeDay` is different to `isBefore` is when the day is the same,  but,  if I'm worried about that,  I'd use:

``` js
differenceInCalendarDays(a, b) < 0
```

enzoferey:
> So maybe `isBeforeCalendarDay` is better? 🤷‍♂

The idea is to have `isBeforeHour`, `isBeforeWeek`, ..., as well. That's why @kossnocorp suggested that name I guess. `isBeforeCalendarMinute` wouldn't be very good. That said, I'm pretty open to whatever naming.

micalevisk (approved):

PR #1599, opened 2020-01-16
Fix for timezones with minute Offsets >0

dcRUSTy:
For example Indian Standard Timezone IST (https://en.wikipedia.org/wiki/Indian_Standard_Time) +05:30 or 330 minutes the output becomes +5.5:30 but as per ISO 8601 standard it should be +05:30. hourOffset should be "05" and minute offset should be "30" but currently hourOffset is set to "5.5"

### DIFF
--- a/src/formatISO/index.js
+++ b/src/formatISO/index.js
@@ -93,7 +93,7 @@ export default function formatISO(dirtyDate, dirtyOptions) {
 
     if (offset !== 0) {
       const absoluteOffset = Math.abs(offset)
-      const hourOffset = addLeadingZeros(absoluteOffset / 60, 2)
+      const hourOffset = addLeadingZeros(Math.floor(absoluteOffset / 60), 2)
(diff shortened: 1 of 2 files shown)


dcRUSTy:
Screenshot 2020-01-16 at 8 52 48 PM
it should be "05" and not "5.5"

dcRUSTy:
Still not merged?

mtkopone:
I found this PR only after also implementing the fix. If you want a test that actually tests the output to go with this, feel free to use: https://github.com/mtkopone/date-fns/commit/c0f4396c4a9e4189b3ccf3a950e2fe2384a15dc9

kossnocorp:
@mtkopone that's perfect, thank you a lot for this!

kossnocorp:
I was overwhelmed lately, thank you for your patience! The fix was released with `2.10.0`.

Issue #1601, opened 2020-01-16
Fate of v1 parse implementation in v2 docs


icleolion:
parseISO v2 documentation currently states
> The previous parse implementation was renamed to parseISO.

parse v2 documentation currently states
> Old parse was renamed to toDate. Now parse is a new function which parses a string using a provided format.

I've just started upgrading to v2, so going just by the documentation I would say that parseISO is correct, and parse documentation should be updated to say
>  Old parse was renamed to parseISO. Now parse is a new function which parses a string using a provided format.

PR #1605, opened 2020-01-17
update docs to expand on changes to v1 parse in v2

icleolion:
Expands on existing comment in parse, to help direct users currently using v1 parse to the right function in v2.

Removes comment from parseISO as there isn't an equivalent comment in toDate. Doesn't make sense to have it in one but not the other.  Alternatively I can update PR with a comment in both parseISO and toDate if that is preferred.

Fixes https://github.com/date-fns/date-fns/issues/1601

Also fixed formatting and typo in parseJSON documentation.

### DIFF
--- a/src/parse/index.js
+++ b/src/parse/index.js
@@ -308,17 +308,10 @@ var unescapedLatinCharacterRegExp = /[a-zA-Z]/
  *
  * - [Changes that are common for the whole library](https://github.com/date-fns/date-fns/blob/master/docs/upgradeGuide.md#Common-Changes).
  *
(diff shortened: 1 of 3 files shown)


dcousens:
NACK in favour of https://github.com/date-fns/date-fns/pull/1624 (it retains the code examples)

rikkalo (changes requested):
Thanks for your contribution. Could you please resolve conflicts?

leshakoss:
As v2 happened years ago at this point, and we're working on v3 already, I don't see the reason to expand v1->v2 migration docs, so closing this

Issue #1606, opened 2020-01-17
Rename parse's backupDate argument to referenceDate


lukewlms:
The naming "`backupDate`" seems strongly to indicate that this date will be used as a "backup" when parsing fails.  However this is not what happens - instead, failure returns Invalid Date.

It seems to me that this data parameter is being used as a reference where needed.  E.g. if a time is missing, it can be gotten from the reference; if the year is 2 digits and therefore unclear, we reference the date passed in to see what year is most likely.

You can probably see where I'm going here. 😄 To keep from causing confusion with the API I'd suggest `referenceDate` is clearer naming.

"backupDate" is never fully used as a backup because there's no case where all of its subvalues would be used.

(I checked out https://github.com/date-fns/date-fns/issues/1219 which is where this naming comes from.)

We love DateFns and evangelize it, thanks!

icleolion:
Created a PR for this as I agree.  `backupDate` was initially confusing for me too.  A different name could potentially be clearer.  The PR changes it to `referenceDate` as you've suggested, although I was tempted to go with `contextDate` instead.

kossnocorp:
I like the idea. The reference date is clearer. Thanks!

lukewlms:
https://github.com/date-fns/date-fns/pull/1607 merged to master.  

Thanks @icleolion!

PR #1607, opened 2020-01-17
Rename parse's backupDate argument to referenceDate

icleolion:
For https://github.com/date-fns/date-fns/issues/1606


dcousens:
This tripped me up when using this function myself,  I strongly agree with this PR!

PR #1615, opened 2020-01-28
Fix error message thrown by requiredArgs

jc-white:
A logical error in the ternary operator causes the error message to be malformed. Adding parentheses around the ternary condition allows the correct error message to be displayed when insufficient arguments are provided to a function that calls `requiredArgs`.


dcousens:
@jc-white maybe add a test to demonstrate this?

PR #1623, opened 2020-01-30
Aligning Czech and Slovak short date fmt with the rest of locales

sovcik:
Just wanted to align Czech and Slovak "short" and "medium" formats with the rest.
My understanding of types:

- *Short*: expecting date with full year
- *Medium*: expecting date with some form of month name
- *Long*: date with full month name
- *Full*: wee day full name, date with full month name

If we can keep formats somehow aligned then switching between locales won't be such pain...

### DIFF
--- a/src/locale/sk/_lib/formatLong/index.js
+++ b/src/locale/sk/_lib/formatLong/index.js
@@ -4,8 +4,8 @@ import buildFormatLongFn from '../../../_lib/buildFormatLongFn/index.js'
 var dateFormats = {
   full: 'EEEE d. MMMM y',
   long: 'd. MMMM y',
-  medium: 'd. M. y',
-  short: 'd. M. y'
+  medium: 'd. MMM. y',
+  short: 'dd.MM.y'
 }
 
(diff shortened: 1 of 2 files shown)


src/locale/sk/_lib/formatLong/index.js
```diff
@@ -4,8 +4,8 @@ import buildFormatLongFn from '../../../_lib/buildFormatLongFn/index.js'
 var dateFormats = {
   full: 'EEEE d. MMMM y',
   long: 'd. MMMM y',
-  medium: 'd. M. y',
-  short: 'd. M. y'
+  medium: 'd. MMM. y',
+  short: 'dd.MM.y'
```
mareksuscak: According to [the standard](https://www.unicode.org/cldr/charts/32/summary/sk.html?hide#1988), this is incorrect. Locales are not meant to be equal. There are some differences and user expectations and that's what standards try to capture.

sovcik: Thanks for pointing that out. My bad.

sovcik:
Closing this PR as incorrect.

PR #1624, opened 2020-01-31
Update examples provided for toDate, once a replacement for parse

keevan:
Documentation example was/is outdated. 

The function toDate is no longer similar to the old parse and thus examples were incorrect

### DIFF
--- a/src/parse/index.js
+++ b/src/parse/index.js
@@ -315,8 +315,9 @@ var unescapedLatinCharacterRegExp = /[a-zA-Z]/
  *   // Before v2.0.0
  *   parse('2016-01-01')
  *
- *   // v2.0.0 onward
- *   toDate('2016-01-01')
+ *   // v2.0.0 onward (toDate no longer accepts a string)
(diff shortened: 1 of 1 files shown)


dcousens:
@keevan could fix up `parseISO` in https://github.com/date-fns/date-fns/blob/62eecac8247f33dd77e059803191811a3519cc6d/src/parseISO/index.js#L33-L55 too?

dcousens:
Fixes https://github.com/date-fns/date-fns/issues/1601

keevan:
> @keevan could you fix up `parseISO` in
> 
> https://github.com/date-fns/date-fns/blob/62eecac8247f33dd77e059803191811a3519cc6d/src/parseISO/index.js#L33-L55
> too?

Hi @dcousens, I'm currently on a break, so if you or anyone else would like to make the changes that would be fine too. Otherwise I'll take a look once I get back in a few weeks.

tan75 (approved):
Thank you!

PR #1625, opened 2020-02-02
update isValid parameter from * to Date | Number

### LINKED ISSUE #1575: isValid throws console.warn when pass the string
I'm trying to check the date is valid or not with isValid and it returns false when I pass the string eg. '2016-01-01' But toDate function is still logging the two console warnings

First one:
`Starting with v2.0.0-beta.1 date-fns doesn't accept strings as arguments. Please use "parseISO" to parse strings. See: https://git.io/fjule`

Second one:
`Error
    at toDate (index.js:50)
    at isValid (index.js:68)
...
`

I don't know it's intended or not but IMHO toDate should not throw warnings when isValid used.

imballinst:
Fixes https://github.com/date-fns/date-fns/issues/1575. 

Currently, the typing in `isValid`'s function documentation expects a parameter of type `any`. This might lead to confusion as users might think they could pass anything into it.

This PR updates the parameter documentation from `*` to `Date | Number`. This is aimed to ensure let users know what parameter is actually expected by the function.

### DIFF
--- a/src/fp/index.js.flow
+++ b/src/fp/index.js.flow
@@ -355,7 +355,7 @@ declare module.exports: {
   isSunday: CurriedFn1<Date | number, boolean>,
   isThursday: CurriedFn1<Date | number, boolean>,
   isTuesday: CurriedFn1<Date | number, boolean>,
-  isValid: CurriedFn1<any, boolean>,
+  isValid: CurriedFn1<Date | number, boolean>,
   isWednesday: CurriedFn1<Date | number, boolean>,
   isWeekend: CurriedFn1<Date | number, boolean>,
   isWithinInterval: CurriedFn2<Interval, Date | number, boolean>,
--- a/src/fp/isValid/index.js.flow
+++ b/src/fp/isValid/index.js.flow
@@ -49,4 +49,4 @@ export type Duration = {
 
 type CurriedFn1<A, R> = <A>(a: A) => R
 
-declare module.exports: CurriedFn1<any, boolean>
+declare module.exports: CurriedFn1<Date | number, boolean>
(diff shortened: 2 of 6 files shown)


dcousens:
Actually,  I think the usage of `isValid` isn't for only `Number | Date`,  but I think your expectation is from the fact the function prints the following when used with anything other than a `Number | Date`:

```
Starting with v2.0.0-beta.1 date-fns doesn't accept strings as arguments. Please use `parseISO` to parse strings. See: https://git.io/fjule
Error
...
```

The problem,  I think,  is that the function should not print that warning.  The functionality of `isValid` returns `false` for say `isValid('not a date')` or `isValid(null)`.

dcousens (approved):
> Returns false if argument is Invalid Date and true otherwise. Argument is converted to Date using toDate. See toDate Invalid Date is a Date, whose time value is NaN.

LGTM.
But the tests aren't passing.

Imballinst:
The purpose of this PR seems to contradict with what we have in the tests, it seems:

```
Cannot call isValid with null bound to date because:
 • Either null [1] is incompatible with Date [2].
 • Or null [1] is incompatible with number [3].

it('treats null as an invalid date', function() {
  var result = isValid(null)
  assert(result === false)
})
```

Either we remove that test, or we label the issue as "won't fix".

1. If we remove the test, then we fully commit to the idea that `isValid` only accepts a `Date` or `Number`. (looks like more to compile time approach)
2. If we don't, then we keep it as it is -- making the scope of `isValid` as broad as possible for its users to check whether "something" is a valid Date or not. (seems like a runtime approach)

WDYT @kossnocorp?

dcousens:
@Imballinst if we broaden `isValid`,  then the documentation should be updated to reflect that too :+1:

Issue #533, opened 2017-07-17
Incorrect result across DST in differenceInDays


standy:
Calculating differenceInDays fails across DST

Example:
For London 2017-03-26, 01:00 clocks were turned forward 1 hour to 02:00
https://www.timeanddate.com/time/change/uk/london?year=2017
```
const d1 = new Date(2017, 2, 27, 1);
const d2 = new Date(2017, 2, 26, 2);
differenceInDays(d1, d2) === 1 // should be '0'
```

kossnocorp:
Thank you for reporting! We'll take a look as soon as possible.

lewell94:
Hi, is there any news on a fix for this? We have just encountered this issue using 1.30.1

victormoukhortov:
I am also having this issue - any updates?

dcousens:
I concur that `differenceInDays` ignores DST,  but `differenceInHours`, `differenceInSeconds` account for DST.

That is to say, `differenceInDays` does not operate on the local time date.

```
const A = new Date('2019-10-06 01:00:00') // day before DST
const B = new Date('2019-10-07 01:00:00') // 1 hour before DST
const C = new Date('2019-10-08 01:00:00') // day after DST
console.log({
  days: differenceInDays(B, A),
  hours: differenceInHours(B, A),
  seconds: differenceInSeconds(B, A),
})
// { days: 1, hours: 23, seconds: 82800 }

console.log({
  days: differenceInDays(C, B),
  hours: differenceInHours(C, B),
  seconds: differenceInSeconds(C, B),
})
// { days: 1, hours: 24, seconds: 86400 }
```

canvaspixels:
differenceInDays(new Date('2019-01-17T00:00:00.000Z'), new Date('2018-10-08T00:00:00.000Z')) for me returns 100 for me in my node environment (using date-fns 2.7.0) right now (current date 18/Nov) from London, UK.

When I set TZ=UTC environment variable before running the script it, it returns 101 which is what I expect.

victormoukhortov:
@canvaspixels how would you expect the difference in days to be different based on time zone? two timestamps should always have the same difference regardless of timezone.

dcousens:
@victormoukhortov each `difference*` function uses local time to calculate the difference - except `differenceInDays`.  This means rudimentary date math doesn't add up, and this difference is not documented.

victormoukhortov:
>  each difference* function uses local time to calculate the difference - except differenceInDays.

[...]

PR #1630, opened 2020-02-05
fix differenceInDays across DST (#533)

dcousens:
Resolves https://github.com/date-fns/date-fns/issues/533

I'm happy to make changes where necessary,  this was simply the simplest solution for the primitives.

### DIFF
--- a/src/differenceInDays/index.js
+++ b/src/differenceInDays/index.js
@@ -1,6 +1,4 @@
 import toDate from '../toDate/index.js'
-import differenceInCalendarDays from '../differenceInCalendarDays/index.js'
-import compareAsc from '../compareAsc/index.js'
 import requiredArgs from '../_lib/requiredArgs/index.js'
 
 /**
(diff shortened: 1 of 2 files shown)


dcousens:
For reference,  before this patch,  the `differenceInDays` function returned the following in `TZ=Australia/Sydney`
``` js
console.log(differenceInDays(c, b))
console.log(differenceInDays(b, a)) // 23 hours < 1 day
console.log(differenceInDays(c, a)) // 47 hours < 2 days
```

And for `TZ=UTC`,  it returned the expected result.

dcousens:
@kossnocorp anything I need to to help this?

src/differenceInDays/index.js
```diff
@@ -42,15 +40,9 @@ export default function differenceInDays(dirtyDateLeft, dirtyDateRight) {
   var dateLeft = toDate(dirtyDateLeft)
   var dateRight = toDate(dirtyDateRight)
 
-  var sign = compareAsc(dateLeft, dateRight)
-  var difference = Math.abs(differenceInCalendarDays(dateLeft, dateRight))
+  var result = (dateLeft - dateRight) / 86400000
```
dkozickis: Please move `86400000` into a `var MILLISECONDS_IN_DAY`

dcousens: @dkozickis moved

dcousens:
> Would be cool if you can add a bit info about the case fixed in this PR to the function description?

What about
> This function accounts for the number of 24 hour periods between two timestamps,  and thereby ignores DST changes.

dkozickis:
@dcousens sound logical to me

PR #1634, opened 2020-02-09
Add millisecond support to `Duration`, `add`, and `sub`

BehindTheMath:


kossnocorp:
Hey! Thank you a lot for your work. However, I've intentionally skipped milliseconds as they aren't present in the [ISO 8601 durations](https://en.wikipedia.org/wiki/ISO_8601#Durations). You probably have some use case in mind, which is crucial for you, but I prefer to stick to the standard and find another solution to your problem. So, please describe the issue you're trying to solve with this PR. Thanks!

BehindTheMath:
It's been a while, but I remember correctly, the use case was simple. I wanted to be able to add milliseconds to a date.

kossnocorp:
Then I believe `addMilliseconds` would suffice.

BehindTheMath:
Correct, but if I want to add other parts as well, it would require a call to `.add()` and then to `.addMilliseconds()`.

kossnocorp:
@BehindTheMath maybe but that's an imaginary case and we prioritize sticking to standards.

mathieumg:
I just got "bit" by this as well, when migrating older code. It was kind of a surprise to me that `set` supports `milliseconds` in its option but `add`/`sub` don't. It's the same use case as described by @BehindTheMath where there were multiple nested `addX` calls and which I wanted to convert into a sole `add` call, given that welcome new feature.

kossnocorp:
@mathieumg I'm sorry about that, but adding this functionality will not only break compatibility with ISO 8601 but also add a huge chunk of work for I18n teams.

dcousens:
@kossnocorp that's an acceptable answer,  but,  let's for `if ('milliseconds' in duration) throw TypeError` then?

kossnocorp:
@dcousens I'm against adding more type checkings to the library. In fact, I'm planning to remove most of them in v3 because the problem should be covered by type checking tools and not inflate the bundle size. That's the direction I want to take.

However, we might introduce another type with more values and accept it in `add` / `sub`:

```
type ExtendedDuration = Duration & { milliseconds?: number }
```

@mathieumg now I'm wondering, did you used milliseconds field and only later figured out that it doesn't work or you got it right away?

Another issue is that `set` accepts date values e.g., year, month, etc. while `add` accepts duration year**s**, month**s**. Was it a problem?

dcousens:
> type ExtendedDuration = Duration & { milliseconds?: number }

This is the best approach.   I don't think anyone was asking you to change the ISO8601 `Duration` object to include `milliseconds`,  merely,  the functions `add` and `sub` should be more-or-less relative equivalents of `set`.

mathieumg:
Sorry in the delay answering!

> However, we might introduce another type with more values and accept it in `add` / `sub`:
> 
> ```
> type ExtendedDuration = Duration & { milliseconds?: number }
> ```

I think that's suitable, as @dcousens mentioned if we can get the convenience in the utility, that's good with me, no need to tamper with ISO specs. :)

> @mathieumg now I'm wondering, did you used milliseconds field and only later figured out that it doesn't work or you got it right away?

Right away, thanks to TypeScript, but I had already migrated the whole line (with nested `addX` calls) and was wondering whether it was a missing typing since it worked for `set`. So, more of an expectation/uniformity thing. (that's when I went to look at the implementation, and also found this issue)

> Another issue is that `set` accepts date values e.g., year, month, etc. while `add` accepts duration year**s**, month**s**. Was it a problem?

Not really a problem, I guess in the end both output a set `Date` at a specific moment so you could want to have a date at a specific number of milliseconds, but you could also want to add/subtract a specific number of milliseconds to an existing date.

PR #1638, opened 2020-02-14
Add sliceInterval (closes #1508)

### LINKED ISSUE #1508: Feature request: subtract interval from a list of intervals
Please implement a new Interval Helper which subtracts an interval from a list of intervals
https://date-fns.org/v2.6.0/docs/Interval

Given an array of Interval objects and a ```remove``` Interval object

have a new helper ```subtractFromIntervals(intervals, remove) ```
which returns an array of Interval objects which represents ```intervals - remove```

eg
```
[ {start: '2019-10-10 9:00', end: '2019-10-10 18:00'}]
minus
{start: '2019-10-10 12:00', end: '2019-10-10 14:00'}
equals
[ {start: '2019-10-10 9:00', end: '2019-10-10 12:00'}, {start: '2019-10-10 14:00', end: '2019-10-10 18:00'}]
```

Thanks!

edumoreira1506:
This PR resolves issue #1508

### DIFF
--- a/src/sliceInterval/index.js
+++ b/src/sliceInterval/index.js
@@ -1,0 +1,40 @@
+import requiredArgs from '../_lib/requiredArgs/index.js'
+import max from '../max/index.js'
+import min from '../min/index.js'
+
+/**
+ * @name sliceInterval
+ * @category Second Helpers
+ * @summary  Returns an Interval objects which represents intervals - remove.
+ *
+ * @description
+ *  Returns an array of Interval objects which represents intervals - remove.
+ *
+ * @param {[Interval, Interval]} intervals
+ * @param {Interval} interval
+ * @returns {Array} an array of Interval objects which represents intervals - remove
+ * @throws {TypeError} 2 arguments required
+ *
+ */
+export default function sliceInterval(intervals, interval) {
+  requiredArgs(2, arguments)
+
+  return intervals.map(item => {
+    const minStart = min([item.start, interval.start])
+    const maxStart = max([item.start, interval.start])
+
+    const minEnd = min([item.end, interval.end])
+    const maxEnd = max([item.end, interval.end])
+
+    return [
+      {
+        start: minStart,
+        end: maxStart
+      },
+      {
+        start: minEnd,
+        end: maxEnd
+      }
+    ]
+  })
+}
--- a/src/sliceInterval/test.js
+++ b/src/sliceInterval/test.js
@@ -1,0 +1,29 @@
+// @flow
+/* eslint-env mocha */
+
+import assert from 'power-assert'
+import sliceInterval from '.'
+
+describe('sliceInterval', function() {
+  it('return a new array with subtracted intervals', function() {
+    var interval = {
+      start: new Date(2019, 10, 10, 9, 0),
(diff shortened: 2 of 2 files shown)


src/subtractFromIntervals/test.js
```diff
@@ -0,0 +1,31 @@
+// @flow
+/* eslint-env mocha */
+
+import assert from 'power-assert'
+import subtractFromIntervals from '.'
+
+describe('subtractFromIntervals', function() {
+  it('return a new array with subctrated intervals', function() {
```
dcousens: `s/subctrated/subtracted`

dcousens:
The naming, terminology and use of 'subtraction' here isn't sensible, personally.

The type of the function is `subtractFromIntervals(dates: [DateRange],  operand: DateRange) => [DateRange + 1]`... what?

My understanding from the #1508 description is that this isn't a subtraction,  but better phrased as a "cut out of" or "cut and splice".

As `Array.prototype.splice` is probably the closest to this action,  I'd use that phrasing.
I'd also change the type of the function to strictly be:

```
spliceInterval(range: DateRange,  operand: DateRange) => [DateRange, DateRange]
```

Instead of a reduction over an input array.

IMHO.

edumoreira1506:
@dcousens Nice! I already fixed

dcousens (approved):
LGTM :)

src/spliceInterval/index.js
```diff
@@ -0,0 +1,40 @@
...
+ * @name spliceInterval
+ * @category Second Helpers
+ * @summary  Returns an Interval objects which represents intervals - remove.
+ *
+ * @description
+ *  Returns an array of Interval objects which represents intervals - remove.
+ *
+ * @param {Array} range
+ * @param {Date} operand
+ * @returns {Array} an array of Interval objects which represents intervals - remove
```
kossnocorp, line 15: 1. Please use proper types `{Interval}` and `{[Interval, Interval]}` (not sure about this one, if it doesn't work, please use `{Array}`)
2. Use proper naming (interval instead of a range, etc.)
3. Please write clear description of how the arguments are used and what they represent (in case of the returned value).

src/spliceInterval/index.js
```diff
@@ -0,0 +1,40 @@
...
+ * @throws {TypeError} 2 arguments required
+ *
+ */
```
kossnocorp, line 18: `@example` section with examples is missing.

src/spliceInterval/test.js
```diff
@@ -0,0 +1,29 @@
...
+      {
+        start: new Date(2019, 10, 10, 9, 0),
+        end: new Date(2019, 10, 10, 12, 0)
+      },
+      {
+        start: new Date(2019, 10, 10, 14, 0),
+        end: new Date(2019, 10, 10, 18, 0)
+      }
+    ])
+  })
```
kossnocorp, line 28: Please cover more cases:

1. What happens if intervals don't overlap?
2. What happens if the second interval includes the first one but not vice versa?
3. What happens if the second interval ends after the first one?
4. Test for exceptions.

kossnocorp (changes requested):
Thank you for your work! However, it requires polishing before I can accept it.

Issue #1637, opened 2020-02-14
setDay's argument "options.weekStartsOn" not working if date is a Sunday.


BoomDev:
In :
setDay(**date**, day, [options])

If "date" is a Sunday, the option "weekStartsOn" is not working properly.

Stackblitz: https://stackblitz.com/edit/date-fns-playground-53c1s5?file=index.ts

PR #1639, opened 2020-02-14
Fix issue #1637: setDay with weekStartsOn != 0

BoomDev:
As demonstrated in the issue, setDay(date, day, [option]) returns an
incorrect value if date is on a Sunday and option.weekStartsOn is != 0.
This includes a fix (there's probably a nicer way to code this) and
2 corresponding test cases.

I also found 4 test cases that I believe to be incorrect, all using
the argument {weekStartsOn: 1}. I hope I'm not mistaken here, but I
changed the test cases accordingly.

- All tests completed

- Build script run

- Added entry to Unreleased section of CHANGELOG

### DIFF
--- a/src/setDay/test.js
+++ b/src/setDay/test.js
@@ -49,6 +49,20 @@ describe('setDay', function() {
     assert.deepEqual(result, new Date(2014, 8 /* Sep */, 7))
   })
 
+  it('specifies Monday as the first day of the week', function() {
+    var result = setDay(new Date(2014, 8 /* Sep */, 7), 1, {
+      weekStartsOn: '1'
+    })
+    assert.deepEqual(result, new Date(2014, 8 /* Sep */, 1))
+  })
+
+  it('specifies Tuesday as the first day of the week', function() {
+    var result = setDay(new Date(2014, 8 /* Sep */, 7), 1, {
+      weekStartsOn: '2'
+    })
+    assert.deepEqual(result, new Date(2014, 8 /* Sep */, 8))
+  })
+
   context('the day index is more than 6', function() {
     it('sets the day of the next week', function() {
       var result = setDay(new Date(2014, 8 /* Sep */, 1), 8)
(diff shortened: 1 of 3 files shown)


src/setDay/test.js
```diff
@@ -49,6 +49,20 @@ describe('setDay', function() {
...
+  it('specifies Monday as the first day of the week', function() {
+    var result = setDay(new Date(2014, 8 /* Sep */, 7), 1, {
+      weekStartsOn: '1'
+    })
+    assert.deepEqual(result, new Date(2014, 8 /* Sep */, 1))
+  })
+
+  it('specifies Tuesday as the first day of the week', function() {
+    var result = setDay(new Date(2014, 8 /* Sep */, 7), 1, {
+      weekStartsOn: '2'
```
dcousens: Why a string?

dcousens:
@BoomDev please see my update to the examples in https://github.com/BoomDev/date-fns/pull/1

src/setDay/test.js
BoomDev: Oh, good catch. In fact I copied the case just above (it 'implicitly converts options') and didn't realize it was using a string... So the answer is: no reason. I'll change it in my 2 cases.

trollepierre:
🙏 @dcousens : can you approve this PR since the change has been taken into consideration? 
+ I opened a new bug that count on this fix

trollepierre:
@BoomDev : have you checked if "lastDayOfWeek startOfWeek differenceInCalendarWeeks" are fixed?

dcousens:
@trollepierre I am not a maintainer,  I am merely a fellow user wanting to help :+1:

BoomDev:
> @BoomDev : have you checked if "lastDayOfWeek startOfWeek differenceInCalendarWeeks" are fixed?

@trollepierre I have not.

dkozickis:
Thank you @BoomDev and @dcousens, this seems GTG. 
Tests might be failing on CI because of DST/Timezones, not sure @kossnocorp

kossnocorp:
@dkozickis yeah, some tests were failing because of a DST issue that was fixed in master. I've rebased the branch but there's still a couple failing tests in the changed function. I'm on it.

Issue #1230, opened 2019-06-27
Include more details in error message


benmccann:
I'm using 2.0 beta 1 and received the error message below:

> RangeError: Use `yyyy` instead of `YYYY` for formatting years

I think the error message could be improved both by including the full format string as well as the input which is being attempted to be parsed or formatted. This error is happening in the internals of another component I'm using and unfortunately this isn't enough detail to help me track down the root cause

kossnocorp:
You're right, that could help!

PR #1641, opened 2020-02-19
Refactor/throw messages

belgamo:
Improvement based on issue #1230


belgamo:
> I like the idea but you've forgot to update the usage of `throwProtectedError` in `parse`.
> 
> Also, instead of:
> 
> ```
> Use `dd` instead of `DD` for formatting days of the month using yyyy-MM-DD; see: https:\/\/git.io\/fxCyr
> ```
> 
> I prefer:
> 
> ```
> Use `dd` instead of `DD` (in `yyyy-MM-DD`) for formatting days of the month; see: https:\/\/git.io\/fxCyr
> ```

Yes, I had forgotten that, but I notice and already fixed it! I ask you to review it again. I'll improve the message. Thank you!

Issue #1642, opened 2020-02-24
differenceInYears calculates wrongly when date is last day of february leap year


vonsky104:
When you calculate difference in years by 'differenceInYears' function, the returned value is wrong for given dates: 2005-03-01 and 2024-02-29.

Example:
`differenceInYears(parseISO('2024-02-29'), parseISO('2005-03-01'))` equals 19, when in reality it should be 18, the same result which this function returns when you give 2024-02-28.

Possible problem:
Date-Fns calculates in this case both same dates, since by the logic of the lib - 2024-02-29 does not exist on this year , it calculates for 2024-03-01, so for the first found valid date, instead of 2024-02-28. This is wrong because, for example, person born 29 Feb 2004, becomes aged 21 on 1 Mar 2025, so it can create big problems (and in case of my company and its app - it does).

I had to create very hack'y workaround which checks given date and if its a leap year, last day of February, it sub one day. (Also I check if day and month are equal for given dates).

PR #1654, opened 2020-03-03
Fix issue: #1642 - differenceInYears leap day fix

leepowellcouk:
As reported in the bug report, `differenceInYears` does not work as expected when dealing with a left hand date which is a leap day.

Consider the following:
```
const result = differenceInYears(new Date(2024, 1, 29), new Date(2005, 02, 01));
```
This would have returned the result of `19`. This is incorrect and should be `18`. The reason for this is the following:
```
// Line 36
dateLeft.setFullYear(dateLeft.getFullYear() - sign * difference)
```
In this instance, if the `sign * difference` end up reassigning the date to a non-leap year i.e. 2023, then internally the date is advanced to `1st March 2023`. While this is the default behaviour of the spec for this utility it is creating invalid results.

The setting of the year was only meant to align the two dates to the same year, so the rest of the date could be compared to see if it fell within a full year. To work around this I'm proposing we assign BOTH dates to a leap year to prevent any incorrect date reassignment.

### DIFF
--- a/src/differenceInYears/index.js
+++ b/src/differenceInYears/index.js
@@ -2,6 +2,7 @@ import toDate from '../toDate/index.js'
 import differenceInCalendarYears from '../differenceInCalendarYears/index.js'
 import compareAsc from '../compareAsc/index.js'
 import requiredArgs from '../_lib/requiredArgs/index.js'
+import isLeapYear from '../isLeapYear'
 
 /**
  * @name differenceInYears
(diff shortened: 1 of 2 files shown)


leepowellcouk:
> tests LGTM
> 
> Maybe add `2004-02-29` v `2002-02-28` (should be `2`).

Added.

dcousens:
Maybe `2008-02-29 - 2004-02-29 = 4 years` too

leepowellcouk:
> Maybe `2008-02-29 - 2004-02-29 = 4 years` too

Also done 😄

leepowellcouk:
Is there anything else this needs to get it merged? Thanks.

dcousens:
@leepowellcouk you're waiting on @kossnocorp for review now,  I am not a maintainer :)

dkozickis:
LGTM, I'll see if I can get @kossnocorp to merge it at some point :)

kossnocorp:
Thanks everyone! I'm working on shipping `v2.11.0`.

dkozickis:
Fixes #1642

vuminhluan:
What about `2021-02-28 - 2020-02-29`: [codesandbox](https://codesandbox.io/s/date-fns-differenceinyears-z6b0c?file=/src/App.js)

Issue #1661, opened 2020-03-12
Format works incorrectly with GMT-0752/GMT-0456 and similar timezones


eugene-platov:
Hello
Found a problem with `format` fn:
System timezone is set to `America/Los_Angeles`
Using Chrome 80.0.3987.132
`format(new Date("1850-01-01T00:00:00"), "HH:mm:ss")` results in `00:01:00` instead of `00:00:00`

Seems like the problem is in the next function: https://github.com/date-fns/date-fns/blob/fadb737fe714c7148163a02ac8ee955cc343cac9/src/_lib/getTimezoneOffsetInMilliseconds/index.js#L14

For this date (01/01/1850) it returns `28318000` milliseconds (or `471.9666` minutes), while the real offset in this timezone is `472.9666` minutes (-07:52:58)

Getting similar behavior in GMT-0456 (`America/New_York`), GMT-0344 (`America/Miquelon`), GMT-0209 (`America/Noronha`) timezones

Tried with date-fns@2.7.0 and date-fns@2.10.0, results are the same

dkozickis:
@eugene-platov are you willing to submit a PR against this one? Looks like rounding errors between Chrome timezone DB (and how Chrome returns `new Date()`) and how `getTimezoneOffsetInMilliseconds` affects final calculation.

Would you also happen to have a correct link to source of information with historical timezones? 

![image](https://user-images.githubusercontent.com/769698/76605241-ceb7e480-6518-11ea-99c7-5ad1c8d5eac9.png)

eugene-platov:
@dkozickis Yes, I'd like to :)
But I'm not sure about rounding errors, more likely calculations in `getTimezoneOffsetInMilliseconds` are slightly incorrect for left from GMT timezones, because for GMT+n timezones everything works fine
I'll look into in and will submit a PR

eugene-platov:
@dkozickis Created a PR with fix ^

kossnocorp:
Closed by #1666

PR #1666, opened 2020-03-13
Fixed timezone offset getting for UTC-n timezones

eugene-platov:
These changes fixes the next issue: https://github.com/date-fns/date-fns/issues/1661

### DIFF
--- a/src/_lib/getTimezoneOffsetInMilliseconds/index.js
+++ b/src/_lib/getTimezoneOffsetInMilliseconds/index.js
@@ -1,5 +1,9 @@
 var MILLISECONDS_IN_MINUTE = 60000
 
+function getDateMillisecondsPart(date) {
+  return date.getTime() % MILLISECONDS_IN_MINUTE
+}
+
 /**
  * Google Chrome as of 67.0.3396.87 introduced timezones with offset that includes seconds.
  * They usually appear for dates that denote time before the timezones were introduced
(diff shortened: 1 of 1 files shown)


src/_lib/getTimezoneOffsetInMilliseconds/index.js
```diff
@@ -16,9 +16,12 @@ export default function getTimezoneOffsetInMilliseconds(dirtyDate) {
...
   date.setSeconds(0, 0)
   var millisecondsPartOfTimezoneOffset = date.getTime() % MILLISECONDS_IN_MINUTE
+  var hasNegativeUTCOffset = baseTimezoneOffset > 0
 
   return (
     baseTimezoneOffset * MILLISECONDS_IN_MINUTE +
-    millisecondsPartOfTimezoneOffset
+    (hasNegativeUTCOffset && millisecondsPartOfTimezoneOffset !== 0
+      ? MILLISECONDS_IN_MINUTE + millisecondsPartOfTimezoneOffset
+      : millisecondsPartOfTimezoneOffset)
```
dkozickis: Could you please extract this part of code into separate `var`?

eugene-platov:
@dkozickis Well, some tests were failing if they are run in any UTC-n timezone :)
But I'll try, it will require to mock `getTimezoneOffset` somehow

dkozickis:
Oh... I thought Travis does a full run with all timezones @kossnocorp?

kossnocorp:
@dkozickis it actually does: https://travis-ci.org/github/date-fns/date-fns/jobs/662076035#L1015-L1055

@eugene-platov what is the TZ value that causes the failure? It would be great to add some regression tests.

eugene-platov:
@kossnocorp Thank you
Guess the reason that tests in ci are not failing is that they are running in timezones that doesn't have seconds offsets
If I set `America/Los_Angeles` timezone on my computer, all dates before Nov 19th 1883 will get additional minute and some tests will fail
After this fix the are not failing
Screenshot at Mar 16 02-31-16

Issue #1663, opened 2020-03-12
Inconsistent typing across getDay fn and weekStartsOn option.


oakhan3:
Hi,

I have run into a typing (TypeScript) issue when trying to take the output of `getDay` and place it in the `weekStartsOn` option on the `isThisWeek` fn (or any fn that accepts that option):

Here is an example:
```typescript
isCurrentOrFutureWeek() {
  const startOfWeek = parseISO("2020-03-10");
  const dayOfWeek = getDay(startOfWeek);

  return (
    isThisWeek(startOfWeek, { weekStartsOn: dayOfWeek }) || startOfWeek >= endOfToday()
  );
}
```

Typescript complains with the following:
```
Argument of type '{ weekStartsOn: number; }' is not assignable to parameter of type '{ locale?: Locale | undefined; weekStartsOn?: 0 | 2 | 1 | 6 | 5 | 4 | 3 | undefined; }'.
  Types of property 'weekStartsOn' are incompatible.
    Type 'number' is not assignable to type '0 | 2 | 1 | 6 | 5 | 4 | 3 | undefined'.
```

This is because `getDay` has been defined to return a `number` and `weekStartsOn` has been typed to take in `0 | 2 | 1 | 6 | 5 | 4 | 3`.

This can be circumvented by asserting types or loosening types on getDay in my code:
```
const dayOfWeek: any = getDay(startOfWeek);
const dayOfWeek = getDay(startOfWeek) as 0|1|2|3|4|5|6;
```

But it would be ideal if the return value for `getDay` was typed as a number from 0 to 6 so that this would not be necessary.

Links to relevant code:

https://github.com/date-fns/date-fns/blob/master/src/getDay/index.js
https://github.com/date-fns/date-fns/blob/master/src/isThisWeek/index.js

Thank you

dkozickis:
Are you interested in submitting a PR for this @oakhan3? 

You can see how `weekStartsOn` param for `isThisWeek` is typed (with JSDoc) here

https://github.com/date-fns/date-fns/blob/16a561df59b9a015517fce568ffd929f46961f82/src/isThisWeek/index.js#L23

and apply it to return type of `getDay` here

https://github.com/date-fns/date-fns/blob/16a561df59b9a015517fce568ffd929f46961f82/src/getDay/index.js#L17

that should work I think

oakhan3:
Sure thing 👍

oakhan3:
#1668 should resolve this

[...]

PR #1668, opened 2020-03-13
Fix inconsistent typing across getDay fn and weekStartsOn option. #1663

oakhan3:
It looks like backupDate was changed to referenceDate at some point in master but FP functions, typings and indices were never rebuilt via `./scripts/build/build.sh`.

### DIFF
--- a/src/getDay/index.js
+++ b/src/getDay/index.js
@@ -14,7 +14,7 @@ import requiredArgs from '../_lib/requiredArgs/index.js'
  * - [Changes that are common for the whole library](https://github.com/date-fns/date-fns/blob/master/docs/upgradeGuide.md#Common-Changes).
  *
  * @param {Date|Number} date - the given date
- * @returns {Number} the day of week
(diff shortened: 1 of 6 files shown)


src/getDay/index.js
```diff
@@ -14,7 +14,7 @@ import requiredArgs from '../_lib/requiredArgs/index.js'
  * - [Changes that are common for the whole library](https://github.com/date-fns/date-fns/blob/master/docs/upgradeGuide.md#Common-Changes).
  *
  * @param {Date|Number} date - the given date
- * @returns {Number} the day of week
+ * @returns {0|1|2|3|4|5|6} the day of week
```
oakhan3: @kossnocorp I believe this is what you are referring to - I did commit the JSDoc change. I generated all other changes after via `./scripts/build/build.sh`.

kossnocorp: @oakhan3 whoa, I've totally missed that. Thank you for pointing it!

oakhan3: no worries at all 👍

PR #1688, opened 2020-03-26
Makes parse optional params optional

Domino987:
As seen in the README, the parameters for parse are optional.
This is not reflected in the types.


dcousens:
@Domino987 I don't think these are optional.

From the [documentation](https://date-fns.org/v2.11.1/docs/parse#arguments)
> parse(dateString, formatString, referenceDate, [options])

That implies only `options` is optional.
I assume that is why you closed this.

PR #1701, opened 2020-03-30
Fix minulý štvrtok in Slovak locale (fixes #1699)

### LINKED ISSUE #1699: SK locale formatRelative returns "minulú štvrtok" instead of "minulý štvrtok"
I have a problem with Slovak **formatRelative** formating

**Now** 2020-03-30
**Relative date** is: 2020-03-26

```js
// Import
import { sk as locale } from 'date-fns/locale'

// Prepare dates
const reference = new Date('2020-03-26 22:26:00')
const now = new Date()

// Return
formatRelative(reference, now, { locale })  // minulú štvrtok o 22:26
```

it returns "**minulú** štvrtok o 22:26" 
it should be "**minulý** štvrtok o 22:26"

Can you check:
@mareksuscak 
@leshakoss

leshakoss:
See #1699

### DIFF
--- a/src/locale/sk/_lib/formatRelative/index.js
+++ b/src/locale/sk/_lib/formatRelative/index.js
@@ -16,7 +16,7 @@ function lastWeek(day) {
 
   switch (day) {
     case 0: /* Sun */
-    case 4: /* Wed */
+    case 3: /* Wed */
     case 6 /* Sat */:
       return "'minulú " + weekday + " o' p"
     default:


mareksuscak:
FWIW, you can use [patch-package](https://www.npmjs.com/package/patch-package) + [the diff](https://patch-diff.githubusercontent.com/raw/date-fns/date-fns/pull/1701.diff) until we push out the next release.

PR #1702, opened 2020-03-31
Change to September in addDays example

tobiastornros:
Change from August to September in example for addDays so that the example reflects example information.


dkozickis:
Hi @tobiastornros!

Months are zero indexed in JS world, so `8` corresponds to Sep, and `9` to Oct.

Here's the reference - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date

tobiastornros:
Hi @dkozickis 

Wow! You learn something new everyday 👍 Thanks

PR #1713, opened 2020-04-13
Add `getISODurationString` function

rhlowe:
This is a function I wanted for a microformat I was implementing in my own website but it didn't exist. I hope that it is a useful addition to the core date-fns library.

This _is_ an implementation of the [ISO8601 Duration Format](https://en.wikipedia.org/wiki/ISO_8601#Durations), thanks for pointing that out @dcousens.


dcousens:
For others information,  this appears to be an [ISO8601 formatted duration](https://en.wikipedia.org/wiki/ISO_8601#Durations)

rhlowe:
Thanks for the feedback @kossnocorp! I think I got all your feedback integrated, I like the way the code looks a little better now.

src/formatISODuration/test.js
```diff
@@ -0,0 +1,78 @@
+// @flow
+/* eslint-env mocha */
+
+import assert from 'power-assert'
+import formatISODuration from '.'
+import intervalToDuration from '../intervalToDuration'
+
+describe.only('formatISODuration', function() {
```
dcousens: no `only`

rhlowe: Yeah! That was a big 'whoopsie' on my part.

kossnocorp:
Thank you so much!

PR #1722, opened 2020-04-16
Fix code comments

tricknotes:
These functions don't depend on `Date.now()`.
This information is useful for mocking.


kossnocorp:
Thanks!

Issue #1510, opened 2019-10-29
Adding one month results in 29 days


otaviobps:
I'm using the `addMonths` function to add 1 month to a given date. The function is inside a `forEach` but that doesn't add nothing to the bug.

Receiving the date `2019-03-01T01:00:00.000Z` and adding 1 month to it `dateFns.addMonths(actualDate, 1)` (actualDate is the referred date object) results in `2019-04-29T01:00:00.000Z` when it should result `2019-05-01T01:00:00.000Z`. I tried starting in 2019-01-01 and add 1 month to each new date. Here are the resulting dates:

```
2019-02-01T00:00:00.000Z
2019-03-01T01:00:00.000Z
2019-03-29T01:00:00.000Z
2019-04-29T01:00:00.000Z
2019-05-29T01:00:00.000Z
2019-06-29T01:00:00.000Z
2019-07-29T01:00:00.000Z
2019-08-29T01:00:00.000Z
2019-09-29T01:00:00.000Z
2019-10-29T01:00:00.000Z
```

Beginning in 2018-01-01, the resulting dates are:

```
2018-02-01T00:00:00.000Z
2018-03-01T01:00:00.000Z
2018-03-29T01:00:00.000Z
2018-04-29T01:00:00.000Z
2018-05-29T01:00:00.000Z
2018-06-29T01:00:00.000Z
2018-07-29T01:00:00.000Z
2018-08-29T01:00:00.000Z
2018-09-29T01:00:00.000Z
2018-10-29T01:00:00.000Z
2018-11-29T00:00:00.000Z
2018-12-29T00:00:00.000Z
2019-01-29T00:00:00.000Z
2019-03-01T01:00:00.000Z
2019-03-29T01:00:00.000Z
2019-04-29T01:00:00.000Z
2019-05-29T01:00:00.000Z
2019-06-29T01:00:00.000Z
2019-07-29T01:00:00.000Z
2019-08-29T01:00:00.000Z
2019-09-29T01:00:00.000Z
2019-10-29T01:00:00.000Z
```

My `nodejs` version is 12.12.0 and `date-fns` version is `2.6.0`

leshakoss:
@otaviobps hi, could you please tell us what timezone are you in?

otaviobps:
Hi @leshakoss, I'm in GMT-3

tanoabeleyra:
We recently realized about this bug in one project. We're also in the GMT-3 timezone.

dkozickis:
[...]

PR #1751, opened 2020-05-03
WIP: Double the speed of addMonths() while reducing its dependencies & code size

justingrant:
_EDIT: Oops, my optimization may have a flaw: if the last day of the new month is a DST transition, then the time on that day may be reset to avoid the "dead hour" or "ambiguous hour" caused by the DST transition. Then, when the date is set to the actual day-of-month, the time needs to be put back to what it was originally. Anyway, I'm marking this PR as a work-in-progress until I can add more tests to capture this case, and to fix it if tests fail. This might mean that this optimization isn't really helpful and I'll close the PR if so._

While investigating #1510 I spent some time optimizing `addMonths`. This PR: 

* Simplifies the code to make it more readable
* Removes the dependency on `getDaysInMonth`
* Removes several lines of code and multiple Date method calls.
* Removes creation of an extra `Date` object
* The changes above doubled the speed of this method (according to `yarn benchmark`) from 526410 ops/sec to 1048890 ops/sec on my 2015 MacBook Pro.

All tests continue to pass.  I also added two passing test cases to ensure that `addMonths` works properly across a DST-start and DST-end boundary.

### DIFF
--- a/src/add/index.js
+++ b/src/add/index.js
@@ -57,10 +57,13 @@ export default function add(dirtyDate, duration) {
   const seconds = 'seconds' in duration ? toInteger(duration.seconds) : 0
 
   // Add years and months
-  const dateWithMonths = addMonths(toDate(dirtyDate), months + years * 12)
+  const date = toDate(dirtyDate)
+  const dateWithMonths =
+    months || years ? addMonths(date, months + years * 12) : date
 
   // Add weeks and days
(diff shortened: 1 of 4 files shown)


justingrant:
Abandoning this PR because other bugs are present in addMonths that should be addressed.  Will open new PR shortly.

PR #1752, opened 2020-05-04
Updated dutch abbreviation

JorenVos:
Fixed "mei" abbreviation


kossnocorp:
Thank you for your help. I committed the change myself as I couldn't be bothered to checkout to your branch, run `yarn locale-snapshots` to fix the tests, and committing the result.

pataar:
Hey, I have a question about this change. Why is it that May (mei) is the only month without an ending dot? Seems a bit inconsistent to me

JorenVos:
Hi,

It has nothing to do with inconsistency, "mei" is just the full month name
in dutch (like May) and shouldn't be abbreviated for three characters.

Kind regards,

Op ma 11 mei 2020 om 12:05 schreef Pataar <notifications@github.com>:

pataar:
Yes, I realized shortly afterwards and deleted my comment.

Issue #1083, opened 2019-02-04
Starting with v2.0.0-beta.1 date-fns doesn't accept strings as arguments. Please use `parseISO` to parse strings


TheoMer:
So, until I deployed today I was using version ^2.0.0-alpha.7 with no issues, but then got the following error:

```
Starting with v2.0.0-beta.1 date-fns doesn't accept strings as arguments. Please use `parseISO` to parse strings
```
I upgraded to the latest version and am now getting an `parseISO is not defined` error. My code is as follows:

```
import { format } from 'date-fns';

<span>{format(parseISO(order.createdAt), 'MMMM d, YYYY h:mm a', { awareOfUnicodeTokens: true })}</span>
```

What am I overlooking here?

TrySound:
@TheoMer 
```
import { parseISO, format } from 'date-fns';
```

TheoMer:
@TrySound I tried that but got

```
'date-fns' does not contain an export named 'parseISO'
```

TrySound:
This is weird. Make sure you don't have two versions of date-fns in your bundle.
Also where this error coming from?

TheoMer:
@TrySound So I uninstalled and re-installed date-fns, and got this:

```
+ date-fns@1.30.1
added 1 package from 1 contributor and audited 24795 packages in 1069.658s
```

I'm not sure what's happening with npm but regressing from `2.0.0-alpha.7` to `1.30.1` doesn't seem right to me, correct?

TrySound:
`yarn add date-fns@next`

TheoMer:
@TrySound That resoloved it. Many thanks.

niquepa:
Had a similar issue, in my case the error I got was:

`Uncaught TypeError: Object(...) is not a function`

After executing `yarn add date-fns@next` got fixed, thanks.

kossnocorp:
@niquepa this you should install `date-fns@2` instead because `@next` will install the latest beta `2.0.0-beta.5`, while the newest version is `2.12.0`.

aounleonardo:
I'm using `^2.13.0` and am getting this warning when I just use:
```
import { addHours } from "date-fns";
//...
addHours(new Date(), <number>)
```

Is this expected?

kossnocorp:
@aounleonardo no, it's not expected and rather impossible for the snippet that you've provided. Please check other places, including the 3rd-party code.

PR #1754, opened 2020-05-05
Fix 2.11->2.12 DST issues with differenceInDays (fix #1750)

### LINKED ISSUE #533: Incorrect result across DST in differenceInDays

### LINKED ISSUE #1750: differenceInDays no longer matches addDays (breaking change in 2.12?)
_UPDATE: I did more investigation and made a few revisions to clarify the original report._ 

I just spent a day tracking down a problem in my app that turned out to be caused by a breaking change or bug in date-fns 2.12.0: the change to `differenceInDays` behavior from #1630.  `differenceInWeeks` is also affected, but `differenceInMonths` is not. The following repro illustrates the problem.  All reproes worked as expected on 2.11.  Live repro: https://codesandbox.io/s/silly-waterfall-85bct?file=/src/index.tsx

```js
const original = new Date(2014, 2, 1);

const addHours28x24 = addHours(original, 28 * 24);
const diffHours28x24 = differenceInHours(addHours28x24, original);
// Actual: 672 (expected: 672)

const addDays28 = addDays(original, 28);
const diffDays28 = differenceInDays(addDays28, original);
// Actual: 27 (expected: 28)

const addWeeks4 = addWeeks(original, 4);
const diffWeeks4 = differenceInWeeks(addWeeks4, original);
// Actual: 3 (expected: 4)

const addMonths1 = addMonths(original, 1);
const diffMonths1 = differenceInMonths(addMonths1, original);
// Actual: 1 (expected: 1)
```

I had a few concerns about this  new behavior:
1. It doesn't match `addDays` behavior. If you add N days to a date-time then `differenceInDays` will return N most of the time but sometimes will return N-1 if the dates are across a DST-start boundary (but not if the dates are across a DST-end boundary!).  This is a breaking change for devs who are expecting that `addDays` and `differenceInDays` would work reversibly, as they did before 2.12.  Note that `differenceInWeeks` has the same problem (it uses `differenceInDays` under the covers), although `differenceInMonths` does not. 
2. Conceptually, a "day" is NOT the same as a 24-hour period because DST makes some days 23 or 25 hours long. A non-technical person (like our users) would expect that the difference between noon on the last day of winter time and noon on the first day of DST is "one day".
4. Devs who want to measure the difference as only-24-hour periods have an easy workaround: `(differenceInHours(date2,date1)/24)|0`. But in 2.12, I don't think the previous behavior has an available workaround in date-fns.

As I understand it, days (and larger) units are fundamentally different from hours (and smaller) units in how they relate to time zones and DST.
* Calculations using hours and smaller units really only make sense in UTC. Calculations using hours (and smaller) units should first convert to UTC (or ticks) and then the calculation is simple arithmetic that can ignore DST (and time zones, calendars, leap days, etc.)
* Calculations using days and larger units, however, are _always_ relative to a time zone because they depend on date boundaries that are only defined in context of a time zone and are inherently affected by DST changes. For this reason, it's not OK to convert dates to UTC and expect to get accurate day (or larger) math to work. Days (and larger) units _must always_ be calculated within the context of a time zone.

Do those assumptions match intended behavior?  If so, I'd be happy to PR (let me know if it's welcome) a solution like this: (or something similar you think is better)
- revise the code and test changes from #1630 so that the original bug reported in #533 is addressed but also that the mainstream case still works, e.g. "noon on the day before DST starts to noon on the day DST starts" should return 1.
- modify the docs to: 
   - clarify the docs that `differenceInDays` / `differenceInWeeks` take time zones into account, and explain why calculating `differenceInDays` / `differenceInWeeks` between dates from different time zones is a bad idea. 
   - add code samples for both APIs to be one that spans a DST boundary in DST-observing countries in the northern hemisphere, e.g.:
```js
// How many full days are between
// 1 March 2020 0:00 and 1 April 2020 0:00 ?
// Note that this will return the same result
// regardless of time zone.
var result = differenceInDays(
  new Date(2020, 3, 1),
  new Date(2020, 2, 1)
)
//=> 31
```
   - Add either documentation text and/or a code sample to illustrate the workaround to calculate 24-hour periods only: e.g.
```js
//
// How many 24-hour periods are between
// 1 March 2020 0:00 and 1 April 2020 0:00 ?
// Note that for time zones that start daylight 
// savings time in March, this will return 30
// not 31 because the first day of DST is only
// 23 hours long.
var result = (differenceInHours(
  new Date(2020, 3, 1),
  new Date(2020, 2, 1)
)/24) | 0
//=> 30 in time zones where DST starts in March
//=> 31 in other time zones
// `| 0` is a fast way to truncate (round towards zero)
```

I read through #533 and I'll take a look at that repro case (where the time is the later edge of a DST-start boundary) to see if there's a way to solve that case while not breaking cases that were working before. 

I admit that I couldn't fully understand the consistency discussion in #533. To me it seems like the pre-2.12 behavior was *more* consistent, because in 2.11 `addDays` was consistent with `differenceInDays`, and `differenceInDays` was also consistent with `differenceInMonths`.

Let's discuss.

justingrant:
As documented in #1750, date-fns 2.12 (specifically #1630) contained a breaking change that made `differenceInDays` stop matching `addDays` across a DST boundary.  This PR replaces #1630 with an implementation that fixes #533 (like #1630 did) without causing #1750 (like #1630 also did).

This PR: 
* Fixes #1750 by restoring the 2.11 implementation of `differenceInDays` (so should code-review against 2.11 implementation) but modifying it to fix #533 without triggering #1750. It does this by replacing the use in `differenceInDays` of `compareAsc` with a new local function (`compareLocalAsc`) that compares using local time not UTC timestamps. This change prevents errors when one of the times is on a DST boundary or contains an ambiguous local time due to DST.
* Adds new DST tests for `differenceInDays` to validate the work above. These tests will work in any timezone, thanks to new shared test code that detects DST start in the local timezone. All DST tests in this PR are relative to the DST start/end dates detected for the current timezone. These tests will be skipped in non-DST time zones. 
* Validates that all tests run OK in the America/Los_Angeles, America/Sao_Paulo, Australia/Sydney, and Europe/London timezones. This was done by temporarily adding a TZ env to the `yarn test` script command and then manually running tests under each TZ.
* Adds new JSDoc content and samples for `differenceInDays` to document DST behavior and to explain how to get non-DST behavior (full 24-hour periods only) using `Math.floor(differenceInHours(dateLeft, dateRight)/24)|0`.
* Adds analogous JSDoc content to `differenceInWeeks` which is implemented using `differenceInDays` and shares its DST behavior.
* Updates the changelog with a new empty release section as requested in CONTRIBUTING.md,

**Fix Details**

#533, like most JS bugs related to DST, happened because the pre-2.12 implementation of `differenceInDays` mixed UTC and local date calculations: `CompareAsc` uses UTC comparisons while `Date.setDate` (used in `differenceInDays`) uses local timezone math. The result: on or around DST boundaries `CompareAsc` returned results which suggested that the difference was not a full day when it really was, or vice versa. This led to off-by-one errors in results like #533.

The fix was to restore the 2.11 implementation of `differenceInDays` (effectively reverting #1630) and simply replace usage of `compareAsc` with a new local function `compareLocalAsc` that uses local date math instead.

If you really want to get into the weeds to understand how `setDate` works and why #533 happened, start here: 
* https://tc39.es/ecma262/#sec-date.prototype.setdate
* https://tc39.es/ecma262/#sec-local-time-zone-adjustment

**Discussion**

_"Full Days": 24-hour periods vs. matching local times_ 

Some contributors in #533 suggested that date diffs should be measured in 24-hour periods only and should ignore DST.  To see why this is problematic, imagine a user measures a 7-day period from (local) noon to (local) noon one week later.  If we only diff 24-hour periods, then that period would measure 7 days most of the year, but would return 6 days if DST started in the middle.  This intermittent variation is a certain cause of hard-to-diagnose bugs, and is also different from how date-fns has behaved for a long time.  That's why this PR goes back to the 2.11 behavior of taking DST into account.  BTW, if a developer wants to ignore DST and just count 24-hour periods, this PR documents one way (`Math.floor(differenceInHours(dateLeft, dateRight)/24)|0`) in user-visible JSDoc comments.

_API Consistency_

I admit that I couldn't fully understand the consistency discussion in #533. To me it seems like the pre-2.12 behavior was *more* consistent, because in 2.11 `differenceInDays` was consistent with `addDays`, and `differenceInDays` was also consistent with `differenceInMonths`.  And with JS Date APIs like `Date.setDate`.

_Guidelines_

In case it's helpful, here are three general rules of thumb that I use for safe date math:
* For adding/subtracting/comparing hours (and smaller) units, convert to UTC and do all math in UTC.
* For adding/subtracting/comparing days (and larger) units, only use local timezone math, not UTC (unless UTC is the local timezone!) because users expect that day and larger APIs return the same results whether or not a DST transition happened.  Here's a discussion about the same topic in moment.js: https://github.com/moment/moment/issues/78#issuecomment-2979637
* Know which APIs use local vs. UTC math, and don't mix them unless you really, really know what you're doing. When you mix UTC APIs like `Date.getDate` with local APIs like `date.setDate`, it's easy to end up with off-by-one-hour bugs like #533 or #1083 1510.

**Future Investigation Needed**

TODO: I suspect that `differenceInMonths`, `differenceInISOWeekYears`, and `differenceInYears` are vulnerable to the same bug but I didn't have the time yet to investigate but I wanted to get this breaking-change fix out ASAP.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,11 +12,12 @@ This change log follows the format documented in [Keep a CHANGELOG].
 
 ## Unreleased
 
-Thanks to [@JorenVos](https://github.com/JorenVos), [@developergouli](https://github.com/developergouli), [@rhlowe](https://github.com/rhlowe) for working on the release!
+Thanks to [@JorenVos](https://github.com/JorenVos), [@developergouli](https://github.com/developergouli), [@rhlowe](https://github.com/rhlowe), [@justingrant](http://github.com/justingrant) for working on the release!
 
 ### Fixed
 
 - [Fixed mei abbreviation in the Dutch locale](https://github.com/date-fns/date-fns/pull/1752).
+- [Fix differenceInDays DST behavior broken in 2.12.0](https://github.com/date-fns/date-fns/pull/1754).
 
 ### Added
 
--- a/src/differenceInCalendarDays/test.js
+++ b/src/differenceInCalendarDays/test.js
@@ -3,6 +3,7 @@
 
 import assert from 'power-assert'
 import differenceInCalendarDays from '.'
+import { getDstTransitions } from '../../test/dst/tzOffsetTransitions'
 
 describe('differenceInCalendarDays', function() {
   it('returns the number of calendar days between the given dates', function() {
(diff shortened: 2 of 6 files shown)


justingrant:
My new tests failed in CI for the Atlantic/Azores timezone. Interestingly, the tests failed for `differenceInCalendarDays` whose implementation didn't change in this PR, so either my tests are bad or there's a DST problem with `differenceInCalendarDays`.  I can repro this failure locally and I'll investigate and push a fix, probably tomorrow.

Kudos to the date-fns team for a great CI setup to catch weird bugs like this!

kossnocorp:
Thank you so much for detailed research, your help is invaluable 👏

justingrant:
OK, fixed the problem with Atlantic/Azores. The culprit was my test code to find DST boundaries, which broke if those boundaries were at midnight. Fixed now.  Also I rebased to the current master and put my changelog edits on top. 

Should be ready to merge.

kossnocorp:
I've shipped the change with `date-fns@2.13.0`, thank you again!

PR #1760, opened 2020-05-07
Fix DST issues with add & addMonths

### LINKED ISSUE #1682: Add gives faulty result shortly after DST transition
Starting out with a Date within an hour after the DST transition from summer time to standard time, when adding a duration using `add` for which the result should be a Date an hour after the DST transition, the result is actually the Date immediately after the transition.

Encountered using v2.11.0 for Dutch DST at 27-10-2019 03:00 GMT+2 to 02:00 GMT+1:

```
// Each variable in the following code is set to 
// 'Sun Oct 27 2019 02:00:00 GMT+0100 (Central European Standard Time)'

let d0 = add(new Date(2019, 9, 27, 2, 0, 0), { hours: 1 });
let d1 = add(d0, { hours: 1 }); 
let d2 = add(add(d0, { minutes: 30 }), { minutes: 30 }); 
```

Using `addHours` does not result in the same behaviour.

### LINKED ISSUE #1758: differenceInMonths reports a full month even if it's one hour short (if period starts at DST)
While implementing #1754, I discovered that `differenceInMonths` is vulnerable to the same problem reported for `differenceInDays` in #533. 

In this `differenceInMonths` issue, a period will be reported as a full month even through the period is really one hour short of a month, as long as the period starts exactly when DST starts. 

I probably won't have time to PR this, but I wanted to record the issue for future folks who run into this problem later, or who want an interesting problem to fix.

**Repro:** 

```
// I ran this in timezone America/Los_Angeles, but will
// repro in any DST timezone-- just change the
// dates to match the DST start time and one month
// after that minus one hour.
const dstStartTime = new Date(2020, 2, 8, 3);
const oneMonthLater = new Date (2020, 3, 8, 2);
const diffMonths = differenceInMonths(oneMonthLater, dstStartTime);
console.log(`Months: ${diffMonths} (expected 0)`);
```

**Notes for Fix**

The fix for `differenceInMonths` is likely similar to #1754: replace use of `CompareAsc` with a version that uses local time instead of UTC.  I assume it makes sense to make this local-time comparer into shared code (and maybe make it a public API?). 

Tests added in #1754 `differenceInDays` and `DifferenceInCalendarDays` can likely be copy/pasted and slightly revised to work for months instead of days. 

Documentation also should be updated like in #1754. 

I suspect that the same problem may affect `differenceInISOWeekYears`, but the problem is being masked by #1757 so may not be worth fixing.

justingrant:
This PR fixes #1682. Changes are fairly straightforward:
* `add` now only calls `addMonths` if `months` or `years` adding is requested, and it only calls `addDays` if `days` or `weeks` adding is requested. This fixes #1682 by preventing calls to `Date.setXXX` in cases where those calls aren't needed, which in turn prevents mutation of the time from DST end to one hour before DST end. See https://github.com/date-fns/date-fns/issues/1682#issuecomment-625062812
* `addMonths` and `addDays` now return early if `amount` of increment is zero. This prevents the same problem as #1682 but when calling `addMonths` or `addDays` directly. 
* Re-worked `addMonths` logic to avoid problems like #1758 which are caused by mixing UTC and local-time APIs.
* Added DST tests to `add`, `addMonths`, and `addDays` to verify the fixes above.
* Updated changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,18 @@ This change log follows the format documented in [Keep a CHANGELOG].
 [semantic versioning]: http://semver.org/
 [keep a changelog]: http://keepachangelog.com/
 
+## Unreleased
+
+Thanks to [@justingrant](http://github.com/justingrant) for working on the release!
+
+### Fixed
+
+- [Fix DST issues with add & addMonths](https://github.com/date-fns/date-fns/pull/1760).
+
+### Added
+
+-
+
 ## [2.13.0] - 2020-05-06
 
 Thanks to [@JorenVos](https://github.com/JorenVos), [@developergouli](https://github.com/developergouli), [@rhlowe](https://github.com/rhlowe) and [@justingrant](http://github.com/justingrant) for working on the release!
--- a/src/add/index.js
+++ b/src/add/index.js
@@ -57,10 +57,13 @@ export default function add(dirtyDate, duration) {
   const seconds = 'seconds' in duration ? toInteger(duration.seconds) : 0
 
   // Add years and months
-  const dateWithMonths = addMonths(toDate(dirtyDate), months + years * 12)
+  const date = toDate(dirtyDate)
+  const dateWithMonths =
+    months || years ? addMonths(date, months + years * 12) : date
 
   // Add weeks and days
-  const dateWithDays = addDays(dateWithMonths, days + weeks * 7)
+  const dateWithDays =
+    days || weeks ? addDays(dateWithMonths, days + weeks * 7) : dateWithMonths
 
   // Add days, hours, minutes and seconds
   const minutesToAdd = minutes + hours * 60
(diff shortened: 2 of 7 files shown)


justingrant:
Whoops, I accidentally linked this PR to #1758. Even though I edited the description to remove the link, GitHub still shows #1758 as linked which will close it if this PR is merged. 

@kossnocorp, are you able to remove this link?  The links section shows as read-only to me.

justingrant:
Also, I forgot to make the parallel change to `addDays` to avoid mutating the date when a zero-day increment is requested. Making that change now.

justingrant:
OK all changes are complete and all tests pass in CI. Ready to merge if code review looks good.

swrobel:
@justingrant any idea if this issue affected `addBusinessDays` & if so, if this fixed it or if it would require a separate fix?

zuzusik:
Similar fix is needed for subtraction functionality - `sub` method still doesn't consider DST shifts

PR #1765, opened 2020-05-10
Fix formatDistance strings in Hungarian locale

jmate0321:
All of the modified lines are containing suffixes (múlva), which breaks localized duration strings.


kossnocorp:
Please run `yarn locale-snapshots` and commit the results

jmate0321:
Committed, thanks! Looking at the generated snapshot made me realize that there were other issues with the Hungarian translation, fixed them as well.

jmate0321:
Thanks!

PR #1775, opened 2020-05-15
Add a missing preposition to the Spanish locale.

### LINKED ISSUE #1770: The Spanish locale is missing a preposition.
First of all, sorry if this has already been discussed elsewhere, I couldn't find anything about it and decided to go ahead and open an issue.

For the Spanish locale, both the `full` and `long` `dateFormats` are missing a `'de' ` before the year (`y`).

If the change is just adding the missing preposition [here](https://github.com/date-fns/date-fns/blob/master/src/locale/es/_lib/formatLong/index.js#L4-L5) I could submit a PR with it!

AlbertoPdRF:
Fixes #1770.


kossnocorp:
Thank you!

PR #1777, opened 2020-05-16
improve it formatRelative (gender + this/last/nextWeek handling)

giogonzo:
code is ["translated" from ru](https://github.com/date-fns/date-fns/blob/master/src/locale/ru/_lib/formatRelative/index.js) (example given at https://date-fns.org/v2.0.0-alpha.24/docs/I18n-Contribution-Guide#formatrelative)


giogonzo:
@vin-car @kossnocorp maybe you can have a look at this?

vin-car:
Yes, that's ok, all italian weekday names are masculine except Sunday (domenica) that is feminine. So it's correct to have "domenica scorsa" (last Sunday) and "lunedì scorso" (last Monday).

kossnocorp:
@vin-car thank you!

giogonzo:
Other than the change highlighted by @vin-car , my PR also introduced an additional change: only using "prossimo/scorso" ("next/past") if the relative day is far enough in time, not part of the current week.

I actually investigated a bit better and found out that the "Accademia della Crusca" (sort of a research institute for the Italian language) is [suggesting differently](https://accademiadellacrusca.it/it/consulenza/prossimo-con-unit%C3%A0-temporali/132): that "prossimo/scorso" ("next/past") should be always used, no matter if the day is in the current week or not.

I personally never heard someone using e.g. "giovedì prossimo" ("next thursday") to refer to the day after if today is "mercoledì" ("wednesday") 🤔 
But if you think it is better, I will undo that part of the changes and only leave the smaller gender change. What do you think?

Issue #1628, opened 2020-02-03
addBusinessDays / subBusinessDays is wrong


KidkArolis:
```diff
const { format, addBusinessDays } = require('date-fns')

format(addBusinessDays(new Date('2019-08-18'), 0), 'yyyy-MM-dd EEEE')
'2019-08-18 Sunday'

format(addBusinessDays(new Date('2019-08-18'), -1), 'yyyy-MM-dd EEEE')
'2019-08-16 Friday'

format(addBusinessDays(new Date('2019-08-18'), -2), 'yyyy-MM-dd EEEE')
'2019-08-15 Thursday'

format(addBusinessDays(new Date('2019-08-18'), -3), 'yyyy-MM-dd EEEE')
'2019-08-14 Wednesday'

format(addBusinessDays(new Date('2019-08-18'), -4), 'yyyy-MM-dd EEEE')
'2019-08-13 Tuesday'

format(addBusinessDays(new Date('2019-08-18'), -5), 'yyyy-MM-dd EEEE')
- '2019-08-11 Sunday' (WRONG)
+ '2019-08-12 Monday'

format(addBusinessDays(new Date('2019-08-18'), -6), 'yyyy-MM-dd EEEE')
'2019-08-09 Friday'

format(addBusinessDays(new Date('2019-08-18'), -7), 'yyyy-MM-dd EEEE')
'2019-08-08 Thursday'

format(addBusinessDays(new Date('2019-08-18'), -8), 'yyyy-MM-dd EEEE')
'2019-08-07 Wednesday'
```

Matsuuu:
I could try out fixing this

PR #1790, opened 2020-05-20
Fix addBusinessDays when input date is a weekend day (#1628)

Matsuuu:
The functionality reduced a full week when the the amount of reducing was 5 or more. If the user gave a weekend date, this would land the user back on a weekend. 

The fix introduced will fix this

### DIFF
--- a/src/addBusinessDays/test.js
+++ b/src/addBusinessDays/test.js
@@ -81,4 +81,11 @@ describe('addBusinessDays', function() {
     assert.throws(addBusinessDays.bind(null), TypeError)
     assert.throws(addBusinessDays.bind(null, 1), TypeError)
   })
+  it('starting from a weekend day should land on a weekday when reducing a divisible by 5', function() {
+    const substractResult = addBusinessDays(new Date(2019, 7, 18), -5)
+    assert.deepEqual(substractResult, new Date(2019, 7, 12))
+
+    const addResult = addBusinessDays(new Date(2019, 7, 18), 5)
+    assert.deepEqual(addResult, new Date(2019, 7, 23))
+  })
 })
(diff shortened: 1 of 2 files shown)


Matsuuu:
The tests seem to fail on Travis, but when I run `yarn test` locally, they succeed.

src/addBusinessDays/test.js
```diff
@@ -81,4 +81,11 @@ describe('addBusinessDays', function() {
     assert.throws(addBusinessDays.bind(null), TypeError)
     assert.throws(addBusinessDays.bind(null, 1), TypeError)
   })
+  it('starting from a weekend day should land on a weekday when reducing a divisible by 5', function() {
+    const substractResult = addBusinessDays(new Date('2019-08-18'), -5)
+    assert.deepEqual(substractResult, new Date('2019-08-12T00:00:00.000Z'))
+
+    const addResult = addBusinessDays(new Date('2019-08-18'), 5)
+    assert.deepEqual(addResult, new Date('2019-08-23T00:00:00.000Z'))
```
kossnocorp, line 89: You have to create dates using numbers instead of strings because with strings you get different local time.

kossnocorp: I.e `new Date('2019-08-18')` → `new Date(2019, 7, 18)`

Matsuuu: Gotcha. Rewrote the tests now

Matsuuu: Also, thanks ! :)

Matsuuu:
Allright now they seem to pass nicely! Thanks for the tip

Matsuuu:
Hi @kossnocorp . This could be merged now

Issue #1748, opened 2020-04-28
parseISO - parsing invalid string into a valid date


MBelniak:
Hi.
I've come across this issue while using date-fns. 
Current behaviour:
`parseISO('2016-03-16T08 ooo') // returns Date(Wed Mar 16 2016 08:00:00 GMT+0100 (CET))`
Expected behaviour (at least momentjs and some online regex-based parsers):
`parseISO('2016-03-16T08 ooo') // returns Invalid Date`
Is this actually a bug or for some reason current bahaviour is the expected one?

Imballinst:
@kossnocorp is this intended behavior? From a quick glance on `parseISO`, it uses RegEx to parse the date and time. I think in this case, it only "partially parses" the invalid string. CMIIW!

kossnocorp:
I've no strong opinion regarding this, to be honest. As I see, not only Moment.js behaves like that:

```js
new Date('2016-03-16T08 ooo')
//=> Invalid Date
Date.parse('2016-03-16T08 ooo')
//=> NaN
```

So I guess this is a bug.

Imballinst:
Alright, thanks for the clarification! I'll try attempting on this issue.

kossnocorp:
@Imballinst thanks 🙏

Imballinst:
@dojchek noted -- thanks for the information! I'll try handling that case as well in my PR.

Imballinst:
@dojchek it seems like the 2 digits are used to represent centuries, so I guess that's intended?

https://github.com/date-fns/date-fns/blob/4cf918f6e66499ced4254decf3573bf365e871c8/src/parseISO/test.js#L103-L120

CMIIW @kossnocorp

dojchek:
@Imballinst Yes, I also think that's intended - although not very intuitive..
I found only one reference to this format in wiki article which says:
`One may simply write "1981" to refer to that year or "19" to refer to the century from 1900 to 1999 inclusive`

So it is in fact a valid ISO date. I still need this to "fail" in my use case - so I will check it with some regex or simply check the length of string before doing parseISO, since I only work with dates like `2020-06-26` and `2020-06-26T04:33:26+00:00`.

Thank you very much for checking this out

PR #1791, opened 2020-05-20
fix parseISO not returning Invalid Date when there are space(s) in passed string

Imballinst:
Fixes https://github.com/date-fns/date-fns/issues/1748. This PR adds additional validation when a datetime string is split using the existing RegEx in `splitDateString`, hence `parseISO` will return "Invalid Date" for strings with space(s).

## Current

```js
// Since we only check first 2 array elements, it won't resolve to "Invalid Date".
splitDateString('2014-02-11T07:00:00  basketball'); // ['2014-02-11', '07:00:00', 'basketball']
parseISO('2014-02-1107:00:00  basketball'); // 2014-02-11T00:00:00.000Z (I'm in GMT+7)

// This is also the same for date-only string.
splitDateString('2014-02-11  basketball'); // ['2014-02-11', '', 'basketball']
parseISO('2014-02-11  basketball'); // 2014-02-10T17:00:00.000Z (if time is not specified, we use 12am)
```

## This PR

With that, the solution of this PR is to always check the resulting array length from `splitDateString`. If the array length is more than 2, then it contains space(s) and it should resolve to "Invalid Date".

This is also considering the fact that ISO8601-formatted dates don't have spaces in it (source: https://www.w3.org/TR/NOTE-datetime).

```js
splitDateString('2014-02-11T07:00:00  basketball'); // ['2014-02-11', '07:00:00', 'basketball']
parseISO('2014-02-11T07:00:00  basketball'); // Invalid Date

splitDateString('2014-02-11  basketball'); // ['2014-02-11', '', 'basketball']
parseISO('2014-02-11  basketball'); // Invalid Date
```

### DIFF
--- a/src/parseISO/index.js
+++ b/src/parseISO/index.js
@@ -155,7 +155,7 @@ function splitDateString(dateString) {
   var timeString
 
   // The regex match should only return at maximum two array.
-  // [date], [time], or [date, time].
+  // [date] or [date, time].
   if (array.length > 2) {
     return dateStrings
   }
--- a/src/parseISO/index.js
+++ b/src/parseISO/index.js
@@ -155,7 +155,7 @@ function splitDateString(dateString) {
   var timeString
 
   // The regex match should only return at maximum two array.
-  // [date] or [date, time].
+  // [date], [time], or [date, time].
   if (array.length > 2) {
     return dateStrings
   }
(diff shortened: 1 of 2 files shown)


src/parseISO/index.js
```diff
@@ -154,6 +154,12 @@ function splitDateString(dateString) {
   var array = dateString.split(patterns.dateTimeDelimiter)
   var timeString
 
+  // The regex match should only return at maximum two array.
```
Imballinst, line 157:
```suggestion
  // The regex match should only return at maximum two array elements.
```

PR #1792, opened 2020-05-20
Fix es round tripping dates with Wednesday

arsnyder16:


src/locale/es/snapshot.md
```diff
@@ -182,7 +182,7 @@
 |                                 |              | 1453-05-29T23:59:59.999Z | 29 may 1453, 23:59:59                                     | 1453-05-29T23:59:59.000Z |
 |                                 | PPPppp       | 1987-02-11T12:13:14.015Z | 11 de febrero de 1987 a las 12:13:14 GMT+0                | Errored                  |
 |                                 |              | 1453-05-29T23:59:59.999Z | 29 de mayo de 1453 a las 23:59:59 GMT+0                   | Errored                  |
-|                                 | PPPPpppp     | 1987-02-11T12:13:14.015Z | miércoles, 11 de febrero de 1987 a las 12:13:14 GMT+00:00 | Invalid Date             |
+|                                 | PPPPpppp     | 1987-02-11T12:13:14.015Z | miércoles, 11 de febrero de 1987 a las 12:13:14 GMT+00:00 | Errored                  |
```
kossnocorp, line 185: Please fix the exception

arsnyder16: @kossnocorp This exception is unrelated and is a fundamental problem with the parser. My change just allowed the parser to get further. If you look through the snapshots for all locales you will see that anytime pppp, ppp always results in an exception. Including en-US

https://github.com/date-fns/date-fns/blob/master/src/locale/en-US/snapshot.md

kossnocorp: Ok, thanks!

PR #1795, opened 2020-05-20
fix roundtrip bug with d/EEEE ordering in formats like PPPPP

arsnyder16:
Anytime two tokens exist in a format string that have the same priority one needs to take precedent over the other when parsing otherwise incorrect values will be parsed. 

The example that is shown in the snapshots that are now fixed are cases where  E and d coexist such as PPPP. Before some locales would handle it find because E would come before d, but in locales where d came before E there were issues. So adding a sub priority so we can correctly sort which token takes precedent.

en-US : PPPP `'EEEE, MMMM do, y`
ja : PPPP `y年M月d日EEEE`


kossnocorp:
I'll appreciate it if you can provide an explanation for your changes. Thanks!

arsnyder16:
@kossnocorp Updated the first comment to include a description

PR #1825, opened 2020-06-04
use full year in fr-CA short date to comply with other locale's short date format

jpmaheux:
was using 'yy'  instead of 'y' or 'yyyy'

#1826


jpmaheux:
@gpetrioli as you initially added it, maybe you want to have a look at it

jpmaheux:
after some research, based on icu its normal that frCA only have 2 digits year so I will reject this issue, sorry
http://www.localeplanet.com/icu/fr-CA/index.html

MickMonaghan:
> after some research, based on icu its normal that frCA only have 2 digits year so I will reject this issue, sorry
> http://www.localeplanet.com/icu/fr-CA/index.html

Hi @jpmaheux ,
Can you reopen / merge this PR please?

The LocalePlanet.com site looks quite out of date. Home page says that there's "localization in JavaScript: there is none". Once upon a time - that was true - but not now.

Looking at the current version of ICU (v69.1), here's what I get for the fr-CA locale:
```
Date d1 = new Date(); 
System.out.println(DateFormat.getDateInstance(DateFormat.SHORT, new ULocale("fr-CA")).format(d1));

// 2021-05-13
```
So the 4 digit year looks correct to me.

Issue #1833, opened 2020-06-08
IE11 error in v2.14.0


jackocnr:
I'm seeing this error:

> Expected identifier, string, or number

When I upgrade from v2.13.0 to v2.14.0.

I'm using the following imports from date-fns:

```js
import format from 'date-fns/format';
import parseISO from 'date-fns/parseISO';
import { da, de, es, fr, id, it, ms, nb, nl, pl, pt, ro, sv, tr } from 'date-fns/locale';
```

Here is a screenshot from IE11 devtools - the error is caused by line 10481.

<img width="592" alt="Screenshot 2020-06-08 at 15 00 39" src="https://user-images.githubusercontent.com/1186883/84039238-e0cee500-a998-11ea-8d20-9043959ee74f.png">

Let me know if there's anything else I can provide.

jeroenkampinga:
I have te same error in ie11.

droyuki:
you may need @babel/plugin-transform-computed-properties

[key]: value is not supported by IE 11

gottsohn:
@jackocnr @jeroenkampinga 
I had the same error, here's a temporal solution I used until this is fixed.

Using npm `postinstall` script. I rewrite the [-1] to '-1'

```sh
#!/usr/bin/env sh

replaceLocale()
{
    if test -f "$1" ; then
        echo "Replacing [-1] in '$1'."
        sed -i -e "s%\[-1\]\:%'-1'\:%" $1
        sed -i -e "s%\[1\]\:%'1'\:%" $1
        sed -i -e "s%\[0\]\:%'0'\:%" $1
        echo "Replaced IE 11 incompatible syntax in '$1'"
    else
        echo "'$1' does not exist."
    fi 
}

replaceLocale "node_modules/date-fns/esm/locale/hu/_lib/formatDistance/index.js"
replaceLocale "node_modules/date-fns/locale/hu/_lib/formatDistance/index.js"


```

kidkkr:
You can transpile `date-fns` with webpack `babel-loader`.

kossnocorp:
Closed by https://github.com/date-fns/date-fns/pull/1842

PR #1842, opened 2020-06-12
fix(HungarianLocale): Use IE 11 friendly syntax for object keys

gottsohn:
Fix for #1833 

IE 11 doesn't support object keys as variables e.g `{....[-1]: 1...}`.
The keys must be string or number literals.


gottsohn:
> LGTM



> LGTM.
> 
> Also, what is the point of using numbers between brackets for key assignment? I thought that was for with variables used as keys `a = "b" c = {[a]: "d"}`

Apparently, the numbers are mapped to the `key` argument in the `translate` method when _addSuffix_ is `true`.

PR #1848, opened 2020-06-19
Removed periods for dutch month abbreviations

dakkafex:
In the dutch language, abbreviating a month does not require to put a period ``.`` at the end, it's merely optional (and old fashioned).
Adding a period is easier than removing it from the format result.

https://onzetaal.nl/taaladvies/afkortingen-dagen-en-maanden/

### DIFF
--- a/src/locale/nl/_lib/localize/index.js
+++ b/src/locale/nl/_lib/localize/index.js
@@ -15,18 +15,18 @@ var quarterValues = {
 var monthValues = {
   narrow: ['J', 'F', 'M', 'A', 'M', 'J', 'J', 'A', 'S', 'O', 'N', 'D'],
   abbreviated: [
-    'jan.',
-    'feb.',
-    'mrt.',
-    'apr.',
+    'jan',
+    'feb',
+    'mrt',
+    'apr',
     'mei',
-    'jun.',
-    'jul.',
-    'aug.',
-    'sep.',
-    'okt.',
-    'nov.',
-    'dec.'
+    'jun',
+    'jul',
+    'aug',
+    'sep',
+    'okt',
+    'nov',
+    'dec'
   ],
   wide: [
     'januari',


curry684:
I'm 👎 on this change. The basic rule in Dutch is that abbreviations are indicated with the trailing dot, and it should always be used as such to avoid confusion, an example that `date-fns` should also handle correctly is:
> Jan begint school weer. (John starts school again)
Which is an entirely different sentence than:
> Jan. begint school weer. (School starts again in January)

https://www.taaltelefoon.be/maanden-afkorting is more explicit on why and how it is allowed to leave the dot out, translated:
> In certain contexts, like in agendas or on calendars, the dot is often left out. This is also correct.

CLDR actually [support this notion via standalone vs. formatted use](http://cldr.unicode.org/translation/date-time-1/date-time-patterns#TOC-When-to-use-Standalone-vs.-Formatting), but [uses the dot in both](https://unicode-org.github.io/cldr-staging/charts/37/summary/nl.html#360c210ad4f23085). One could argue that standalone should drop the dot, but that would imho have to start with the CLDR, and this library could then follow.

curry684:
https://taaladvies.net/taal/advies/vraag/1727 is also explicit btw for the same reasons as the Belgian site:
> Jan/jan., feb/feb. enzovoort zijn 'echte' afkortingen. Dat wil zeggen dat jan niet wordt uitgesproken als 'jan', maar als 'januari', dus als het volledige woord waar de afkorting voor staat. Echte afkortingen worden doorgaans met een punt geschreven, bijvoorbeeld: afb. (afbeelding), sgt. (sergeant), zgn. (zogenaamd). Dat leidt tot jan., feb. of febr., mrt., apr., jun., jul., aug., sep. of sept., okt., nov. en dec. – met punt dus.
>
>De schrijfwijze zonder punt is echter ook juist. In lijsten, tabellen, agenda's en dergelijke wordt de punt meestal weggelaten: jan, feb, enzovoort.

This also boils down to: correct is with a dot, without is commonly accepted and also correct, but mostly used like that in tables, lists and agendas and the like.

stefanvermaas:
I agree with @curry684 on this. I'm not a fan of this change either.

kossnocorp:
@curry684 @stefanvermaas thank you so much for the input!

@dakkafex if you badly need it, I recommend making a custom locale that depends on `nl` and using it.

PR #1862, opened 2020-06-30
Fix Swedish (sv) locale

eckberg:
Typos and style fixes for Swedish (sv) locale:

- Incorrect pluralization for ”vecka” (should be ”veckor”).
- Abbreviations of month names to follow recommendation from Språkrådet (Svenska skrivregler, 2017).
- Abbreviations of day names to follow same convention as for month names.
- Abbreviations of am/pm (_förmiddag_/_eftermiddag_) now follow recommendation from Språkrådet (Svenska skrivregler, 2017) – i.e. having a period at the end of ”fm.” and ”em.” and not e.g. ”f.m.” or ”fm”.
- Abbreviations for morning and afternoon removed (i.e. fully spelled out instead) in alignment with en-US locale. (Also better aligned with abbreviation best practices in Swedish.)


kossnocorp:
@date-fns/i18n-sv what do you think about the change? Appreciate your input!

alexandernanberg (commented):
Changes looks good to me

Regarding Språkrådet vs CLDR, I'm not sure which should be seen as the source of truth. AFAIK both versions are correct. But might be easier to use CLDR just because there is an online reference (couldn't find any for Språkrådet because you have buy the book)

tan75 (approved):
Changes approved by 
@alexandernanberg

kossnocorp:
After consideration, I decided to close the PR. @eckberg, I realize that it might not be the conclusion you expected. After all, your changes make sense. However, we need a single source of truth to function efficiently, and the one that is open and accessible is CLDR.

If you're passionate about it, I would love to connect with you and help you make a separate package with Swedish locale following Språkrådet and host it https://github.com/date-fns.

eckberg:
> After consideration, I decided to close the PR. @eckberg, I realize that it might not be the conclusion you expected. After all, your changes make sense. However, we need a single source of truth to function efficiently, and the one that is open and accessible is CLDR.
> 
> If you're passionate about it, I would love to connect with you and help you make a separate package with Swedish locale following Språkrådet and host it https://github.com/date-fns.

No problem – seems fair! I might reach out late when I got a bit more time to dig into this.

kossnocorp:
@eckberg awesome, we want eventually publish locales as separate packages, so Språkrådet could be accessed via `@date-fns/sv/sprakradet`.

Issue #816, opened 2018-07-24
Add function to set locale globally


moimael:
Hi,

It would be great to have the possibility to call a function that set the locale globally, instead of having to wrap date-fns functions. Something similar to `moment.locale()` where we would pass the imported locale.

Thanks !

moimael:
Is it a functionality that the core team would agree to have ?

moimael:
We are still blocked from migrating from moment to date-fns because of this, is it planned in upcoming releases ?

TheAifam5:
I would be happier with a global configuration for date-fns :) Not only locale but default formats, and everything which is set currently "by-default".

nghiepit:
+1

Brunohes:
+1

olee:
I think it would be great to add it kinda like this:
```js
import { formatDistance } from 'date-fns'
// Require Esperanto locale
import { setDefaultLocale } from 'date-fns/locale'
import { eo } from 'date-fns/locale'
setDefaultLocale(eo);
```

This would still preserve the benefit of only importing used locales when bundling but also allows setting a default locale.
This is especially important when using date-fns in other libraries, because each one would have to do their own implementation of a default locale right now and a project using such libraries would have to set the default locale many times then.

DaniilVysotskiy:
I may bump this issue once again.

MartinDevillers:
+1 I know it's not very "pure" to do a global init somewhere in your code, but for the purposes of localization it's just really handy to set the user's locale once and then not having to worry about it anywhere else in your code. Without this feature, I have to wrap all locale aware methods from this library (or wire in the locale manually each time I use such a method-- ugh).

olee:
Same reason here.
Even though you could argue like "this could break when used in libraries" etc etc, there are soo many cases where this makes so much sense like even just when you want to use it in a small nodeJs application or whatever. 
Also having to wrap it each time will just create a mess when you write libraries using date-fns, because every library will have to use it's own wrapping code which is just useless boilerplate and makes everything quite cumbersome.

Also you could argue: Why can English localization be the default, but not other locales?

joaovbibiano:
+1

billyjov:
+1

xsigmar:
+1

syrepol:
+1

olee:
@leshakoss @kossnocorp would you mind looking into this please?
Right now this issue is a huge blocker for many users.
[...]

PR #1885, opened 2020-07-28
add setGlobalLocale

SneakyFish5:
Closes https://github.com/date-fns/date-fns/issues/816

When trying to run the build script, I ran into this error:
```
  const returns = getType(content.returns[0].type.names)
                                         ^

TypeError: Cannot read property '0' of undefined
```

Please let me know if any changes are required

### DIFF
--- a/src/_lib/getGlobalLocale/index.js
+++ b/src/_lib/getGlobalLocale/index.js
@@ -1,0 +1,5 @@
+export default function getGlobalLocale() {
+  var globalScope = typeof global === undefined ? window : global
+
+  return globalScope.globalLocale
+}
--- a/src/_lib/getGlobalLocale/test.js
+++ b/src/_lib/getGlobalLocale/test.js
@@ -1,0 +1,26 @@
+// @flow
+/* eslint-env mocha */
+
+import assert from 'power-assert'
+import getGlobalLocale from '.'
+import setGlobalLocale from '../../setGlobalLocale'
+
+describe('getGlobalLocale', function() {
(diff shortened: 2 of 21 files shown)


kossnocorp:
Hey! Thank you a lot for the PR! I'm going to evaluate the PR on the week and help with the build issue.

amoerie:
We are also interested in this change. To give some more context: react-datepicker already implements its own form of "default locale", see https://github.com/Hacker0x01/react-datepicker#localization

I would also suggest a few changes to this PR:

- Rename setGlobalLocale to setDefaultLocale. This makes it explicit that passing a locale in the options overrides the default one. 
- Don't pollute the window / global scope, but use a function closure to save the current default locale. There must be some initialization function where this can be stored.

liegeandlief:
Is this still active? Global locale support would be really useful!

hornta:
I want this feature as well. @SneakyFish5 Have you had time to check up on this more or do you want to hand over it to someone?

SneakyFish5:
Hey, sorry for the late reply. Yea, I don't think I can keep up with this PR anymore, I think I'll go ahead and close it and if someone wants to retake it up, be my guest.

Nuboskill:
@SneakyFish5 Why have you closed this PR? It is very important feature.

olee:
Same thought here - this would be such a simple improvement, but would give a huge benefit.
This is the only thing which really annoys me with date-fns.

jdriesen:
Same here...

PR #1889, opened 2020-07-31
link to changelog

TheJaredWilcurt:
the changelog has more details than the upgrade guide on how to upgrade and would have been more useful if I could have found it sooner.


kossnocorp (approved):
Thank you a lot!

PR #1890, opened 2020-08-01
Fix formatRFC3339: adjust formatting offset for timezones with minutes

piotrl:
…with minutes (e.g. India, Australia).

First of all, thanks for making such a modular library! Once we found the issue, we could fix it in minutes in production by forking single-function :)

* Affected time-zones: 
  * Australia, 
  * East of Canada, 
  * India and Sri Lanka, 
  * Islands near by New Zealand
  * I think India is the most populated affected timezone
* Caused formatting offset in expression like `+5.5:30` instead `+05:30`) which makes invalid date in `parseISO()`
* It was simple mistake in division (`-330/60` caused `5.5` hours offset instead full hour)
* I used `toInteger` as it already have condition for positive and negative offset, using `Math.floot()` or `Math.ceil()`. I see in some other functions
* Add specs for interesting and regular timezones for this use case

### DIFF
--- a/scripts/test/formatRFC3339.sh
+++ b/scripts/test/formatRFC3339.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+# The script runs formatRFC3339 in a non%60 timezone offset
+#
+# It's a part of the test process.
+
+set -ex
+
+export PATH="$(yarn bin):$PATH"
+export NODE_ENV=test
+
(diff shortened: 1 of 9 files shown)


kossnocorp (approved):
Thank you so much for such a considerate work, I appreciate it! And sorry about waiting, but you did good forking it locally, that's the way!

Issue #1888, opened 2020-07-31
dead link in docs


TheJaredWilcurt:
On the [Upgrade guide](https://date-fns.org/v2.15.0/docs/Upgrade-Guide)

It points to unicode tokens:

* https://date-fns.org/v2.15.0/docs/unicodeTokens.md

which is a page that doesn't exist. it should instead point to

* https://date-fns.org/docs/Unicode-Tokens

danmana:
Can be closed. Fixed in #1891

PR #1891, opened 2020-08-05
Fix a dead link on Upgrade Guide

SidKH:
Addressing #1888


kossnocorp (approved):
Good catch, thank you!

PR #1892, opened 2020-08-06
add locale to formatDuration typings

katasup:
Hi! I've just found that `locale` property missed in `formatDuration` typings. So, there is fix.

I would like to see it in next build, currently I have to use `// @ts-ignore` to allow use locale with this function. 

Thanks.


katasup:
Oh, sorry. Didn't seen that PR #1881 is already exist.

PR #1893, opened 2020-08-06
he locale: use shorter Hebrew alternative for "about"

yotamofek:
This PR changes all instances of the Hebrew word "בערך" to the prefix "כ" which has the exact same meaning, is much shorter, and actually sounds a little better to native Hebrew speakers (I polled around with my friends and family, obviously this is a subjective statement but I think it's not a harmful assumption).

I checked on the [unicode charts](https://unicode-org.github.io/cldr-staging/charts/37/summary/he.html) and there is no translation of the word "about" anywhere,
and here is an article from the Hebrew Academy (which in Israel holds the mandate for regulating the Hebrew language) which states that bove "בערך" and "כ-" can be used interchangeably: https://hebrew-academy.org.il/2020/05/27/%D7%91%D7%A2%D7%A8%D7%9A-%D7%9B%D6%BE/


kossnocorp (approved):
I can't read it, but it sounds good to me! Thank you a lot for your contribution and patient waiting. I Will ship it soon!

yotamofek:
Thank you, for this amazing library!

Issue #1899, opened 2020-08-17
Luxembourgish language support


dwaxweiler:
I want to add Luxembourgish language support as I have been translating a project that uses your library.

kossnocorp:
Great, looking forward!

PR #1900, opened 2020-08-17
Add Luxembourgish (lb) locale

dwaxweiler:
I have added Luxembourgish language support as announced in #1899.


kossnocorp (approved):
Thank you so much! It looks good to me. I Will ship soon!

PR #1905, opened 2020-08-18
Correct lightFormat usage example

nfantone:
Stumbled upon this very small mistake when checking out the docs and decided to contribute.


kossnocorp (approved):
Good catch, appreciate it!

PR #1910, opened 2020-08-19
Clarifies that strings are still allowed as arguments, but not when a date is expected

bradevans:
After upgrading to v2 I kept seeing this warning. I just felt that it was somewhat misleading and could be clarified.


kossnocorp (approved):
Sounds good to me, thanks!

Issue #1913, opened 2020-08-23
Scottish Gaelic locale issues


leedriscoll:
Since we've started using `date-fns` for Scottish Gaelic dates on the BBC News website in production. We've found a number issues that we need fixed.

**1. 'o chionn' suffix to should actually be a prefix** 
11 diogan o chionn -> o chionn 11 diogan
1 diog o chionn -> o chionn 1 diog

**2. We should be using the modern version of days / day**
7 làithean o chionn -> o chionn 7 là
1 latha o chionn -> o chionn 1 là

**3. Incorrect Seconds/Minutes/Hour for the 1, 2,20:**
_Seconds_
20 diogan  -> o chionn 20 diog
2 diogan  -> o chionn 2 dhiog

_Minutes_
20 mionaidean o chionn -> o chionn 20 mionaid
2 mionaidean o chionn -> o chionn 2 mhionaid

_Hours_
20 uairean de thìde o chionn -> o chionn 20 uair de thìde
2 uairean de thìde o chionn -> o chionn 2 uair de thìde
1 uair o chionn -> o chionn 1 uair de thìde

**4. Ordinal Indicators**
1st – We've checked the latest guidance from the Gaelic Orthographic Conventions group and they have dropped the ‘c’ from     
31st – works in the same way as 11th, so that should be be ‘31mh’
3rd – bundled it in with ‘mh’ so it now ‘3mh’, ‘13mh’, ‘23mh’.
12th – needs to be ‘12na’

PR #1914, opened 2020-08-23
Fixes for Scottish Gaelic locale issues

leedriscoll:
Resolves https://github.com/date-fns/date-fns/issues/1913


kossnocorp (approved):
Thank you a lot 🙏

PR #1920, opened 2020-08-26
Updating readme by making it a bit easier to read

SaschaKrause:
Also adding a jobs section pointing to job from jobs.date-fns.org

image


kossnocorp (approved):
Works for me, I still have few nitpicks, but I'll fix that in `master`.

PR #1925, opened 2020-08-28
Space should be after 'o chionn' not before 

leedriscoll:
I'm really sorry about this but I introduced a small bug in https://github.com/date-fns/date-fns/pull/1914. 
Should have really noticed this in the snapshots really 🙈 

'o chionn' was moved to be a prefix not a suffix but when doing this I left the space in the wrong place.

I had the space at the start of the string when it should have been at the end.


kossnocorp (approved):
Thank you! Will release it shortly.

kossnocorp:
@leedriscoll shipped as `2.16.1`.

leedriscoll:
> @leedriscoll shipped as `2.16.1`.

Thank you 👍

PR #1928, opened 2020-08-30
Fix sr-Latn typos

aleksaps:
Monday and Sunday in Serbian language do not have extra **j** after **d** like other similar languages from Balkan.
Removed those.


kossnocorp (approved):
Thank you, will ship it soon with a bug fix release!

PR #1930, opened 2020-08-31
Fixed greek grammar for Saturday on formatRelative

BanForFun:


src/locale/el/_lib/formatLong/index.js
```diff
@@ -4,7 +4,7 @@ var dateFormats = {
   full: 'EEEE, d MMMM y',
   long: 'd MMMM y',
   medium: 'd MMM y',
-  short: 'd/M/yy'
+  short: 'd/M/y'
```
kossnocorp, line 7: I'm pretty sure this change is not correct. `y` does not provide shorter result:

```
> dateFns.format(new Date(), 'd/M/y')
'31/8/2020'
> dateFns.format(new Date(), 'd/M/yy')
'31/8/20'
```

I know it's confusing, but it's needed for dates like 5 AD:

```
> const date = new Date()
undefined
> date.setFullYear(5)
-61988407055185
> dateFns.format(date, 'd/M/y')
'31/8/5'
> dateFns.format(date, 'd/M/yy')
'31/8/05'
```

kossnocorp (changes requested):
I appreciate the change in `formatRelative` but you did unrelated change and also forgot to commit the snapshot (run `yarn locale-snapshots` and commit the result)

src/locale/el/snapshot.md
```diff
@@ -176,8 +176,8 @@
 |                                 |              | 1453-05-29T23:59:59.999Z | 11:59:59 μ.μ. GMT+0                                    | Errored                  |
 |                                 | pppp         | 1987-02-11T12:13:14.015Z | 12:13:14 μ.μ. GMT+00:00                                | Errored                  |
 |                                 |              | 1453-05-29T23:59:59.999Z | 11:59:59 μ.μ. GMT+00:00                                | Errored                  |
-| Combination of date and time    | Pp           | 1987-02-11T12:13:14.015Z | 11/2/87, 12:13 μ.μ.                                    | 1987-02-11T12:13:00.000Z |
-|                                 |              | 1453-05-29T23:59:59.999Z | 29/5/53, 11:59 μ.μ.                                    | 1453-05-29T23:59:00.000Z |
+| Combination of date and time    | Pp           | 1987-02-11T12:13:14.015Z | 11/2/1987, 12:13 μ.μ.                                  | 1987-02-11T12:13:00.000Z |
+|                                 |              | 1453-05-29T23:59:59.999Z | 29/5/1453, 11:59 μ.μ.                                  | 1453-05-29T23:59:00.000Z |
```
kossnocorp, line 180: You can see the unexpected result of changing the short locale.

src/locale/el/_lib/formatLong/index.js
BanForFun: Sorry, about that. I just committed the locale snapshot. The change in formatRelative was intentional, but I forgot to mention it. I didn't want for the year component to become shorter, I just think that the full year format is more readable.

kossnocorp: Please rollback the change and use a custom format in your code. `d/M/yy` corresponds to the standard: https://unicode-org.github.io/cldr-staging/charts/37/summary/el.html#57dac0d1b36c1261

kossnocorp (changes requested):

kossnocorp (changes requested):
Please rollback the change and use a custom format in your code. `d/M/yy` corresponds to the standard: https://unicode-org.github.io/cldr-staging/charts/37/summary/el.html#57dac0d1b36c1261

src/locale/el/_lib/formatLong/index.js
BanForFun: I didn't know that, but I can assure you that the vast majority of people in Greece write the year as four digits. Do you want me change it to `yyyy` or roll it back to `yy`?

kossnocorp: Sorry, I can't rely on subjective feelings. It's already challenging to support locales as I can only speak two languages, but date-fns supports dozens. Without a standard to consult, this will turn into chaos.

If you are 100% sure that the standard is incorrect, I will be happy to reconsider, but we'll have to do it in a separate pull-request with a review from other native speakers.

Please revert to `yy`.

I'm about to ship a new version, and if you can make this change and commit once again updated snapshot right now, it will go live right away.

BanForFun: Done

kossnocorp (approved):
Thank you a lot!

PR #1933, opened 2020-09-02
Fix closestTo should possibly return undefined

### LINKED ISSUE #1915: `closestTo` TS typings are incorrect
If you give the `closestTo` function an empty list of dates, it returns `undefined`. This is not represented in the TS typings.

V-Gutierrez:
Closes #1915 

closestTo can possibly return undefined and this is foreseen by typings

### DIFF
--- a/src/closestTo/test.js
+++ b/src/closestTo/test.js
@@ -91,18 +91,18 @@ describe('closestTo', function() {
     assert.deepEqual(result, new Date(2015, 7 /* Aug */, 31))
   })
 
-  it('converts undefined into empty array', function() {
+  it('returns undefined when undefined is passed as second argument', function() {
     var date = new Date(2014, 6 /* Jul */, 2).getTime()
     // $ExpectedMistake
     var result = closestTo(date, undefined)
-    assert(result == null)
+    assert(result === undefined)
   })
 
-  it('converts null into empty array', function() {
+  it('converts null into undefined', function() {
     var date = new Date(2014, 6 /* Jul */, 2).getTime()
     // $ExpectedMistake
     var result = closestTo(date, null)
-    assert(result == null)
+    assert(result === undefined)
   })
 
   it('throws TypeError exception if passed less than 2 arguments', function() {
--- a/src/closestTo/test.js
+++ b/src/closestTo/test.js
@@ -36,7 +36,7 @@ describe('closestTo', function() {
   it('returns undefined if the given array is empty', function() {
     var date = new Date(2014, 6 /* Jul */, 2).getTime()
     var result = closestTo(date, [])
-    assert(result == null)
+    assert(result === undefined)
   })
 
   it('returns `Invalid Date` if the given date is `Invalid Date`', function() {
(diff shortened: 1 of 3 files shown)


kossnocorp (changes requested):
Hey, thank you for the fix, but it won't have an effect because we generate typings from JSDoc definitions. Please update those instead.

wenderjean:
@kossnocorp The test file hosted in [/src/closestTo/test.js](https://github.com/date-fns/date-fns/blob/master/src/closestTo/test.js) is still relevant? We're discussing about the assertions on lines `36`, `94` and `101` and they look pretty weird actually. Should we take the opportunity to work on them?

1. The assertion in the line `36` could refer itself to a `falsy` value considering the comparison
2. The assertions in the lines `94` and `101` seem to be false positives :thinking: 

What do you think?

V-Gutierrez:
> @kossnocorp The test file hosted in [/src/closestTo/test.js](https://github.com/date-fns/date-fns/blob/master/src/closestTo/test.js) is still relevant? We're discussing about the assertions on lines `36`, `94` and `101` and they look pretty weird actually. Should we take the opportunity to work on them?
> 
> 1. The assertion in the line `36` could refer itself to a `falsy` value considering the comparison
> 2. The assertions in the lines `94` and `101` seem to be false positives thinking
> 
> What do you think?

That's right, if they still relevant I would gladly update these too !

kossnocorp:
Yes, please fix the tests as well, much appreciated! 36 def should check for `=== undefined`. Regarding 94 and 101, I'm not sure what to do with it. Do you have any suggestions?

Thank you, guys!

V-Gutierrez:
> Yes, please fix the tests as well, much appreciated! 36 def should check for `=== undefined`. Regarding 94 and 101, I'm not sure what to do with it. Do you have any suggestions?
> 
> Thank you, guys!

I choose to fix the tests for checking the current behavior of closestTo. Not all cases returns Invalid Date as said in docs but for this issue I believe those fixes will do for now.

src/closestTo/test.js
```diff
@@ -91,18 +91,18 @@ describe('closestTo', function() {
     assert.deepEqual(result, new Date(2015, 7 /* Aug */, 31))
   })
 
-  it('converts undefined into empty array', function() {
+  it('returns undefined when undefined is passed as second argument', function() {
```
hjJunior, line 94:
```suggestion
  it('returns undefined when the second argument is omitted', function() {
```

src/closestTo/test.js
```diff
@@ -91,18 +91,18 @@ describe('closestTo', function() {
...
     var date = new Date(2014, 6 /* Jul */, 2).getTime()
     // $ExpectedMistake
     var result = closestTo(date, undefined)
-    assert(result == null)
+    assert(result === undefined)
   })
 
-  it('converts null into empty array', function() {
+  it('converts null into undefined', function() {
```
hjJunior, line 101:
```suggestion
  it('returns undefined if the given array is null', function() {
```

ekilah:
@V-Gutierrez thanks for picking up the issue I created! 🎉

V-Gutierrez:
@kossnocorp Closing this one since I made further improvements in #1968

PR #1936, opened 2020-09-04
Update Webpack docs

undecaf:
Webpack did not include the supportedLocales in the build because the newContentRegExp was an absolute path whereas the ContextReplacementPlugin doc says: "If newContentResource is relative, it is resolved relative to the previous resource."


leshakoss (approved):
Thanks!

PR #1963, opened 2020-09-22
Improvement: minor changes in previous migrations

V-Gutierrez:
- Remove deprecated asserts
- Change 'var' to 'const' 
- Add @ts-expect-error flags where needed


kossnocorp (approved):
Thanks!

Issue #1965, opened 2020-09-23
Invalid date in parseISO function


aryanshridhar:
Hi everyone , 
As suggested in docs <a href ='https://date-fns.org/v2.16.1/docs/parseISO'>here</a>
<pre>parseISO('2018-13-32')
//=> Invalid Date
</pre>
But , seems like this error is also thrown when I pass an date as <code>1900-1-1</code> rather than passing  <code>1900-01-01</code> . 

For more clarification : 
<pre>
parseISO('1900-1-1');
//=> Invalid Date

parseISO('1900-01-01');
//=> Date Mon Jan 01 1900 00:00:00 GMT+0521 (India Standard Time)
</pre>

Beigelman:
Hi @aryanshridhar as far as i could tell it is a simple regex problem. I'm working on it right now.

andersk:
This doesn’t seem like a bug. `1900-1-1` is not valid as an ISO 8601 date.

politicsruinseverything:
> This doesn’t seem like a bug. `1900-1-1` is not valid as an ISO 8601 date.

Is there another way to interpret 1900-1-1 other than 1900-01-01? If not then it would seem like an improvement / benefit to the user to accept both.

andersk:
For a general date parser, sure. For an **ISO 8601** date parser, no. Python agrees:

```pycon
>>> datetime.datetime.fromisoformat('1900-01-01')
datetime.datetime(1900, 1, 1, 0, 0)
>>> datetime.datetime.fromisoformat('1900-1-1')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: Invalid isoformat string: '1900-1-1'
```

`parseISO` is clear about the syntax it intends to accept. Accepting dates outside of that syntax could mislead the user into assuming they’ve validated something that they haven’t.

politicsruinseverything:
When I call isValid(parseISO('111111-01')) I get a value of true.  Is that correct? It was not returning true in a different library I was using.

PR #1978, opened 2020-09-24
Feature/issue/#1965

Beigelman:
Feat: update parseISO to accept formats d and M. Example: parseISO('2020-1-1')

Migrate parseISO to TypeScript


andersk:
Linking #1965. Like I mentioned there, I’m not so sure `parseISO` should accept `2020-1-1`, since that’s not a valid ISO 8601 date.

PR #1985, opened 2020-09-25
Update docs to make localised date formats clearer

lexanth:
"PP" and "PPpp" use abbreviated months, whereas "PPP" and "PPPppp" use full month names. The existing example (29th May 1453) is in May, which is the only month (in English) where the abbreviated and full month names are the same.


kossnocorp (approved):
Great point, thanks!

PR #1989, opened 2020-09-27
Fix countReporter data type for node 14

O4epegb:
In Node v14 `fs.writeFile` validates incoming `data` type and number type is not acceptable, test runner crashes after that. I am aware that v12 is LTS right now, but fixing this issue a little bit ahead of time is a nice thing, I guess.


tan75:
@O4epegb Thank you for your PR - your help is much appreciated. 🙏
I am going to close this PR because this has already been fixed.

PR #1993, opened 2020-09-28
Migration to TS: getDays Functions

Beigelman:
Migration: migrating getDate, getDay, getDayOfYear, getDaysInMonth and getDaysInYear to TS.


kossnocorp (approved):
👏

PR #2005, opened 2020-10-02
Migration: differenceIn functions to TS [BULK]

V-Gutierrez:
Hello, since Hacktoberhell is on I decided to proceed with the migration with bulkier PR to prevent spam in the PR queue. All these changes have a specific branch so let me know if the review has become to compĺex and you want me to split it.

- Types needs improvement but it will do for this initial migration, let me know the best way to handle them.
- Previous PR related to differenceIn functions and dependencies has been closed and merged into this one

Welcome back!


src/_lib/getTimezoneOffsetInMilliseconds/index.ts
```diff
@@ -15,12 +15,15 @@ function getDateMillisecondsPart(date) {
  *
  * This function returns the timezone offset in milliseconds that takes seconds in account.
  */
-export default function getTimezoneOffsetInMilliseconds(dirtyDate) {
-  var date = new Date(dirtyDate.getTime())
-  var baseTimezoneOffset = Math.ceil(date.getTimezoneOffset())
+export default function getTimezoneOffsetInMilliseconds(
+  dirtyDate: Date | number
+): number {
+  const receivedDate = dirtyDate as Date
```
wenderjean, line 21: I noticed that `dirtyDate` may be a `Date` or `number`, converting a `number` to `Date` directly wouldn't be a mistake?

V-Gutierrez: That was spot on! 

I did it in a better way in the last commit    

The 'as' type assertion keyword just flags to the compiler that where it was invoked it should consider the specified type instead of the inferred type (which is number and leads to TypeScript error, since Date methods are used from this line and on), it does not change the variable type in fact.

However, if you have a better idea I could use some piece of advice :D

leshakoss:
Thanks for your PR! Closing it as the functions have already been migrated

PR #2011, opened 2020-10-07
Keep docs links inside date-fns.org

gibson042:
I found it a little jarring to bounce from date-fns.org into pages like https://github.com/date-fns/date-fns/blob/master/docs/unicodeTokens.md and https://github.com/date-fns/date-fns/blob/master/docs/upgradeGuide.md , which are the _source_ for documentation pages rather than nicely formatted documentation.

Are you willing to replace those links with proper date-fns.org targets?

### DIFF
--- a/src/_lib/protectedTokens/index.js
+++ b/src/_lib/protectedTokens/index.js
@@ -12,19 +12,19 @@ export function isProtectedWeekYearToken(token) {
 export function throwProtectedError(token, format, input) {
   if (token === 'YYYY') {
     throw new RangeError(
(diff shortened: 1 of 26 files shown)


tan75:
@gibson042 
Thank you for this PR - your help is much appreciated however we can not accept it. 
The reason we do not have our `date-fns.org` URLs referenced in the code is that our doc site is not available in some countries due to their local restrictions (the doc website is running on Google Firebase). 
Hense we decided to keep our documentation as is.

PR #2017, opened 2020-10-11
Migrate to TS: eachHourOfInterval

minitesh:


src/eachHourOfInterval/test.ts
```diff
@@ -84,7 +84,7 @@ describe('eachHourOfInterval', () => {
     assert.throws(block, RangeError)
   })
 
-  it('throws an exception if the interval is undefined', () => {
+  it.only('throws an exception if the interval is undefined', () => {
```
kossnocorp, line 87: Could you please remove it?

src/eachHourOfInterval/index.ts
```diff
@@ -32,7 +32,7 @@ import requiredArgs from '../_lib/requiredArgs/index'
  * //   Mon Oct 06 2014 15:00:00
  * // ]
  */
-export default function eachHourOfInterval(dirtyInterval, options) {
+export default function eachHourOfInterval(dirtyInterval: Interval, options?: {step?: number}): Date[] {
```
kossnocorp, line 35: I don't see `Interval` import.

kossnocorp (changes requested):
Please import `Interval` correctly and remove `only` from tests.

src/eachHourOfInterval/test.ts
minitesh: Yeah sorry my bad.
I realized it later on after I switched off my laptop :(
Will do the changes

src/eachHourOfInterval/index.ts
minitesh: There is no need to import.
There is a definition in `typings.d.ts`
```
type Interval = {
  start: Date | number
  end: Date | number
}
```

kossnocorp: The thing is that typings.d.ts is automatically generated and won't be a thing in the future, so we better use explicit imports. No worries, though, I will do it myself in the main branch.

kossnocorp (approved):
Thanks!

src/eachHourOfInterval/index.ts
minitesh: Ok.
I will take care of this in my next PRs.

Issue #2012, opened 2020-10-08
Stop returning Date from differenceInBusinessDays


kossnocorp:
[Right now `differenceInBusinessDays` returns `Date` if the arguments are invalid](https://github.com/date-fns/date-fns/blob/master/src/differenceInBusinessDays/index.js#L44). This is not right and we need to fix it.

suraj-desai:
Hey @kossnocorp can you assign it to me? I can give a shot

kossnocorp:
@suraj-desai, this is a breaking change and won’t be shipped until v3, so better I would do it myself.

suraj-desai:
okay good luck

suraj-desai:
Okay. no problem.

fturmel:
This was actually fixed by #2414 when the function got converted to TypeScript, and released in v2.21.2

PR #2022, opened 2020-10-13
Make differenceInBusinessDays return NaN instead of Invalid Date

dmarcr1997:
Saw there were already tests in place for this issue. I got them to pass by changing the return from new Date(NaN), which returns invalid date, to NaN.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,5 @@
+{
+  "name": "date-fns",
+  "version": "DON'T CHANGE; IT'S SET AUTOMATICALLY DURING DEPLOYMENT; ALSO, USE YARN FOR DEVELOPMENT",
+  "lockfileVersion": 1
+}
--- a/src/differenceInBusinessDays/index.js
+++ b/src/differenceInBusinessDays/index.js
@@ -41,7 +41,9 @@ export default function differenceInBusinessDays(
   var dateLeft = toDate(dirtyDateLeft)
   var dateRight = toDate(dirtyDateRight)
 
-  if (!isValid(dateLeft) || !isValid(dateRight)) return new Date(NaN)
+  // fix for issue # 2012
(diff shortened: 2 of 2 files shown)


package-lock.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "date-fns",
+  "version": "DON'T CHANGE; IT'S SET AUTOMATICALLY DURING DEPLOYMENT; ALSO, USE YARN FOR DEVELOPMENT",
+  "lockfileVersion": 1
+}
```
kossnocorp, line 5: Please remove the file from the PR.

src/differenceInBusinessDays/index.js
```diff
@@ -41,7 +41,9 @@ export default function differenceInBusinessDays(
   var dateLeft = toDate(dirtyDateLeft)
   var dateRight = toDate(dirtyDateRight)
 
-  if (!isValid(dateLeft) || !isValid(dateRight)) return new Date(NaN)
+  // fix for issue # 2012
+  // returning undefined to fix issue of returning a Date when either date is invalid
+  if (!isValid(dateLeft) || !isValid(dateRight)) return NaN
```
kossnocorp, line 46: No need for the comment.

```suggestion
  if (!isValid(dateLeft) || !isValid(dateRight)) return NaN
```

kossnocorp (changes requested):
Thank you for the change, but I will not merge it right now as it introduces a breaking change. However, it will be handy when I start working on v3. Tests actually check for Invalid Date, which internally represented by NaN, so the PR didn't fix tests 😉

fturmel:
Going to close this PR if you don't mind. This got fixed and merged when the function was converted to TypeScript (perhaps early by accident), and it's been in v2 releases for a while. See #2012 and #2414

Issue #2027, opened 2020-10-13
closestIndexTo with multiple indexes


robinglen:
Hello,

Really like the library, I just have a use case where they might be an array of objects with multiple dates that are the same.

```JavaScript
const now = new Date()
const items = [{
        id: 000001,
        date: now.setHours(18, 00, 01)
    },
    {
        id: 000002,
        date: now.setHours(18, 00, 04)
    },
    {
        id: 000003,
        date: now.setHours(15, 00, 01)
    },
    {
        id: 000004,
        date: now.setHours(18, 00, 04)
    }];

const dateArray = items.map(item => {
  return item.date
})
const result = closestIndexTo(now, dateArray); // returns 1
// what I want it to return is [1,3]
```

So for my use case, I want to return all the closes indexes not the first, is this possible or something you are interested in it doing?

Thanks,
Robin

Tanveeshs:
Created a pull request for this issue. Will be solved
Please assign this issue to me

thearegee:
Sorry I'm not sure I understand, you have created the PR here:
https://github.com/date-fns/date-fns/pull/2029

Not sure what you wanted me to do, but also, thanks :)

Tanveeshs:
I've made changes to the code to solve your issue.
Once the PR is accepted closestIndexTo will give multiple indexes

kossnocorp:
Adding this to `closestIndexTo` will introduce a breaking change that I would like to avoid. Besides, this will make function API inconsistent which is more important. I will accept a new function that does that, but not sure about the name. Maybe `allClosestIndicesTo`?

PR #2029, opened 2020-10-14
ClosestIndexOf Issue Solved

Tanveeshs:
Solved the closestIndexOf Issue
It will now return array of indexes of minimum distances


Tanveeshs:
This was an issue

Tanveeshs:
So created a pull request solving the issue

kossnocorp:
Thank you a lot for your effort, but this is a breaking change and also introduces an inconsistency to the API. See my comment: https://github.com/date-fns/date-fns/issues/2027#issuecomment-716977460

PR #2030, opened 2020-10-15
Use numerals in Finnish distances as well

ljani:
I cannot think any context where one would display eg. `Duration: hour` instead of `Duration: 1 hour`. [English seems to be using numerals as well.](https://github.com/date-fns/date-fns/blob/a91e059dce796c580d54d5894dd9f19348dd86fa/src/locale/en-US/_lib/formatDistance/index.js#L30)

Source: I'm a native speaker.

I hope the CI will run tests for me as I don't have Chrome. Also, could someone run `yarn locale-snapshots` for me as it fails to find `env` on my (Windows) box? Marking the PR as a draft until then.

### DIFF
--- a/src/locale/fi/_lib/formatDistance/index.js
+++ b/src/locale/fi/_lib/formatDistance/index.js
@@ -34,7 +34,7 @@ var formatDistanceLocale = {
   },
 
   xSeconds: {
-    one: 'sekunti',
+    one: '1 sekunti',
     other: '{{count}} sekuntia',
     futureTense: futureSeconds
   },
@@ -54,7 +54,7 @@ var formatDistanceLocale = {
   },
 
   xMinutes: {
-    one: 'minuutti',
+    one: '1 minuutti',
     other: '{{count}} minuuttia',
     futureTense: futureMinutes
   },
@@ -66,13 +66,13 @@ var formatDistanceLocale = {
   },
 
   xHours: {
-    one: 'tunti',
+    one: '1 tunti',
     other: '{{count}} tuntia',
     futureTense: futureHours
   },
 
   xDays: {
-    one: 'päivä',
+    one: '1 päivä',
     other: '{{count}} päivää',
     futureTense: futureDays
   },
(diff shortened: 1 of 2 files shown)


kossnocorp (approved):
For me, it looks good, but I don't know Finnish. @date-fns/i18n-fi what do you think about this change?

Pyppe:
I do believe it is, at least somewhat, context specific. The `futureTense` in particular seems a bit off: `vuoden kuluttua` -> `1 vuoden kuluttua` does not sound like an improvement to me – on the contrary.

juhovh:
Another native speaker here chiming in. 👋 

While I generally agree with the original patch of adding number `1` to the front of the duration of single units, I'm afraid that the slightly awkward future tense that @Pyppe mentioned is, in fact, an unintended consequence resulting from the automatic conversion. It doesn't sound very good to me either. If we want to have this change anyway, I'm suggesting to change it to something more like in [the attached patch](https://github.com/date-fns/date-fns/files/5443726/fix-future-distance.txt) to make it a bit less invasive, but I do admit that it makes the code slightly more convoluted. Would be happy to hear comments from @ljani as well.

ljani:
One could discuss `in a minute` vs `in 1 minute` with English locale as well, but I agree that in Finnish it's more awkward as @Pyppe mentioned.

To be honest, if you ask my opinion, I don't like `vuoden kuluttua` either, and would prefer something like `vuoden päästä`. [`kuluttua`](https://fi.wiktionary.org/wiki/kuluttua) refers to "after some event" where as [`päästä`](https://fi.wiktionary.org/wiki/p%C3%A4%C3%A4st%C3%A4#Postpositio) is used more like "after some time".

Personally, I'd prefer `minuutti sitten` over `1 minuutti sitten` as well, what do you think? With English, I'd also prefer `a minute ago` over `1 minute ago`.

@juhovh: Your patch looks good to me. Can you make the changes to this PR?

leshakoss:
Hi guys! Thanks a lot for your contribution. This seems like a dead PR at this point. Feel free to reopen it with the patch mentioned above.

PR #2037, opened 2020-10-16
Add yearStartsOn option for indicating the start of the fiscal/academic year (closes #2014)

### LINKED ISSUE #2014: Add support for fiscal year
# What 

A new `fiscalStart` option gets added to the following functions
  - startOfQuarter, startOfYear
  - endOfQuarter, endOfYear
  - isSameQuarter, isSameYear

This option is a number that can range from 1 to 12, with 1 representing January and 12 representing December.

# Example

```js
console.log(startOfQuarter(new Date('2020-07-01'), { fiscalStart: 5 })) // 2020-05-01
console.log(endOfQuarter(new Date('2020-07-01'), { fiscalStart: 5 })) // 2020-07-31

console.log(isSameQuarter(new Date('2020-05-01'), new Date('2020-06-01'), { fiscalStart: 5 })) // true

console.log(startOfYear(new Date('2020-08-01'), { fiscalStart: 6 })) // 2020-06-01
console.log(endOfYear(new Date('2020-08-01'), { fiscalStart: 6 })) // 2021-05-31

console.log(isSameYear(new Date('2020-12-30'), new Date('2021-01-01'), { fiscalStart: 6 })) // true
```

# Why

It is not uncommon to have businesses that start tracking their yearly reports from another month than January. This functionality would save a lot of converting to and from the fiscal year.

#  What could be improved

The naming is very specific. Perhaps we could make it more generic like `monthOffset` or something.

Beigelman:
This pull request adds a new `yearStartsOn` option to the following functions:

- startOfQuarter, startOfYear

- endOfQuarter, endOfYear

- isSameQuarter, isSameYear

This new feature allows us to count the start of a year from another month than January. The `yearStartsOn` option is a number that can range from 0 to 11, with 0 representing January and 11 representing December.

It is not uncommon to have businesses that start tracking their yearly reports from another month than January. This functionality would save a lot of converting to and from the fiscal year.

Closes #2014


src/endOfQuarter/index.ts
```diff
@@ -15,21 +16,32 @@ import requiredArgs from '../_lib/requiredArgs/index'
...
+export default function endOfQuarter(dirtyDate: Date | number, options?: YearStart): Date {
   requiredArgs(1, arguments)
 
   const date = toDate(dirtyDate)
   const currentMonth = date.getMonth()
-  const month = currentMonth - (currentMonth % 3) + 3
+
+  const diffToQuarterStartArray = [0, 1, 2, 0, 1, 2, 0, 1, 2, 0, 1, 2]
+
+  const month = currentMonth - diffToQuarterStartArray[currentMonth - (options ? options.yearStartsOn : 0)] + 3
```
kossnocorp, line 41: If in the future we decide to add more options this will break. Please use `??` instead:

```suggestion
  const month = currentMonth - diffToQuarterStartArray[currentMonth - (options?.yearStartsOn ?? 0)] + 3
```

src/endOfQuarter/index.ts
```diff
@@ -15,21 +16,32 @@ import requiredArgs from '../_lib/requiredArgs/index'
...
  */
-export default function endOfQuarter(dirtyDate: Date | number): Date {
+export default function endOfQuarter(dirtyDate: Date | number, options?: YearStart): Date {
   requiredArgs(1, arguments)
 
   const date = toDate(dirtyDate)
   const currentMonth = date.getMonth()
-  const month = currentMonth - (currentMonth % 3) + 3
+
+  const diffToQuarterStartArray = [0, 1, 2, 0, 1, 2, 0, 1, 2, 0, 1, 2]
```
kossnocorp, line 39: Could you please add a short comment about what is this and also move it out of the function?

src/endOfYear/index.ts
```diff
@@ -15,20 +19,35 @@ import requiredArgs from '../_lib/requiredArgs/index'
...
   requiredArgs(1, arguments)
 
-  var date = toDate(dirtyDate)
-  var year = date.getFullYear()
-  date.setFullYear(year + 1, 0, 0)
-  date.setHours(23, 59, 59, 999)
+  let date = toDate(dirtyDate)
+  const year = date.getFullYear()
+
+  if (options) {
```
kossnocorp, line 42: Same as above, if options change, this will break.

```suggestion
  if (typeof options?.yearStartsOn === 'number') {
```

src/isSameYear/index.js
```diff
@@ -1,32 +0,0 @@
-import toDate from '../toDate/index'
```
kossnocorp, line 1: Could you please recreate the commit so this file is renamed and not removed? This will make us lose all the history.

src/isSameYear/index.ts
```diff
@@ -0,0 +1,53 @@
...
+export default function isSameYear(
+  dirtyDate: Date | number,
+  dirtyDateToCompare: Date | number,
+  options?: YearStart
+  ): boolean {
+  requiredArgs(2, arguments)
+
+  if (options && options.yearStartsOn > toDate(dirtyDate).getMonth()){
+    throw Error('Month must be after the start of year param')
+  }
```
kossnocorp, line 45: At the first glance, this is a weird condition. Why does it here?

src/startOfQuarter/index.ts
```diff
@@ -16,19 +17,28 @@ import requiredArgs from '../_lib/requiredArgs/index'
...
-export default function startOfQuarter(dirtyDate: Date | number): Date {
+export default function startOfQuarter(dirtyDate: Date | number, options?: YearStart): Date {
   requiredArgs(1, arguments)
 
   const date = toDate(dirtyDate)
   const currentMonth = date.getMonth()
-  const month = currentMonth - (currentMonth % 3)
+  const diffToQuarterStartArray = [0, 1, 2, 0, 1, 2, 0, 1, 2, 0, 1, 2]
+
+  const month = currentMonth - diffToQuarterStartArray[currentMonth - (options ? options.yearStartsOn : 0)]
```
kossnocorp, line 40:
```suggestion
  const month = currentMonth - diffToQuarterStartArray[currentMonth - (options?.yearStartsOn ?? 0)]
```

src/startOfQuarter/index.ts
```diff
@@ -16,19 +17,28 @@ import requiredArgs from '../_lib/requiredArgs/index'
...
+ * //=> Mon Sep 01 2014 00:00:00
  */
-export default function startOfQuarter(dirtyDate: Date | number): Date {
+export default function startOfQuarter(dirtyDate: Date | number, options?: YearStart): Date {
   requiredArgs(1, arguments)
 
   const date = toDate(dirtyDate)
   const currentMonth = date.getMonth()
-  const month = currentMonth - (currentMonth % 3)
+  const diffToQuarterStartArray = [0, 1, 2, 0, 1, 2, 0, 1, 2, 0, 1, 2]
```
kossnocorp, line 38: Same here, please add a comment and move out of the function.

src/startOfYear/index.ts
```diff
@@ -15,20 +16,32 @@ import requiredArgs from '../_lib/requiredArgs/index'
...
 
   const cleanDate = toDate(dirtyDate)
   const date = new Date(0)
-  date.setFullYear(cleanDate.getFullYear(), 0, 1)
+
+  if (options) {
+    date.setFullYear(cleanDate.getFullYear(), options.yearStartsOn, 1)
+  } else {
+    date.setFullYear(cleanDate.getFullYear(), 0, 1)
+  }
```
kossnocorp, line 43:
```suggestion
  date.setFullYear(cleanDate.getFullYear(), options.yearStartsOn ?? 0, 1)
```

kossnocorp (changes requested):
Great job, thank you 👍 Please address my comments

src/isSameYear/index.js
Beigelman: I'm not quite sure how to do that. I have tried several ways but all seem to get the same result where one file is deleted and another is created. I think this happens because git does not track file history [https://stackoverflow.com/questions/47571168/git-does-not-track-history-after-rename-move-of-modified-file](url) . The only solution I found was to separate into two different commits.

src/isSameYear/index.ts
Beigelman: This condition ensures that the `dirtydate` is within the `startTime` and the `endTime`. For example, if he calls
```js
 isSameYear('01-2020', '12-2020', { yearStartsOn: 3 })
```
he would check if `12-2020` is between `04-2020` and `03-2021` which is true but `01-2020` is not between this interval so it doesn't make sense. So this condition is to make sure that the `dirtydate` provided is inside the year offset otherwise this function with this option wouldn't make sense. I hope I made sense hehe. But I think the error message could be better I'm thinking of a better one

zhirzh:
i think `startOfYear` is not correct. if target date is "1st March 2020" and year start is set to "April" then target date's year should start on "1st April 2019" but current implementation returns "1st April 2020"

```js
var d = new Date(2020, 2, 1)
console.log(d.toString())
//> Sun Mar 01 2020 00:00:00 GMT+0530 (India Standard Time)

var x = startOfYear(d, { yearStartsOn: 3 }) // year starts on `April`
console.log(x.toString())
//> Wed Apr 01 2020 00:00:00 GMT+0530 (India Standard Time)  ---  incorrect

// correct value should be "Mon Apr 01 2019 00:00:00 GMT+0530 (India Standard Time)"
```

PR #2042, opened 2020-10-21
Update test.js

simone-fracassetti:
In this case, const is usefull and better!


kossnocorp:
I reckon this is Hacktoberfest shenanigans. I suggest you instead make something useful for the project and follow the steps in this issue: https://github.com/date-fns/date-fns-typescriptify/issues/218.

Issue #2039, opened 2020-10-19
Feature Request: nextSaturday(...)


xErik:
Would it be helpful in general to add methods like `nextSaturday(...)` ?

```
var result = nextSaturday(new Date(2020, 10, 19))
// result = 24.10.2020 ...

// 19.10.2020 is a Monday
// The next Saturday is on the 24.10.2020


```

redbmk:
Could the naming be confusing? If it were Monday 2020-10-19 and I said "this Saturday", I would be referring to 2020-10-24. If I said "next Saturday" I would probably mean 2020-10-31. 🎃 👻 

I don't know how common that is though.

fturmel:
Thanks for the suggestion! This actually shipped in v.2.20 (see #2291)

PR #2046, opened 2020-10-22
Add next function

robtarr:
This is in response to https://github.com/date-fns/date-fns/issues/2039. The function signature is a bit different, but it allows for checking for any upcoming day.


robtarr:
Would it be better to create individual functions for each Day? `nextSaturday`, `nextFriday`, etc.?

kossnocorp:
Hey, thank you a lot for the function. I like the addition 👍 I would only like to change the name to make it more explicit.

> Would it be better to create individual functions for each day, like nextSaturday, nextFriday, etc.?

We could add it, but that's very specific and niche, not sure so many people will prefer those functions over `next`.

src/next/index.ts
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @example
+ * // The Sunday after Jan 1, 2020
+ * next(0, new Date('1/1/2020'));
+ * //=> 2020-01-05T05:00:00.000Z
+ *
+ */
+export default function next(
+  day: number,
+  date: date
```
kossnocorp, line 24: Please change the order of the arguments. All the functions have `date` as the first argument.

src/next/index.ts
```diff
@@ -0,0 +1,30 @@
...
+ * @param {date} date - Date to start counting from
+ * @returns {date} the date of the next specified day of the week
+ *
+ * @example
+ * // The Sunday after Jan 1, 2020
+ * next(0, new Date('1/1/2020'));
+ * //=> 2020-01-05T05:00:00.000Z
+ *
+ */
+export default function next(
```
kossnocorp, line 22: Let's rename the function to `nextDay`. It will make it consistent with other day functions like `getDay` or `setDay`.

src/next/index.ts
```diff
@@ -0,0 +1,30 @@
...
+  day: number,
+  date: date
+) : Date {
+  const today = getDay(date)
+  const offset = day > today ? day : day + 7
+
+  return add(date, { days: offset - today })
```
kossnocorp, line 29: Please use `addDays` instead as it is more lightweight (as a dependency) than `add`.

src/next/index.ts
```diff
@@ -0,0 +1,30 @@
...
+ * @name next
+ * @category Weekday Helpers
+ * @summary Returns the date of the next specified day of the week
+ *
+ * @description
+ * Returns the date of the next specified day of the week
+ *
+ * @param {number} day - Day of the week
+ * @param {date} date - Date to start counting from
+ * @returns {date} the date of the next specified day of the week
```
kossnocorp, line 14: Please add JSDoc types ([an example](https://github.com/date-fns/date-fns/blob/master/src/addDays/index.ts#L17-L20)). I know that TSDoc doesn't need those, but we still in transition and need it for generating the docs.

src/next/index.ts
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @example
+ * // The Sunday after Jan 1, 2020
+ * next(0, new Date('1/1/2020'));
+ * //=> 2020-01-05T05:00:00.000Z
+ *
+ */
+export default function next(
+  day: number,
```
kossnocorp, line 23: Let's make it optional and use `today` if `day` is not passed. I think it covert two use cases:

- Actually set the next day.
- Next given day (Saturday, Friday, etc.).

kossnocorp (changes requested):
Thank you a lot, please address my comments

src/next/index.ts
robtarr: I think I fixed this like the example, it looked like it was just the case of the types?

robtarr:
@kossnocorp, thanks for the feedback. I made some updates. I also renamed `today` as it seemed confusing as it's not actually today. I looked for a convention in the code and didn't see anything.

tan75:
Closed via https://github.com/date-fns/date-fns/pull/2214

PR #2066, opened 2020-10-30
Add weekStartsOn options to getDay

smmoosavi:
Most of the week-related functions have `weekStartsOn` options, but `getDay` doesn't.

```js
// Which day of the week is 29 February 2012?
getDay(new Date(2012, 1, 29))
//=> 3

// Which day of the week is 29 February 2012, with a weekStartsOn of Monday:
getDay(new Date(2012, 1, 29), { weekStartsOn: 1 })
//=> 2
```

### DIFF
--- a/src/getDay/index.ts
+++ b/src/getDay/index.ts
@@ -1,5 +1,7 @@
+import toInteger from '../_lib/toInteger/index'
 import toDate from '../toDate/index'
 import requiredArgs from '../_lib/requiredArgs/index'
+import { WeekStartOptions } from '../types'
 
 /**
  * @name getDay
@@ -14,6 +16,9 @@ import requiredArgs from '../_lib/requiredArgs/index'
  * - [Changes that are common for the whole library](https://github.com/date-fns/date-fns/blob/master/docs/upgradeGuide.md#Common-Changes).
  *
  * @param {Date|Number} date - the given date
+ * @param {Object} [options] - an object with options.
+ * @param {Locale} [options.locale=defaultLocale] - the locale object. See [Locale]{@link https://date-fns.org/docs/Locale}
+ * @param {0|1|2|3|4|5|6} [options.weekStartsOn=0] - the index of the first day of the week (0 - Sunday)
  * @returns {0|1|2|3|4|5|6} the day of week, 0 represents Sunday
  * @throws {TypeError} 1 argument required
  *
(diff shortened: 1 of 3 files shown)


kossnocorp:
Hey! Could you please explain the rationale behind the change? The day is just the index (of `["Monday", "Tuesday", ...]`), which is by itself confuses a lot of people, introducing the option will only make it worse. So unless there's a specific reason (like it enables integration with the Jalali calendar) and we find a better name for it, I won't be merging it. Sorry!

Mehdi-Hp:
FYI, This was the original issue that raised this one here
https://github.com/smmoosavi/date-fns-jalali/issues/2

smmoosavi:
the `setDay` and `getDay` functions should be symmetrical.

example:

```js
const date = setDay(new Date(), 3, {weekStartsOn: 1})
const day = getDay(date)
assert.equal(day, 3) // false: 3 !== 2
```

And yes, the options argument is required for the Jalali calendar. In Jalali, calendar weeks start from Saturday.

smmoosavi:
can you remove the `hacktoberfest-accepted` label and merge this pull request? we really need the `weekStartsOn` option in `getDay` function

kossnocorp:
Sorry, this is not a change we can accept, so I have to close the PR. If you're ready to discuss the problem and come up with an acceptable approach, please [open a discussion](https://github.com/date-fns/date-fns/discussions/new) and describe the issue in detail with links to the code.

Adding this option to `setDay` was a mistake and will be removed in the next major version.

First of all, the day is a static value and has nothing to do with `weekStartsOn`. Look at it as an array where Sunday is always the first element and Saturday is the last.

Secondly, the day is already a source of confusion as  i.e. `new Date(2021, 1, 1)` represents the 1st of February and not January. Altering it on a locale change is just an unnecessary brain twister.

Mehdi-Hp:
@kossnocorp So I think we can accept a convention here, where "Saturday" does not mean the start of a week. It's just the first item in an array of week days.  
If we agree on that, don't you think we need a `getWeekDay` method, which returns how many days is passed from first week day?

kossnocorp:
@Mehdi-Hp I think you're right. We just need to come up with a non-confusing name. `getWeekDay` won't cut it because weekday and day are synonyms (not to confuse with the date, which the number of days since the begging of the month).

`getDaysSinceWeekStart` is a good option. Another question is how to call a setter.

Anyway, PR is welcome!

smmoosavi:
what about `setDaysSinceWeekStart`?

kossnocorp:
@smmoosavi sounds good to me!

smmoosavi:
another suggestion: `getDeyOfWeek` and `setDeyOfWeek`.

PR #2080, opened 2020-11-12
Basque (eu) lange correction

inigoiparragirre:
Datefomats is modified for basque language correct translation, because there was some text from spanish copied incorrectly.


kossnocorp (approved):
Thank you!

PR #2090, opened 2020-11-16
migrate formatDistance to TS

a-korzun:
Fixes https://github.com/date-fns/date-fns-typescriptify/issues/32

### DIFF
--- a/src/formatDistance/index.ts
+++ b/src/formatDistance/index.ts
@@ -7,10 +7,10 @@ import cloneObject from '../_lib/cloneObject/index'
 import getTimezoneOffsetInMilliseconds from '../_lib/getTimezoneOffsetInMilliseconds/index'
 import requiredArgs from '../_lib/requiredArgs/index'
 
-var MINUTES_IN_DAY = 1440
-var MINUTES_IN_ALMOST_TWO_DAYS = 2520
-var MINUTES_IN_MONTH = 43200
-var MINUTES_IN_TWO_MONTHS = 86400
+const MINUTES_IN_DAY = 1440
+const MINUTES_IN_ALMOST_TWO_DAYS = 2520
+const MINUTES_IN_MONTH = 43200
+const MINUTES_IN_TWO_MONTHS = 86400
 
 /**
  * @name formatDistance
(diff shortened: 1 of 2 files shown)


tan75 (changes requested):
Please resolve conflicts

tan75:
Hello @a-korzun,

could you please resolve conflicts in your code?

src/formatDistance/test.ts
```diff
@@ -1,7 +1,6 @@
-// @flow
 /* eslint-env mocha */
 
-import assert from 'power-assert'
```
tan75, line 4: Is there a reason for a this switch from power-assert to assert?

tan75 (changes requested):

src/formatDistance/test.ts
a-korzun: Nope. Updated

a-korzun:
> Hello @a-korzun,
> 
> could you please resolve conflicts in your code?

Hi! Don't see any conflicts, may you provide a bit more information about it?
Just updated PR in order to fix migration from `power-assert`.

tan75:
> > Hello @a-korzun,
> > could you please resolve conflicts in your code?
> 
> Hi! Don't see any conflicts, may you provide a bit more information about it?
> Just updated PR in order to fix migration from `power-assert`.

@a-korzun 
It turns out we are using assert now rather than power-assert and you were right when changing it. 
Could you please switch it back? 
Apologies for the confusion!

a-korzun:
@tan75 Ok, done.

tan75 (approved):

src/formatDistance/index.ts
```diff
@@ -107,41 +107,48 @@ var MINUTES_IN_TWO_MONTHS = 86400
...
  */
-export default function formatDistance(dirtyDate, dirtyBaseDate, dirtyOptions) {
+
+export interface Options {
+  includeSeconds?: boolean
+  addSuffix?: boolean
+  locale?: Locale
+}
+
+export default function formatDistance(dirtyDate: Date | number, dirtyBaseDate: Date | number, dirtyOptions?: Options) {
```
kossnocorp, line 130: Please import and use shared types from https://github.com/date-fns/date-fns/blob/master/src/types.ts

```suggestion
export default function formatDistance(dirtyDate: Date | number, dirtyBaseDate: Date | number, options : LocaleOptions & { includeSeconds?: boolean, addSuffix?: boolean } = {}) {
```

I appreciate the effort of exporting the type, but unless it's a common approach in the whole library, we don't want to do it. So please remove the exported type and put the type inline as I suggested. In the future, we might opt to export option types, but we'll need to think about the naming convention first.

kossnocorp (changes requested):
Looks good, thanks! Please address my comment about the options type.

a-korzun:
No problem! Done

tan75 (approved):
Thank you!

PR #2091, opened 2020-11-16
migrate formatDistanceStrict to TS

a-korzun:
Fixes https://github.com/date-fns/date-fns-typescriptify/issues/33

### DIFF
--- a/src/formatDistanceStrict/index.ts
+++ b/src/formatDistanceStrict/index.ts
@@ -170,7 +170,7 @@ export default function formatDistanceStrict(
     unit?: 'second' | 'minute' | 'hour' | 'day' | 'month' | 'year',
     roundingMethod?: 'floor' | 'ceil' | 'round',
   } = {}
-) {
+): string {
   requiredArgs(2, arguments)
 
   const locale = options.locale || defaultLocale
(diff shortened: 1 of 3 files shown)


src/formatDistanceStrict/index.ts
```diff
@@ -155,37 +156,41 @@ var MINUTES_IN_YEAR = MINUTES_IN_DAY * 365
...
 export default function formatDistanceStrict(
-  dirtyDate,
-  dirtyBaseDate,
-  dirtyOptions
-) {
+  dirtyDate: Date | number,
+  dirtyBaseDate: Date | number,
+  options: LocaleOptions & {
+    addSuffix?: boolean,
+    unit?: 'second' | 'minute' | 'hour' | 'day' | 'month' | 'year',
```
tan75, line 170: Could you please move this to [types.ts](https://github.com/date-fns/date-fns/blob/master/src/types.ts) and import it from that file?
see example below:
```
export type Unit =
  | 'year'
  | 'quarter'
  | 'month'
  | 'week'
  | 'day'
  | 'hour'
  | 'minute'
  | 'second'
```

tan75: As well, could you please change the remaining `var`s.
Many thanks!

tan75 (changes requested):
Thanks a lot for this PR!
Please address my comments!

src/formatDistanceStrict/index.ts
a-korzun: Yeah, sure! Will do it asap

tan75 (approved):
Thank you!

Issue #1849, opened 2020-06-19
parse does seem to observe the reference date when parsing a 2-digit year with yyyy


doberkofler:
I already created SR #1569 to advocate an option for strict parsing when using the function parse but I now discovered that `parse` also seems to have a strange behavior when not strictly parsing a date.

`parse('06/19/20', 'MM/dd/yyyy', new Date());` converts the string to the year 20 but should it not take the century from the `reference date` and therefore convert it to the year 2020?

lamhungypl:
I have same problem

```javascript
console.log({ date }, isValid(parse(date, 'dd-MM-yyyy', new Date())));
```

![Screen Shot 2020-07-24 at 14 54 26](https://user-images.githubusercontent.com/43105576/88371118-a0e38600-cdbd-11ea-940d-6f6db27179ae.png)

czabaj:
Also relates to #1998 and #2072 and probably more and I encountered the same problem yesterday.

The problem IMHO is, that parse _should_ accept _only_ zero-padded digits, e.g.

`20` is not a valid pattern for `yyyy`
`202` is not a valid pattern for `yyyy`
`2020` is a valid pattern for `yyyy`
`0020` is a valid pattern for `yyyy` and year 20

czabaj:
I'm willing to fix this, but I need to elaborate more on this.

* shall there be some sort of strict flag for `parse`, or shall we rather introduce a new `validate` function?
* numbers which exceed repeated tokens length are generally valid, e.q. 12 is valid for token `m` and 2020 is valid for tokens `yy`, but `yy` should not be valid for, say 0020, i.e. numbers which exceeds tokens must not be zeroes (zero padded), right?

doberkofler:
Thank you for caring. In my opinion the parsing should be as strict as possible to prevent any unwanted interpretation of a date.

> * shall there be some sort of strict flag for `parse`, or shall we rather introduce a new `validate` function?

I would add a strict flag and even prefer it to be on by default although this clearly might break existing code.

> * numbers which exceed repeated tokens length are generally valid, e.q. 12 is valid for token `m` and 2020 is valid for tokens `yy`, but `yy` should not match 0020, i.e. numbers which exceeds tokens must not be zeroes (zero padded), right?

I would also not allow to parse a 2-digit month if only one m-token is given and the same for all other tokens.
I my opinion the date must exactly match the format without any exception and if someone needs a more relaxed parsing he can just opt out of the strict mode.

[...]

PR #2096, opened 2020-11-22
feat: Add strict parse feature

czabaj:
This PR adds a new option `strict` to [`parse`](https://date-fns.org/v2.16.1/docs/parse) to enforce the same length of the date part in the date string as the length of tokens in the format string.

See #1849

### DIFF
--- a/src/parse/_lib/parsers/index.js
+++ b/src/parse/_lib/parsers/index.js
@@ -42,21 +42,6 @@ var timezonePatterns = {
   extendedOptionalSeconds: /^([+-])(\d{2}):(\d{2})(:(\d{2}))?|Z/
 }
 
-function parseNumericPattern(pattern, string, valueCallback) {
-  var matchResult = string.match(pattern)
-
-  if (!matchResult) {
-    return null
-  }
-
-  var value = parseInt(matchResult[0], 10)
-
-  return {
-    value: valueCallback ? valueCallback(value) : value,
-    rest: string.slice(matchResult[0].length)
-  }
-}
-
 function parseTimezonePattern(pattern, string) {
   var matchResult = string.match(pattern)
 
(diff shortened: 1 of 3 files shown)


src/parse/_lib/parsers/index.js
```diff
@@ -87,81 +72,32 @@ function parseTimezonePattern(pattern, string) {
   }
 }
 
-function parseAnyDigitsSigned(string, valueCallback) {
-  return parseNumericPattern(
-    numericPatterns.anyDigitsSigned,
-    string,
-    valueCallback
```
czabaj, line 94: I established one universal function for parsing numbers, it handles un/-signed, limits length, and could enforce minimum length in strict mode. I needed access to many bits of information, such as length of tokens, weather `options.strict` is the matching number as well and this information was scattered across different functions and hard to maintain. This change significantly reduced the code complexity and shall have negligible performance implications.

src/parse/test.js
```diff
@@ -2650,4 +2650,64 @@ describe('parse', function() {
...
+    })
+
+    // it('valid ISO date year larger than tokens', function() {
+    //   var result = parse('10476', 'yyyy', referenceDate, { strict: true })
+    //   var expected = new Date(10476, 0 /* Jan */, 1)
+    //   var dateTimeString = '10476-01-01'
+    //   var formatString = 'yyyy-MM-dd'
+    //   var result = parse(dateTimeString, formatString, referenceDate)
+    //   assert.deepEqual(result, expected)
+    // })
```
czabaj, line 2711: This test is commented out as it is currently impossible to fulfill.

In #1849 I stated, that year longer than `tokens.length` shall be considered as valid, as `10111-01-01` is a valid date string for format `yyyy-MM-dd` and year 10111. But it is currently impossible to make this work for scenarios, where date parts are not delimited, e.g. in format string `yyyyMMdd` and input `202000101` is it a valid number 20200 or is month zero-padded improperly as 001? One could argue that we could treat month as always at most 2 digits, but then there are other tokens, which could always exceed the number of tokens, and if two such tokens are combined without delimiter and one of it exceeds tokens length it may be ambiguous to which this excessive number belongs. Also, a single occurrence of formatting token does not enforce length, e.g. `R` could be any number, i.e. any length of signed digits, if not delimited with other similar tokens, we are in trouble.

czabaj:
## Overall test of strict mode

I temporary hardcoded `strict` to `true` to test the strict option with all tests and it performed well, only this test failed

```
it('zero-padding', function() {
    var result = parse('1', 'KK', referenceDate)
    assert.deepEqual(result, new Date(1986, 3 /* Apr */, 4, 1))
})
```

which IMHO shall return `Invalid Date` in strict mode, for format string `KK` requires two digits.

## Regarding EcmaScript version

The source code contains a wide range of ES syntax, from ES5 with keywords `var` and `function` to ES6 with `let`, `const` and fat arrow functions. I choose the modern flavor for tests, as it runs in Node.js or Chrome, and I kept the library code in ES5. I also read the CONTRIBUTION.md and saw a note about migration to TypeScript, but rewriting the parsing code to TypeScript is a much larger project, than adding a strict parse feature. What do you recommend me to do? Use ES6 syntax and rely on babel to transpile it for supported browsers? Write library code in ES5? Or use TypeScript? I'm quite comfortable with TypeScript btw.

juliushuck:
I need this

PR #2104, opened 2020-11-26
docs: Support TypeScript in REPL example

czabaj:
I tried REPL with `./src/parse` and it crashed in an attempt to load a .ts file. The fix is simple, tell the `babel-node` to load also `.ts` files with a `-x` flag 🎉


kossnocorp (approved):
Thank you!

PR #2108, opened 2020-11-30
Fix dutch locale month may

bryantebeek:
Noticed a small typo in the Dutch translations.


kossnocorp:
Hey! [According to CLDR](https://unicode-org.github.io/cldr-staging/charts/37/summary/nl.html), it's `mei`:

CleanShot 2021-02-05 at 13 25 13@2x

Maybe it looks weird to you, but we stick to the standard so we can have a single source of the truth.

PR #2112, opened 2020-12-08
Add Type Guard to isDate

willium:
Apply a [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates) to the return type of `isDate`


tan75:
hi @willium,

Thank you for your PR!
Im going to close this as this function has already been migrated to TS.

PR #2117, opened 2020-12-10
Update formatRFC3339 documentation

fungiboletus:
It should be about the RFC 3339 (Date and Time on the Internet: Timestamps) and not the ISO 3339 (Tractors and Machinery for Agriculture and Forestry - Classification and Terminology).


kossnocorp (approved):
Appreciate it!

PR #2125, opened 2020-12-21
Fix formatDuration week translation in pt and pt-BR locales

FusRoDah061:
`formatDuration` output when using pt-BR locale was the following:
formatDuration_wrong

_mês_ translates to _month_. _Week_ in Portuguese is translated as _semana_, so the correct output is:
formatDuration_right


kossnocorp (approved):
Thanks!

PR #1495, opened 2019-10-22
[1310] round timezones offset in milliseconds

kpr-hellofresh:
## Description
Chrome and FireFox parse time zone offsets differently

## Reference
[Issue 1310](https://github.com/date-fns/date-fns/issues/1310)

## Checklist:
* [x]  - run lint
* [x] - run tests
* [x] - rebuild FP functions, typings and indices

## How to test
1. Build the package `env PACKAGE_OUTPUT_PATH="$(pwd)/../PATH-TO-YOUR-MODULE/node_modules/date-fns" ./scripts/build/package.sh`
2. Copy it into your project
3. Test with different timezones and dates

## Screenshots
Screen Shot 2019-10-22 at 01 38 27
Screen Shot 2019-10-22 at 14 36 59
Screen Shot 2019-10-22 at 15 37 17
Screen Shot 2019-10-22 at 23 37 39
Screen Shot 2019-10-23 at 00 38 11

PR #2162, opened 2021-01-14
fix(format): Fix round-error for Firefox in AKST timezone

### LINKED ISSUE #2161: v2 regression: `format` is off by one in AKST timezone in Firefox 84
## Context

We are migrating from date-fns v1 and I have setup experiments to run both v1 and v2 code paths at the same time and report differences. I caught this issue because in the Alaskan timezone users who are using Firefox 84 were showing up in my reports as having different function return values from v1.

Additional context: https://twitter.com/kamranayub/status/1349482511617961985

Related issues: #1310, #1495

## The Bug

In Firefox 84 when your machine is using the Anchorage/Alaska timezone, date-fns v2 `format` function is incorrectly formatting the string and is off by one. Suspicion is that `getTimezoneOffset()` is returning `599.6` and that is creating a rounding error.

In Chrome, it is OK and works as expected (`getTimezoneOffset() => 599`). In `date-fns` v1 in Firefox 84, it also works as expected. So I believe this to be a regression in the `format` function (or deeper).

## Reproduction

1. Change timezone to `Anchorage/Alaska` on your machine
2. Open Firefox 84
3. Visit this codesandbox: https://codesandbox.io/s/date-fns-v2-format-alaska-rsjji?file=/index.html
4. Observe the output in the console

### Expected Results

    00:30

### Actual Results

    59:30

kamranayub:
Closes #2161, Related: #1495

To run the test:

    yarn test --browser LocalFirefox

## Context

In #1495 there was a fix to `Math.ceil` timezone offsets. This must have fixed some issues but in this case, the `format` function returns the incorrect format when trying to format a duration due to the rounded-up offset.

I tried `Math.floor` but that made other tests fail. Instead I propose using `Math.trunc` which will chop off any additional digits from the offset which I think will match what Chrome does. There could still be edge cases I'm missing so would like additional feedback.

## Changes

- Add failing test case when run under Firefox 84
- Switch to `Math.trunc` which makes all existing tests pass and the FF test pass

## Open Questions

1. I would guess there's additional configuration needed to run the suites in Sauce Labs Firefox? Can someone help with that (or follow-up in separate PR?)
2. Commit hooks keep formatting the files I touch which mixes in format changes. I know most folks don't like mixing behavior with formatting in PRs, should I keep those changes out like I've been trying?


config/karma.js
```diff
@@ -11,56 +11,56 @@ const sauceLabsLaunchers = {
     base: 'SauceLabs',
     platform: 'OS X 10.11',
     browserName: 'safari',
-    version: '9.0'
+    version: '9.0',
```
kamranayub, line 14: Ugh formatting

src/_lib/getTimezoneOffsetInMilliseconds/index.js
```diff
@@ -17,7 +17,7 @@ function getDateMillisecondsPart(date) {
  */
 export default function getTimezoneOffsetInMilliseconds(dirtyDate) {
   var date = new Date(dirtyDate.getTime())
-  var baseTimezoneOffset = Math.ceil(date.getTimezoneOffset())
+  var baseTimezoneOffset = Math.trunc(date.getTimezoneOffset())
```
AndersDJohnson, line 20: @kamranayub Not sure but looks like maybe [`Math.trunc` isn't supported in Internet Explorer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc#browser_compatibility) - is IE support a requirement for this project? Maybe not - maybe consumers that need to support IE are able to polyfill that function and/or transpile this dependency with Babel `preset-env` or whatever.

AndersDJohnson: @kamranayub Maybe we could do something like this instead:

```suggestion
  var offset = date.getTimezoneOffset();
  var baseTimezoneOffset = offset < 0 ? Math.ceil(offset) : Math.floor(offset);
```

kamranayub: The problem are values that aren't integers, which seems to be an issue in FF... they can still be positive or negative and have decimal digits that would need to be chopped off. There could be another approach that avoids using Math.trunc but yeah it seems core-js polyfills this.

kamranayub:
I just noticed that this fn was updated recently. Confirming in FF that the issue is fixed and if so I can close this.

kamranayub:
2.19.0 fixes this issue (`getTimezoneOffset` is no longer used). Might still be valuable to add Karma launcher for FF to catch issues like this in the future. Closing for now since no longer have a bug.

PR #2173, opened 2021-01-24
Intl format relative

tan75:
#### New Feature
Implement intlFormatDistance for date-fns

### DIFF
--- a/src/constants/index.ts
+++ b/src/constants/index.ts
@@ -9,6 +9,18 @@
 export const daysInWeek = 7
 
 /**
+ * Days in 1 year
+ * One years equals 365.2425 days according to the formula:
+ *
+ * > Leap year occures every 4 years, except for years that are divisable by 100 and not divisable by 400.
+ * > 1 mean year = (365+1/4-1/100+1/400) days = 365.2425 days
+ *  @constant
+ *  @type {number}
+ *  @default
+ */
+export const daysInYear: number = 365.2425
+
+/**
  * Maximum allowed time.
  *
  * @name maxTime
@@ -117,3 +129,53 @@ export const secondsInHour = 3600
  * @default
  */
 export const secondsInMinute = 60
+
+/**
+ * Seconds in 1 day
+ *
+ * @name secondsInDay
+ * @constant
+ * @type {number}
+ * @default
+ */
+export const secondsInDay = secondsInHour * 24
+
+/**
+ * Seconds in 1 week
+ *
+ * @name secondsInWeek
+ * @constant
+ * @type {number}
+ * @default
+ */
+export const secondsInWeek = secondsInDay * 7
+
+/**
+ * Seconds in 1 year
+ *
+ * @name secondsInYear
+ * @constant
+ * @type {number}
+ * @default
+ */
+export const secondsInYear = secondsInDay * daysInYear
+
+/**
+ * Seconds in 1 month
+ *
+ * @name secondsInMonth
+ * @constant
+ * @type {number}
+ * @default
+ */
+export const secondsInMonth = secondsInYear / 12
+
+/**
+ * Seconds in 1 quarter
+ *
+ * @name secondsInQuarter
+ * @constant
+ * @type {number}
+ * @default
+ */
+export const secondsInQuarter = secondsInMonth * 3
--- a/src/intlFormatDistance/index.ts
+++ b/src/intlFormatDistance/index.ts
@@ -177,9 +177,9 @@ export default function intlFormatDistance(
       unit = 'minute'
     } else if (
       Math.abs(diffInSeconds) < secondsInDay &&
-      (value = differenceInHours(dateLeft, dateRight)) &&
       Math.abs(differenceInCalendarDays(dateLeft, dateRight)) < 1
     ) {
+      value = differenceInHours(dateLeft, dateRight)
       unit = 'hour'
     } else if (
       Math.abs(diffInSeconds) < secondsInWeek &&
--- a/src/intlFormatDistance/index.ts
+++ b/src/intlFormatDistance/index.ts
@@ -1,0 +1,232 @@
+import { IntlOptionsUnit } from 'src/types'
+import {
+  secondsInDay,
+  secondsInHour,
+  secondsInMinute,
+  secondsInMonth,
+  secondsInQuarter,
+  secondsInWeek,
+  secondsInYear,
+} from '../constants'
+import differenceInCalendarDays from '../differenceInCalendarDays/index'
+import differenceInCalendarMonths from '../differenceInCalendarMonths/index'
+import differenceInCalendarQuarters from '../differenceInCalendarQuarters/index'
+import differenceInCalendarWeeks from '../differenceInCalendarWeeks/index'
+import differenceInCalendarYears from '../differenceInCalendarYears/index'
+import differenceInHours from '../differenceInHours/index'
+import differenceInMinutes from '../differenceInMinutes/index'
+import differenceInSeconds from '../differenceInSeconds/index'
+import toDate from '../toDate/index'
+import requiredArgs from '../_lib/requiredArgs/index'
+
+interface Options {
+  unit?: IntlOptionsUnit
+  locale?: Intl.UnicodeBCP47LocaleIdentifier
+  localeMatcher?: Intl.RelativeTimeFormatLocaleMatcher
+  numeric?: Intl.RelativeTimeFormatNumeric
+  style?: Intl.RelativeTimeFormatStyle
+}
+
+/**
+ * @name intlFormatDistance
+ * @category Common Helpers
+ * @summary Formats distance between two dates in a human-readable format
+ * @description
+ * The function calculates the difference between two dates and either picks the most appropriate unit
+ * depending on the distance (the less the distance the smaller the unit) or allows the user to specify a unit.
+ * If the unit is specified, it will be applied accordingly. Otherwise - see the table below:
+ *
+ * |    Distance between dates   |         Result         |
+ * |-----------------------------|------------------------|
+ * |  0 seconds                  | now                    |
+ * |  1 second                   | in 1 second            |
+ * |  1 second ago               | 1 second ago           |
+ * |  2-59 seconds               | in X seconds           |
+ * |  60 seconds                 | in 1 minute            |
+ * |  60 seconds ago             | 1 minute ago           |
+ * |  1 minute                   | in 1 minute            |
+ * |  1 minute ago               | 1 minute ago           |
+ * |  2-59 minutes               | in X minutes           |
+ * |  from 2 to 59 minutes ago   | X minutes ago          |
+ * |  60 minutes                 | in 1 hour              |
+ * |  60 minutes ago             | 1 hour ago             |
+ * |  1 hour                     | in 1 hour              |
+ * |  1 hour ago                 | 1 hour ago             |
+ * |  2-23 hours                 | in X hours             |
+ * |  from 2 to 23 hours ago     | X hours ago            |
+ * |  24 hours                   | in 1 day               |
+ * |  1 day                      | tomorrow               |
+ * |  1 day ago                  | yesterday              |
+ * |  2-6 days                   | in X days              |
+ * |  from 2 to 6 days ago       | X days ago             |
+ * |  7 days                     | next week              |
+ * |  7 days ago                 | last week              |
+ * |  X weeks                    | in X weeks             |
+ * |  X weeks ago                | X weeks ago            |
+ * |  1 month                    | next month             |
+ * |  1 month ago                | last month             |
+ * |  X months                   | in X months            |
+ * |  X months ago               | X months ago           |
+ * |  1 quarter                  | next quarter           |
+ * |  1 quarter                  | last quarter           |
+ * |  X quarters                 | in X quarters          |
+ * |  X quarters ago             | X quarters ago         |
+ * |  1 year                     | next year              |
+ * |  1 year ago                 | last year              |
+ * |  X years                    | in X years             |
+ * |  X years ago                | X years ago            |
+ *
+ *
+ * @param {Date|Number} date - the date
+ * @param {Date|Number} baseDate - the date to compare with.
+ * @param {Object} [options] - an object with options.
+ * @param {String} [options.unit] - formats the distance with the given unit ('year', 'quarter', 'month', 'week', 'day', 'hour', 'minute', 'second').
+ * @param {String|String[]} [options.locale] - the locale to use.
+ * @param {String} [options.localeMatcher='best fit'] - the locale matching algorithm to use. Other value: 'lookup'.
+ * See MDN for details [Locale identification and negotiation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#locale_identification_and_negotiation)
+ * @param {String} [options.numeric='auto'] - the output message format. The values are 'auto' (e.g. `yesterday`), 'always'(e.g. `1 day ago`).
+ * @param {String} [options.style='long'] - the length of the result. The values are: 'long' (e.g. `1 month`), 'short' (e.g. 'in 1 mo.'), 'narrow' (e.g. 'in 1 mo.').
+ * The narrow one could be similar to the short one for some locales.
+ * @returns {String} the distance in words according to language-sensitive relative time formatting.
+ * @throws {TypeError} 2 arguments required
+ * @throws {RangeError} `date` must not be Invalid Date
+ * @throws {RangeError} `baseDate` must not be Invalid Date
+ * @throws {RangeError} `options.unit` must not be invalid Unit
+ * @throws {RangeError} `options.locale` must not be invalid locale
+ * @throws {RangeError} `options.localeMatcher` must not be invalid localeMatcher
+ * @throws {RangeError} `options.numeric` must not be invalid numeric
+ * @throws {RangeError} `options.style` must not be invalid style
+ *
+ * @example
+ * // What is the distance between the dates when the fist date is after the second?
+ * intlFormatDistance(
+ *   new Date(1986, 3, 4, 11, 30, 0),
+ *   new Date(1986, 3, 4, 10, 30, 0)
+ * )
+ * //=> 'in 1 hour'
+ *
+ * // What is the distance between the dates when the fist date is before the second?
+ * intlFormatDistance(
+ *   new Date(1986, 3, 4, 10, 30, 0),
+ *   new Date(1986, 3, 4, 11, 30, 0)
+ * )
+ * //=> '1 hour ago'
+ *
+ * @example
+ * // The option represents the unit which will be used for the result calculation.
+ * // What is the distance in quarters between the dates when the fist date is after the second?
+ * intlFormatDistance(
+ *   new Date(1987, 3, 4, 10, 30, 0),
+ *   new Date(1986, 3, 4, 10, 30, 0),
+ *   { unit: 'quarter' }
+ * )
+ * //=> 'in 4 quarters'
+ *
+ * @example
+ * // The options represent the unit and the locale which will be used for the result representation and calculation.
+ * // What is the distance in Spanish in minutes between the dates when the fist date is after the second?
+ * intlFormatDistance(
+ *   new Date(1986, 3, 4, 11, 30, 0),
+ *   new Date(1986, 3, 4, 10, 30, 0),
+ *   { unit: 'minute', locale: 'es' }
+ * )
+ * //=> 'dentro de 60 minutos'
+ *
+ * @example
+ * // The option represents the numeric format which will be used for the result representation.
+ * // What is the distance in numeric format between the dates when the fist date is after the second?
+ * intlFormatDistance(
+ *   new Date(1986, 3, 5, 11, 30, 0),
+ *   new Date(1986, 3, 4, 11, 30, 0),
+ *   { numeric: 'always' }
+ * )
+ * //=> 'in 1 day'
+ *
+ * @example
+ * // The option represents the style format which will be used for the result representation.
+ * // What is the distance in short format between the dates when the fist date is after the second?
+ * intlFormatDistance(
+ *   new Date(1988, 3, 4, 11, 30, 0),
+ *   new Date(1986, 3, 4, 11, 30, 0),
+ *   { style: 'short' }
+ * )
+ * //=> 'in 2 yr'
+ */
+
+export default function intlFormatDistance(
+  date: Date | number,
+  baseDate: Date | number,
+  options?: Options
+): string {
+  requiredArgs(2, arguments)
+
+  let value: number = 0
+  let unit: Intl.RelativeTimeFormatUnit
+  const dateLeft = toDate(date)
+  const dateRight = toDate(baseDate)
+
+  if (!options?.unit) {
+    // Get the unit based on diffInSeconds calculations if no unit is specified
+    const diffInSeconds = differenceInSeconds(dateLeft, dateRight) // The smallest unit
+
+    if (Math.abs(diffInSeconds) < secondsInMinute) {
+      value = differenceInSeconds(dateLeft, dateRight)
+      unit = 'second'
+    } else if (Math.abs(diffInSeconds) < secondsInHour) {
+      value = differenceInMinutes(dateLeft, dateRight)
+      unit = 'minute'
+    } else if (Math.abs(diffInSeconds) < secondsInDay) {
+      value = differenceInHours(dateLeft, dateRight)
+      unit = 'hour'
+    } else if (Math.abs(diffInSeconds) < secondsInWeek) {
+      value = differenceInCalendarDays(dateLeft, dateRight)
+      unit = 'day'
+    } else if (Math.abs(diffInSeconds) < secondsInMonth) {
+      value = differenceInCalendarWeeks(dateLeft, dateRight)
+      unit = 'week'
+    } else if (Math.abs(diffInSeconds) < secondsInQuarter) {
+      value = differenceInCalendarMonths(dateLeft, dateRight)
+      unit = 'month'
+    } else if (Math.abs(diffInSeconds) < secondsInYear) {
+      if (differenceInCalendarQuarters(dateLeft, dateRight) < 4) {
+        // To filter out cases that are less than a year but match 4 quarters
+        value = differenceInCalendarQuarters(dateLeft, dateRight)
+        unit = 'quarter'
+      } else {
+        value = differenceInCalendarYears(dateLeft, dateRight)
+        unit = 'year'
+      }
+    } else {
+      value = differenceInCalendarYears(dateLeft, dateRight)
+      unit = 'year'
+    }
+  } else {
+    // Get the value if unit is specified
+    unit = options?.unit
+    if (unit === 'second') {
+      value = differenceInSeconds(dateLeft, dateRight)
+    } else if (unit === 'minute') {
+      value = differenceInMinutes(dateLeft, dateRight)
+    } else if (unit === 'hour') {
+      value = differenceInHours(dateLeft, dateRight)
+    } else if (unit === 'day') {
+      value = differenceInCalendarDays(dateLeft, dateRight)
+    } else if (unit === 'week') {
+      value = differenceInCalendarWeeks(dateLeft, dateRight)
+    } else if (unit === 'month') {
+      value = differenceInCalendarMonths(dateLeft, dateRight)
+    } else if (unit === 'quarter') {
+      value = differenceInCalendarQuarters(dateLeft, dateRight)
+    } else if (unit === 'year') {
+      value = differenceInCalendarYears(dateLeft, dateRight)
+    }
+  }
+
+  const rtf = new Intl.RelativeTimeFormat(options?.locale, {
+    localeMatcher: options?.localeMatcher,
+    numeric: options?.numeric || 'auto',
+    style: options?.style,
+  })
+
+  return rtf.format(value, unit)
+}
--- a/src/intlFormatDistance/index.ts
+++ b/src/intlFormatDistance/index.ts
@@ -175,7 +175,10 @@ export default function intlFormatDistance(
     } else if (Math.abs(diffInSeconds) < secondsInHour) {
       value = differenceInMinutes(dateLeft, dateRight)
       unit = 'minute'
-    } else if (Math.abs(diffInSeconds) < secondsInDay) {
+    } else if (
+      Math.abs(diffInSeconds) < secondsInDay &&
+      Math.abs(differenceInCalendarDays(dateLeft, dateRight)) < 1
+    ) {
       value = differenceInHours(dateLeft, dateRight)
       unit = 'hour'
     } else if (
--- a/src/intlFormatDistance/index.ts
+++ b/src/intlFormatDistance/index.ts
@@ -180,7 +180,6 @@ export default function intlFormatDistance(
       (value = differenceInHours(dateLeft, dateRight)) &&
       Math.abs(value) < 24
     ) {
-      // value = differenceInHours(dateLeft, dateRight)
       unit = 'hour'
     } else if (
       Math.abs(diffInSeconds) < secondsInWeek &&
--- a/src/intlFormatDistance/index.ts
+++ b/src/intlFormatDistance/index.ts
@@ -177,16 +177,16 @@ export default function intlFormatDistance(
       unit = 'minute'
     } else if (
       Math.abs(diffInSeconds) < secondsInDay &&
-      Math.abs(differenceInCalendarDays(dateLeft, dateRight)) < 1
+      (value = differenceInHours(dateLeft, dateRight)) &&
+      Math.abs(value) < 24
     ) {
-      value = differenceInHours(dateLeft, dateRight)
+      // value = differenceInHours(dateLeft, dateRight)
       unit = 'hour'
     } else if (
       Math.abs(diffInSeconds) < secondsInWeek &&
       (value = differenceInCalendarDays(dateLeft, dateRight)) &&
       Math.abs(value) < 7
     ) {
-      value = differenceInCalendarDays(dateLeft, dateRight)
       unit = 'day'
     } else if (Math.abs(diffInSeconds) < secondsInMonth) {
       value = differenceInCalendarWeeks(dateLeft, dateRight)
--- a/src/intlFormatDistance/index.ts
+++ b/src/intlFormatDistance/index.ts
@@ -178,7 +178,11 @@ export default function intlFormatDistance(
     } else if (Math.abs(diffInSeconds) < secondsInDay) {
       value = differenceInHours(dateLeft, dateRight)
       unit = 'hour'
-    } else if (Math.abs(diffInSeconds) < secondsInWeek) {
+    } else if (
+      Math.abs(diffInSeconds) < secondsInWeek &&
+      (value = differenceInCalendarDays(dateLeft, dateRight)) &&
+      Math.abs(value) < 7
+    ) {
       value = differenceInCalendarDays(dateLeft, dateRight)
       unit = 'day'
     } else if (Math.abs(diffInSeconds) < secondsInMonth) {
--- a/src/intlFormatDistance/index.ts
+++ b/src/intlFormatDistance/index.ts
@@ -178,7 +178,7 @@ export default function intlFormatDistance(
     } else if (
       Math.abs(diffInSeconds) < secondsInDay &&
       (value = differenceInHours(dateLeft, dateRight)) &&
-      Math.abs(value) < 24
+      Math.abs(differenceInCalendarDays(dateLeft, dateRight)) < 1
     ) {
       unit = 'hour'
     } else if (
--- a/src/intlFormatDistance/index.ts
+++ b/src/intlFormatDistance/index.ts
@@ -1,4 +1,4 @@
-import { IntlOptionsUnit } from 'src/types'
+import type { IntlOptionsUnit } from '../types'
 import {
   secondsInDay,
   secondsInHour,
--- a/src/intlFormatDistance/index.ts
+++ b/src/intlFormatDistance/index.ts
@@ -7,7 +7,7 @@ import {
   secondsInQuarter,
   secondsInWeek,
   secondsInYear,
-} from '../constants'
+} from '../constants/index'
 import differenceInCalendarDays from '../differenceInCalendarDays/index'
 import differenceInCalendarMonths from '../differenceInCalendarMonths/index'
 import differenceInCalendarQuarters from '../differenceInCalendarQuarters/index'
--- a/src/intlFormatDistance/index.ts
+++ b/src/intlFormatDistance/index.ts
@@ -105,7 +105,6 @@ export interface IntlFormatDistanceOptions {
  *
  * @example
  * // The options represent the unit and the locale which will be used for the result representation and calculation.
- * // What is the distance in Spanish in minutes between the dates when the fist date is after the second?
  * intlFormatDistance(
  *   new Date(1986, 3, 4, 11, 30, 0),
  *   new Date(1986, 3, 4, 10, 30, 0),
--- a/src/intlFormatDistance/index.ts
+++ b/src/intlFormatDistance/index.ts
@@ -57,7 +57,6 @@ export interface IntlFormatDistanceOptions {
  * | 1 year                 | last year      | next year       |
  * | 2+ years               | X years ago    | in X years      |
  *
- *
  * @param {Date|Number} date - the date
  * @param {Date|Number} baseDate - the date to compare with.
  * @param {Object} [options] - an object with options.
--- a/src/intlFormatDistance/index.ts
+++ b/src/intlFormatDistance/index.ts
@@ -114,7 +114,7 @@ export interface IntlFormatDistanceOptions {
  * //=> 'dentro de 60 minutos'
  *
  * @example
- * // The option represents the numeric format which will be used for the result representation.
+ * // Use the numeric option to force the function to use numeric values. Without setting it, the example would return "tomorrow". 
  * // What is the distance in numeric format between the dates when the fist date is after the second?
  * intlFormatDistance(
  *   new Date(1986, 3, 5, 11, 30, 0),
--- a/src/intlFormatDistance/index.ts
+++ b/src/intlFormatDistance/index.ts
@@ -115,7 +115,6 @@ export interface IntlFormatDistanceOptions {
  *
  * @example
  * // Use the numeric option to force the function to use numeric values. Without setting it, the example would return "tomorrow". 
- * // What is the distance in numeric format between the dates when the fist date is after the second?
  * intlFormatDistance(
  *   new Date(1986, 3, 5, 11, 30, 0),
  *   new Date(1986, 3, 4, 11, 30, 0),
--- a/src/intlFormatDistance/index.ts
+++ b/src/intlFormatDistance/index.ts
@@ -94,7 +94,7 @@ export interface IntlFormatDistanceOptions {
  * //=> '1 hour ago'
  *
  * @example
- * // The option represents the unit which will be used for the result calculation.
+ * // Use the unit option to force the function to output the result in quarters. Without setting it, the example would return "next year"
  * intlFormatDistance(
  *   new Date(1987, 6, 4, 10, 30, 0),
  *   new Date(1986, 3, 4, 10, 30, 0),
@@ -103,8 +103,7 @@ export interface IntlFormatDistanceOptions {
  * //=> 'in 5 quarters'
  *
  * @example
- * // The option represents the locale which will be used for the result representation and calculation.
- * // What is the distance in Spanish between the dates when the fist date is after the second?
+ * // Use the locale option to get the result in Spanish. Without setting it, the example would return "in 1 hour".
  * intlFormatDistance(
  *   new Date(1986, 3, 4, 11, 30, 0),
  *   new Date(1986, 3, 4, 10, 30, 0),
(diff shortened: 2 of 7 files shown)


src/intlFormatRelative/index.js
```diff
@@ -0,0 +1,77 @@
...
+ * Default is: 
+ * {
+    localeMatcher: "best fit", // other values: "lookup"
+    numeric: "always", // other values: "auto"
+    style: "long", // other values: "short" or "narrow"
+   }
+ *  Default options :  { style: 'narrow' }
+ */
+
+export default function intlFormatRelative(date, unit, locale, options) {
```
kossnocorp, line 33: Let's make `locale` an options field.

src/intlFormatRelative/index.js
```diff
@@ -0,0 +1,77 @@
...
+    value = differenceInMonths(new Date(date), new Date())
+  } else if (unit === units[6] || unit === units[7]) {
+    value = differenceInWeeks(new Date(date), new Date())
+  } else if (unit === units[8] || unit === units[9]) {
+    value = differenceInDays(new Date(date), new Date())
+  } else if (unit === units[10] || unit === units[11]) {
+    value = differenceInHours(new Date(date), new Date())
+  } else if (unit === units[12] || unit === units[13]) {
+    value = differenceInMinutes(new Date(date), new Date())
+  } else if (unit === units[14] || unit === units[15]) {
```
kossnocorp, line 73: These conditions are hard to read. Let's simplify and use a single unit. Right now, we have `formatDistanceStrict` that uses 'second', let's stick to it.

Also, since we don't need to check unit argument for validity (see my comments above and below), we can just write 

```
} else if (unit === 'second') {
```

src/intlFormatRelative/index.js
```diff
@@ -0,0 +1,77 @@
...
+    value = differenceInSeconds(new Date(date), new Date())
+  }
+  return rtf.format(value, unit)
```
kossnocorp, line 76: Regarding what to do when `unit` argument is not valid.

If unit is undefined, JS throws this error:

```js
(new Intl.RelativeTimeFormat()).format(0)
//=> VM267:1 Uncaught RangeError: Invalid unit argument for Intl.RelativeTimeFormat.prototype.format() 'undefined'
//=>    at RelativeTimeFormat.format (<anonymous>)
//=>    at <anonymous>:1:33(new Intl.RelativeTimeFormat()).format(0)
```

It's not super clear but not terrible either. Our stance - users should use type systems (TypeScript or Flow) for type checks. Our focus is small build size, and excluding type checks code is an easy way to ensure it.

I hope you get my point. You can ping me for more details.

src/intlFormatRelative/test.js
```diff
@@ -0,0 +1,63 @@
...
+/* eslint-env mocha */
+
+import assert from 'power-assert'
+import intlFormatRelative from '.'
+import subDays from '../subDays/index'
+import addHours from '../addHours/index'
+import addMinutes from '../addMinutes/index'
+import addSeconds from '../addSeconds/index'
+
+it('should return 1 year ago', function () {
```
kossnocorp, line 11: Please wrap all tests in `describe('intlFormatDistance', () => {`

src/intlFormatRelative/test.js
```diff
@@ -0,0 +1,63 @@
...
+
+it('should return 1 minute ago', function () {
+  var result = intlFormatRelative(addMinutes(new Date(), -1), 'minute')
+  assert(result === '1 minute ago')
+})
+
+it('should return 1 second ago', function () {
+  var result = intlFormatRelative(addSeconds(new Date(), -1), 'second')
+  assert(result === '1 second ago')
+})
```
kossnocorp, line 49: After adding the second argument, please add cases for `in 1 second`, `in 1 week`, etc.

kossnocorp (changes requested):
Great job, code and tests look good.

Besides review comments I have few requests:

- Please rewrite it to TypeScript. If you never worked with it or just need advice - ping me, I'll help you.

- Let's change the behavior. You can see that our `formatDistance` picks the most appropriate unit depending on the distance. The less the distance the smaller the unit. Let emulate this algo but allow users pass `unit` as well. So that we can move `unit` to options and make it optional.
  
  Here's the algo I suggest (tell me if you have other ideas)

  - Less than minites: `second`
  - Less than hour: `minute`
  - Less than day: `hour`
  - Less than week: `day`
  - Less than quarter: `month`
  - Less than year: `quarter`
  - More or equal to year: `year`

- Let's rename the function to `intlFormatDistance` and make API similar to https://date-fns.org/v2.16.1/docs/formatDistance. Initially, I wanted to keep it similar to Intl API, but realized that it will confuse **our** users.

- Add a second date argument like in `formatDistance` and use it in calculations instead of `new Date()`.

-  Add another function variation `intlFormatDistanceToNow` similar to https://date-fns.org/v2.16.1/docs/formatDistanceToNow that uses Date.now() as the second argument.

src/intlFormatRelative/index.js
```diff
@@ -0,0 +1,77 @@
...
+ * @param {Object} [options={}] - optional object with the following properties: 
+ * localeMatcher, numeric, style. 
+ * Default is: 
+ * {
+    localeMatcher: "best fit", // other values: "lookup"
+    numeric: "always", // other values: "auto"
+    style: "long", // other values: "short" or "narrow"
+   }
+ *  Default options :  { style: 'narrow' }
+ */
```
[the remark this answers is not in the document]
tan75: Addressed as below:
* @example
 * // What is the distance between Mar, 4 1986 11:30:00 and Mar, 4 1986 10:30:00 in Intl?
 * const result = intlFormatDistance(
 *   new Date(1986, 3, 4, 11, 30, 0),
 *   new Date(1986, 3, 4, 10, 30, 0));
 * // => in 1 hour
 *

src/intlFormatRelative/index.js
```diff
@@ -0,0 +1,77 @@
...
+ * @summary Enables language-sensitive relative time formatting according to
+ * according to the locale and formatting options of the given Intl.RelativeTimeFormat object.
+ * @description 
+ * Get difference between today and the given date 
+ * which will become a value for Intl.RelativeTimeFormat.prototype.format()
+ * @param {Date|Number} date - the given date
+ * @param {String} unit - unit values "year", "quarter", "month", "week", 
+ * "day", "hour", "minute", "second" as per MDN Web Docs
+ * @param {String} [locale='en'] - optional locale, default value is 'en'. 
+ * @param {Object} [options={}] - optional object with the following properties:
```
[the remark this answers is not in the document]
tan75: Addressed as below: 

`@param {String|String[]} [rtfOptions.locales='en'] - a string (or array of strings) with a BCP 47 language tag. `

src/intlFormatRelative/index.js
```diff
@@ -0,0 +1,77 @@
...
+    'minute',
+    'minutes',
+    'second',
+    'seconds',
+  ]
+  const rtf = new Intl.RelativeTimeFormat(locale, options)
+  let value = 0
+
+  if (units.indexOf(unit) === -1) {
+    throw new Error('Invalid unit supplied')
```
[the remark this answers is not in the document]
tan75: Addressed.

src/intlFormatRelative/index.js
tan75: Addressed.

src/intlFormatRelative/index.js
tan75: Addressed.

src/intlFormatRelative/index.js
tan75: Addressed.

src/intlFormatRelative/test.js
tan75: Addressed.

src/intlFormatRelative/test.js
tan75: Addressed.

tan75:
WIP:
* intlFormatDistanceToNow
* Test Cases

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,149 @@
...
+import differenceInQuarters from '../differenceInQuarters/index';
+import differenceInYears from '../differenceInYears/index';
+
+const SECONDS_IN_MINUTE = 60;
+const SECONDS_IN_HOUR = 3600;
+const SECONDS_IN_DAY = 86400;
+const SECONDS_IN_WEEK = 604800;
+const SECONDS_IN_MONTH = 2592000;  // 30 days in month
+const SECONDS_IN_3_MONTHS = 7776000;  // 3 months in quarter
+const SECONDS_IN_YEAR = 31104000; // 12 months in year
```
kossnocorp, line 18: Regrading these. Let use the same approach I used in the milliseconds function: https://github.com/date-fns/date-fns/blob/master/src/milliseconds/index.ts#L6

You can move the constant to the shared module and reuse it from there: https://github.com/date-fns/date-fns/blob/master/src/constants/index.js

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,149 @@
...
+ * depending on the distance (the less the distance the smaller the unit),
+ * or allowes a user to pass in a unit as well.
+ *
+ * | Distance between the below dates, no unit passed in:         | Result        |
+ * | new Date(1986, 3, 4, 11, 30, 0),                             | in 1 hour     |
+ * | new Date(1986, 3, 4, 10, 30, 0)                              |               |
+*  |--------------------------------------------------------------|---------------|
+ * | Distance between the below dates, hour passed in as a unit:  | Result        |
+ * | new Date(1986, 3, 4, 11, 30, 0),                             | in 60 minutes |
+ * | new Date(1986, 3, 4, 10, 30, 0)                              |               |
```
kossnocorp, line 36: Let's repurpose this table instead: https://github.com/date-fns/date-fns/blob/master/src/formatDistance/index.js#L23-L40

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,149 @@
...
+ *   new Date(1986, 3, 4, 10, 30, 0),
+ *   { unit: 'minute'});
+ * // => in 60 minutes
+ *
+ * @example
+ * // What is the distance between Mar, 4 1986 11:30:00 and Mar, 4 1986 10:30:00 in minutes in Spanish in Intl?
+ * const result = intlFormatDistance(
+ *   new Date(1986, 3, 4, 11, 30, 0),
+ *   new Date(1986, 3, 4, 10, 30, 0),
+ *   { unit: 'minute', locales: 'es' });
```
kossnocorp, line 70: Please format examples (and code btw) using Prettier. If you use VS Code, I would recommend making it to format the code on save.

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,149 @@
...
+ * | new Date(1986, 3, 4, 10, 30, 0)                              |               |
+*  |--------------------------------------------------------------|---------------|
+ * | Distance between the below dates, hour passed in as a unit:  | Result        |
+ * | new Date(1986, 3, 4, 11, 30, 0),                             | in 60 minutes |
+ * | new Date(1986, 3, 4, 10, 30, 0)                              |               |
+
+ * @param {Date|Number} date - the date
+ * @param {Date|Number} baseDate - the date to compare with.
+ * @param {Object} [rtfOptions] - an object with options.
+ * @param {String} [rtfOptions.unit=''] - values "year", "quarter", "month", "week", "day", "hour", "minute", "second" as per MDN Web Docs.
```
kossnocorp, line 41:
```suggestion
 * @param {String} [rtfOptions.unit] - formats the distance with the given unit ("year", "quarter", "month", "week", "day", "hour", "minute", "second").
```

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,149 @@
...
+ * @param {String|String[]} [rtfOptions.locales='en'] - a string (or array of strings) with a BCP 47 language tag. For reference see:
+ * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#locale_identification_and_negotiation
```
kossnocorp, line 43:
```suggestion
 * @param {String|String[]} [rtfOptions.locales] - the locale to use (BCP 47 language tag). [For the reference see MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#locale_identification_and_negotiation).
```

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,149 @@
...
+ * @param {String} [rtfOptions.options.localeMatcher='best fit'] - the locale matching algorithm to use. Other value: "lookup".
+ * @param {String} [rtfOptions.options.numeric='always'] - the output message format. Other value: "auto".
+ * @param {String} [rtfOptions.options.style='long'] - the length of the internationalized message. Other values: "short" or "narrow";
```
kossnocorp, line 46: Let's make the options object flat and move all properties to rtfOptions (and rename it simply options).

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,149 @@
...
+ * For reference see:
+ * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat/RelativeTimeFormat
```
kossnocorp, line 48: Instead of forcing users to read MDN, we can add either example for each option variant or a table to the description.

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,149 @@
...
+  localeMatcher?: String,
+  numeric?: String,
+  style?: String,
+};
+
+type Options = {
+  unit?: String,
+  locales?: String;
+  options?: IrtfOptions;
+}
```
kossnocorp, line 85: Please upgrade typescript to the latest (4.1.3) and add `ESNext.Intl` to https://github.com/date-fns/date-fns/blob/master/tsconfig.json#L4

```suggestion
interface Options {
  localeMatcher?: Intl.RelativeTimeFormatLocaleMatcher,
  numeric?: Intl.RelativeTimeFormatNumeric,
  style?: Intl.RelativeTimeFormatStyle,
  unit?: Unit,
  locale?: Intl.RelativeTimeFormatStyle
}
```

Note that I renamed `locales` -> `locale` as I think a single locale will be used the most, so better make it look good for the default case.

Then add `Unit` to https://github.com/date-fns/date-fns/blob/master/src/types.ts:

```ts
export type Unit =
  | 'year'
  | 'quarter'
  | 'month'
  | 'week'
  | 'day'
  | 'hour'
  | 'minute'
  | 'second'
```

There's `RelativeTimeFormatUnit` but I prefer limit options (it's easy to extend than limit later) and use the same type in other functions.

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,149 @@
...
+    } else if (unit === 'quarter') {
+      value = differenceInQuarters(date, baseDate);
+    } else if (unit === 'year') {
+      value = differenceInYears(date, baseDate);
+    }
+  }
+
+  const rtf = new Intl.RelativeTimeFormat(
+    rtfOptions?.locales || defaultLocale,
+    rtfOptions?.options || {}
```
kossnocorp, line 146: Now I understand why you did make nested `options`, but let's instead explicitly specify the options here.

src/intlFormatDistance/test.ts
```diff
@@ -0,0 +1,216 @@
...
+  });
+
+  it('prints out in 60 minutes', function () {
+    const result = intlFormatDistance(new Date(1986, 3, 4, 11, 30, 0),
+      new Date(1986, 3, 4, 10, 30, 0), {unit: 'minute'});
+    assert(result === 'in 60 minutes');
+  });
+});
+
+describe('intlFormatDistance edge cases', () => {
```
kossnocorp, line 177: Add this as a nested descript to `describe('intlFormatDistance',`.  We follow a rule: "one root-level `describe` per file".

kossnocorp (changes requested):
Good progress. Please format the code using Prettier and address my suggestions.

src/intlFormatDistance/index.ts
tan75: I will still have to modify the table because we've got units passed in as well. E.g. a value of 60 seconds can be represented by either **'in 60 seconds'** or **'in 1 minute'** result
Do I need to create examples that will satisfy every unit?

src/intlFormatDistance/index.ts
tan75: Done

src/intlFormatDistance/index.ts
tan75: Done as below:
```
type Options = {
  unit?: String
  locales?: String
  localeMatcher?: String
  numeric?: String
  style?: String
}

export default function intlFormatDistance(
  date: Date | number,
  baseDate: Date | number,
  options?: Options
)
```

src/intlFormatDistance/index.ts
tan75: Will this suffice?
```
  const rtf = new Intl.RelativeTimeFormat(
    options?.locale || defaultLocale,
    {
      localeMatcher: options?.localeMatcher,
      numeric: options?.numeric,
      style: options?.style,
    } || {}
  )

```

src/intlFormatDistance/test.ts
tan75: Done

src/intlFormatDistance/index.ts
tan75: @kossnocorp please check my approach in my latest commit and let me know

src/intlFormatDistance/index.ts
tan75: @kossnocorp 
WIP

src/intlFormatDistance/index.ts
tan75: done

src/intlFormatDistance/index.ts
tan75: WIP

src/intlFormatDistance/index.ts
tan75: @kossnocorp 
the only thing i decided to change was the order or the properties (unit goes first) to follow the order in the MDN documentation (if thats ok). I think this makes more sense :

```
interface Options {
  unit?: Unit
  locale?: Intl.RelativeTimeFormatStyle
  localeMatcher?: Intl.RelativeTimeFormatLocaleMatcher
  numeric?: Intl.RelativeTimeFormatNumeric
  style?: Intl.RelativeTimeFormatStyle
}


```

src/constants/index.ts
```diff
@@ -0,0 +1,11 @@
...
+ * Days In Year
+ * One years equals 365.2425 days according to the formula:
+ *
+ * > Leap year occures every 4 years, except for years that are divisable by 100 and not divisable by 400.
+ * > 1 mean year = (365+1/4-1/100+1/400) days = 365.2425 days
+ *  @constant
+ *  @type {number}
+ *  @default
+ */
+export let yearInDays: number = 365.2425
```
leshakoss, line 11: Shouldn't it be the other way around, `daysInYear`, as in [number of] days in [a] year?

tan75: > Shouldn't it be the other way around, `daysInYear`, as in [number of] days in [a] year?

it totally makes sense but i took it from here:
https://github.com/date-fns/date-fns/blob/master/src/milliseconds/index.ts#L6

tan75: > Shouldn't it be the other way around, `daysInYear`, as in [number of] days in [a] year?

yeah, i think I will change it - your way is more explicit

kossnocorp: 👍

robertvanhoesel:
Awesome work! While working on my own implementation I found a a gotcha with `numeric: 'auto'` which might be wise to address (or at least mention in documentation). Because `RelativeTimeFormat` only takes a value and unit, it has no context/awareness of years. 

When I compare two dates, today (Feb 24, 2020) and almost 2 years ago (March 30, 2019).`dateFns.differenceInYears` will return '-1', which is correct, it is not 2 years ago, just one and a bit. 

```ts
new Intl.RelativeTimeFormat('en', { 
  numeric: "auto" 
}).format(-1, 'year')
// 'last year'
```

The majority of Intl implementations will return `last year` in response to above code, which to humans is just incorrect when we are aware of comparing 2021 and 2019, which was not 'last year'. 

https://codesandbox.io/s/musing-river-90m6t?file=/src/index.ts

Since your proposed implementation always takes two dates, you could prevent this error from occurring by having additional checks in the [unit resolver](https://github.com/date-fns/date-fns/pull/2173/files#diff-85e3d08d9b516732ef39e93d178a81815de328e6929a6e218282166c1ab42556R237-R249) that only applies 'auto' when the two compared years differ exactly 1 year, and force 'numeric' in all other cases. 

@kossnocorp It occurs to me that maybe this more opinionated/smart logic should be in `Intl.RelativeTimeFormat` instead? Alongside rounding instead of flooring (i.e. 1 year 9 months becomes 2 years) and optional joining in localized time?

robertvanhoesel:
Or... split this implementation to:
- `intlFormatDistanceStrict` which uses `numeric: always` and takes 'ceil', 'round', 'floor' arguments.
-  `intlFormatDistance` which uses `numeric: auto` where applicable and generally rounds the years.

My 2cts would be to avoid providing the exact arguments of `Intl.RelativeTimeFormat` and allow room to algorithmically use the right unit and formatting options.

An other thing to review would be wether outputting 'quarters' by default makes sense to most implementations. Date-fns is slightly opinionated in its other implementations (i.e. when to include time in 'relative' and never returning '4 quarters ago' in formatDistance), providing that as an option to control would make this more useful, otherwise I would find myself going back to first comparing the dates, then calling intlFormatDistance with unit months when I know we're talking about 3-12 months distance.

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,282 @@
...
+
+export default function intlFormatDistance(
+  date: Date | number,
+  baseDate: Date | number,
+  options?: Options
+): String {
+  requiredArgs(2, arguments)
+
+  let value
+  let unit
```
robertvanhoesel, line 211:
```suggestion
  let value: number
  let unit: Intl.RelativeTimeFormatUnit
```

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,282 @@
...
+ *   new Date(1986, 3, 4, 10, 30, 0),
+ *   { unit: 'minute', locales: 'es' });
+ * // => dentro de 60 minutos
+ *
+ * @example
+ * // What is the distance between Apr, 4 1986 11:30:00 and Apr, 4 1986 10:30:00 in minutes in German in Intl?
+ * const result = intlFormatDistance(
+ *   new Date(1986, 3, 4, 11, 30, 0),
+ *   new Date(1986, 3, 4, 10, 30, 0),
+ *   { unit: 'minute', locales: 'de' });
```
kossnocorp, line 190:
```suggestion
 *   { unit: 'minute', locale: 'de' });
```

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,282 @@
...
+ *   new Date(1986, 3, 4, 10, 30, 0),
+ *   { unit: 'second'});
+ * // in 31,536,000 seconds
+ *
+ * @example
+ * // What is the distance between Apr, 4 1986 11:30:00 and Apr, 4 1986 10:30:00 in minutes in Spanish in Intl?
+ * const result = intlFormatDistance(
+ *   new Date(1986, 3, 4, 11, 30, 0),
+ *   new Date(1986, 3, 4, 10, 30, 0),
+ *   { unit: 'minute', locales: 'es' });
```
kossnocorp, line 182:
```suggestion
 *   { unit: 'minute', locale: 'es' });
```

src/intlFormatDistance/index.ts
kossnocorp: I can see that examples are still not formatted, for instance here the final `)` must be on a new page and the semicolon should be committed. Please check other examples as well.

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,282 @@
...
+ * @throws {RangeError} `options.localeMatcher=` must not be invalid localeMatcher
+ * @throws {RangeError} `options.numeric=` must not be invalid numeric
+ * @throws {RangeError} `options.style=` must not be invalid style
+ *
+ * @example
+ * // What is the distance between Apr, 4 1986 11:30:00 and Apr, 4 1986 10:30:00 in Intl?
+ * const result = intlFormatDistance(
+ *   new Date(1986, 3, 4, 11, 30, 0),
+ *   new Date(1986, 3, 4, 10, 30, 0));
+ * // => in 1 hour
```
kossnocorp, line 112: Please here and below 1) make output consistent (always has `=>`, no space between it and the comment) and 2) wrap strings into quotes.

```suggestion
 * //=> 'in 1 hour'
```

kossnocorp (changes requested):
Some more new comments

kossnocorp:
@robertvanhoesel, this is an excellent point. Adding special logic for `numeric: 'auto'` and `-1`/`1`/`0` makes sense at first glance, not only for `year` but also `month` and `quarter`. However, I'm afraid it's a slippery slope. For instance, this is not a problem for `minute` and `hour` in English, but I imagine it might be a case for other languages or Intl implementations or wise versa might not have an issue with `year`.

Using `ceil`/`floor` won't help because rounding `4 months` to `0 years` while working with December will produce `this year` and its inaccuracy.

I think instead we can use `differenceInCalendarDays`, `differenceInCalendarWeeks`, `differenceInCalendarMonths` and `differenceInCalendarYears` **unless** `numeric` is set to `always`. Then with your example, you'll get `-2 "2 years ago"`.

We also should make `numeric: 'auto'` the default value because such:

```ts
new Intl.RelativeTimeFormat('en').format(0, 'year')
//=> "in 0 years"
```

..it is not a good default option. However, I'm not sure what are side effects of this change. @tan75 could you please explore that?

@robertvanhoesel @tan75 @leshakoss @rikkalo WDYT?

tan75:
@kossnocorp 
If `numeric: auto` is a default option, then when adding `style: short` it falls back to `numeric: always` (if the result is 1 day or 1 month or 1 year) e.g.:
```
const result = intlFormatDistance(
      new Date(1985, 4, 5, 10, 30, 0),
      new Date(1985, 4, 4, 10, 30, 0),
      { style: 'short' }
) //=> 'in 1 day' instead of 'tomorrow'
```

however it looks like explicitly adding `numeric: auto`  when calling the function will fix it (even though we know that's the case by default):

```
    const result = intlFormatDistance(
      new Date(1985, 4, 5, 10, 30, 0),
      new Date(1985, 4, 4, 10, 30, 0),
      { style: 'short', numeric: 'auto' }
   )//=> 'tomorrow'
```

The same happens to 1 month and 1 year

tan75:
hi @robertvanhoesel 
my tests are failing on this [Error](https://github.com/date-fns/date-fns/pull/2173/checks?check_run_id=2371282512#step:8:2108)

image

Any ideas why?

robertvanhoesel:
@tan75 the typescript compiler needs to know that you are using es2020.intl definitions. In .tsconfig you should add es2020 to the target. Also, this is a good case to catch, as older browser will also not know about this constructor and fail.

robertvanhoesel:
I use the below logic to check wether the API is available in the browser. 
```
const IntlRelativeSupported = typeof Intl.RelativeTimeFormat === 'function';
```

tan75:
> @tan75 the typescript compiler needs to know that you are using es2020.intl definitions. In .tsconfig you should add es2020 to the target. Also, this is a good case to catch, as older browser will also not know about this constructor and fail.

Thank a lot for your answer! Do you mean tsconfig.json? I added it here 
[tsconfig](https://github.com/date-fns/date-fns/pull/2173/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0R3-R4)
But is it still failing:
[Run](https://github.com/date-fns/date-fns/pull/2173/checks?check_run_id=2373675432#step:8:1950)

leshakoss (approved):
⭐️

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,287 @@
...
+  date: Date | number,
+  baseDate: Date | number,
+  options: Options = { numeric: 'auto' }
+): String {
+  requiredArgs(2, arguments)
+
+  let value: number = 0
+  let unit: Intl.RelativeTimeFormatUnit
+  let dateLeft = toDate(date)
+  let dateRight = toDate(baseDate)
```
kossnocorp, line 218:
```suggestion
  const dateLeft = toDate(date)
  const dateRight = toDate(baseDate)
```

kossnocorp (changes requested):

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,283 @@
...
+ * @throws {RangeError} `date` must not be Invalid Date
+ * @throws {RangeError} `baseDate` must not be Invalid Date
+ * @throws {RangeError} `options.unit` must not be invalid Unit
+ * @throws {RangeError} `options.locale` must not be invalid locale
+ * @throws {RangeError} `options.localeMatcher` must not be invalid localeMatcher
+ * @throws {RangeError} `options.numeric` must not be invalid numeric
+ * @throws {RangeError} `options.style` must not be invalid style
+ *
+ * @example
+ * // What is the distance between Apr, 4 1986 11:30:00 and Apr, 4 1986 10:30:00 in Intl?
```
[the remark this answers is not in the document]
kossnocorp: Please make sure that the examples say exactly what you trying to show instead of repeating what it does: https://www.loom.com/share/516194288d764b5b8fc233c5de47b10a

Also, please reduce the number of examples to the required minimum.

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,283 @@
...
+ * @category Common Helpers
+ * @summary Enables language-sensitive relative time formatting according to the locale and formatting options
+ * of the given Intl.RelativeTimeFormat object.
+ * @description
+ * The API gets a difference between two given dates and either picks the most appropriate unit
+ * depending on the distance (the less the distance the smaller the unit),
+ * or allows a user to pass in a unit as well.
+ * If a unit is passed in, it will be applied accordingly. Otherwise - see the table below:
+ *
+ * |    Distance between dates   |               Result               |
```
kossnocorp, line 32: Please rework the doc tables: https://www.loom.com/share/ede734c88e1c4943b095e6b17d387f83

kossnocorp (changes requested):
I believe that the implementation is perfect, but it's distracting because both documentation and tests aren't descriptive, and I have a hard time parsing it. I imagine users will have trouble too.

tan75:
Current test structure:
image

src/intlFormatDistance/index.ts
tan75: is this ok ? 
`e.g. "hi": Hindi (language) or "de-AT": German (language) as used in Austria (region)`

src/constants/index.ts
```diff
@@ -8,6 +8,18 @@
...
+ * Days in 1 year
+ * One years equals 365.2425 days according to the formula:
+ *
+ * > Leap year occures every 4 years, except for years that are divisable by 100 and not divisable by 400.
+ * > 1 mean year = (365+1/4-1/100+1/400) days = 365.2425 days
+ *  @constant
+ *  @type {number}
+ *  @default
+ */
+export let daysInYear: number = 365.2425
```
kossnocorp, line 21:
```suggestion
export const daysInYear: number = 365.2425
```

src/constants/index.ts
```diff
@@ -8,6 +8,18 @@
  */
 export const daysInWeek = 7
 
+/**
+ * Days in 1 year
+ * One years equals 365.2425 days according to the formula:
+ *
+ * > Leap year occures every 4 years, except for years that are divisable by 100 and not divisable by 400.
+ * > 1 mean year = (365+1/4-1/100+1/400) days = 365.2425 days
+ *  @constant
```
kossnocorp, line 17:
```suggestion
 *
 * @name secondsInDay
 * @constant
```

src/intlFormatDistance/index.ts
kossnocorp: I meant to change the table to express what they get:

```
0 seconds | now
1-59 seconds | in X seconds
1-59 minutes | in X minutes
```

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,284 @@
...
+}
+
+/**
+ * @name intlFormatDistance
+ * @category Common Helpers
+ * @summary Formats distance between two dates in human-readable format
+ * @description
+ * The function calculates the difference between two given dates and either picks the most appropriate unit
+ * depending on the distance (the less the distance the smaller the unit),
+ * or allows a user to pass in a unit as well.
```
kossnocorp, line 37:
```suggestion
 * or allows the user to specify the unit as well.
```

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,284 @@
...
+ * If a unit is passed in, it will be applied accordingly. Otherwise - see the table below:
```
kossnocorp, line 38:
```suggestion
 * If the unit is specified, it will be applied accordingly. Otherwise - see the table below:
```

src/intlFormatDistance/test.ts
```diff
@@ -0,0 +1,647 @@
...
+          { unit: 'second' }
+        )
+        assert(result === 'in 31,536,000 seconds')
+      })
+
+      it('works with seconds in past', () => {
+        const result = intlFormatDistance(
+          new Date(1986, 3, 4, 10, 30, 0),
+          new Date(1987, 3, 4, 10, 30, 0),
+
```
kossnocorp, line 178:
```suggestion
```

src/intlFormatDistance/test.ts
```diff
@@ -0,0 +1,647 @@
...
+          null,
+          new Date(1986, 3, 4, 10, 30, 0),
+          new Date(1986, 3, 4, 10, 30, 0),
+          { localeMatcher: 'wrongValue' }
+        ),
+        RangeError
+      )
+    })
+
+    it('checks numerick', () => {
```
kossnocorp, line 621:
```suggestion
    it('checks numeric', () => {
```

kossnocorp (approved):

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,238 @@
+import { IntlOptionsUnit } from 'src/types'
+import {
+  secondsInDay,
+  secondsInHour,
+  secondsInMinute,
+  secondsInMonth,
+  secondsInQuarter,
+  secondsInWeek,
+  secondsInYear,
+} from '../constants'
```
kossnocorp, line 10:
```suggestion
} from '../constants/index'
```

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,238 @@
+import { IntlOptionsUnit } from 'src/types'
```
kossnocorp, line 1:
```suggestion
import type { IntlOptionsUnit } from '../types'
```

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,238 @@
...
+import differenceInCalendarQuarters from '../differenceInCalendarQuarters/index'
+import differenceInCalendarWeeks from '../differenceInCalendarWeeks/index'
+import differenceInCalendarYears from '../differenceInCalendarYears/index'
+import differenceInHours from '../differenceInHours/index'
+import differenceInMinutes from '../differenceInMinutes/index'
+import differenceInSeconds from '../differenceInSeconds/index'
+import toDate from '../toDate/index'
+import requiredArgs from '../_lib/requiredArgs/index'
+
+interface Options {
```
kossnocorp, line 22:
```suggestion
export interface IntlFormatDistanceOptions {
```

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,238 @@
...
+}
+
+/**
+ * @name intlFormatDistance
+ * @category Common Helpers
+ * @summary Formats distance between two dates in a human-readable format
+ * @description
+ * The function calculates the difference between two dates and either picks the most appropriate unit
+ * depending on the distance (the less the distance the smaller the unit) or allows the user to specify a unit.
+ * If the unit is specified, it will be applied accordingly. Otherwise - see the table below:
```
kossnocorp, line 37:
```suggestion
 * The function calculates the difference between two dates and formats it as a human-readable string.
 *
 * The function will pick the most appropriate unit depending on the distance between dates. For example, if the distance is a few hours, it might return `x hours`. If the distance is a few months, it might return `x months`.
 *
 * You can also specify a unit to force using it regardless of the distance to get a result like `123456 hours`.
 *
 * See the table below for the unit picking logic:
```

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,238 @@
...
+ * |  X months                   | in X months            |
+ * |  X months ago               | X months ago           |
+ * |  1 quarter                  | next quarter           |
+ * |  1 quarter                  | last quarter           |
+ * |  X quarters                 | in X quarters          |
+ * |  X quarters ago             | X quarters ago         |
+ * |  1 year                     | next year              |
+ * |  1 year ago                 | last year              |
+ * |  X years                    | in X years             |
+ * |  X years ago                | X years ago            |
```
kossnocorp, line 77:
```suggestion
 * | Distance between dates | Result (past)  | Result (future) |
 * | ---------------------- | -------------- | --------------- |
 * | 0 seconds              | now            | now             |
 * | 1-59 seconds           | X seconds ago  | in X seconds    |
 * | 1-59 minutes           | X minutes ago  | in X minutes    |
 * | 1-23 hours             | X hours ago    | in X hours      |
 * | 1 day                  | yesterday      | tomorrow        |
 * | 2-6 days               | X days ago     | in X days       |
 * | 7 days                 | last week      | next week       |
 * | 8 days-1 month         | X weeks ago    | in X weeks      |
 * | 1 month                | last month     | next month      |
 * | 2-3 months             | X months ago   | in X months     |
 * | 1 quarter              | last quarter   | next quarter    |
 * | 2-3 quarters           | X quarters ago | in X quarters   |
 * | 1 year                 | last year      | next year       |
 * | 2+ years               | X years ago    | in X years      |
```

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,238 @@
...
+ *
+ * @example
+ * // The option represents the unit which will be used for the result calculation.
+ * // What is the distance in quarters between the dates when the fist date is after the second?
+ * intlFormatDistance(
+ *   new Date(1987, 3, 4, 10, 30, 0),
+ *   new Date(1986, 3, 4, 10, 30, 0),
+ *   { unit: 'quarter' }
+ * )
+ * //=> 'in 4 quarters'
```
kossnocorp, line 123: Please make the result `5 quarters` to illustrate the point

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,238 @@
...
+ *
+ * // What is the distance between the dates when the fist date is before the second?
+ * intlFormatDistance(
+ *   new Date(1986, 3, 4, 10, 30, 0),
+ *   new Date(1986, 3, 4, 11, 30, 0)
+ * )
+ * //=> '1 hour ago'
+ *
+ * @example
+ * // The option represents the unit which will be used for the result calculation.
```
kossnocorp, line 116: Please put the option name upfront

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,238 @@
...
+ *
+ * @example
+ * // The options represent the unit and the locale which will be used for the result representation and calculation.
+ * // What is the distance in Spanish in minutes between the dates when the fist date is after the second?
+ * intlFormatDistance(
+ *   new Date(1986, 3, 4, 11, 30, 0),
+ *   new Date(1986, 3, 4, 10, 30, 0),
+ *   { unit: 'minute', locale: 'es' }
+ * )
+ * //=> 'dentro de 60 minutos'
```
kossnocorp, line 133: Please make the example locale focused to avoid confusion, and also don't forget to update the comment.

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,238 @@
...
+ * intlFormatDistance(
+ *   new Date(1987, 3, 4, 10, 30, 0),
+ *   new Date(1986, 3, 4, 10, 30, 0),
+ *   { unit: 'quarter' }
+ * )
+ * //=> 'in 4 quarters'
+ *
+ * @example
+ * // The options represent the unit and the locale which will be used for the result representation and calculation.
+ * // What is the distance in Spanish in minutes between the dates when the fist date is after the second?
```
kossnocorp, line 127:
```suggestion
```

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,238 @@
...
+ * // What is the distance between the dates when the fist date is before the second?
+ * intlFormatDistance(
+ *   new Date(1986, 3, 4, 10, 30, 0),
+ *   new Date(1986, 3, 4, 11, 30, 0)
+ * )
+ * //=> '1 hour ago'
+ *
+ * @example
+ * // The option represents the unit which will be used for the result calculation.
+ * // What is the distance in quarters between the dates when the fist date is after the second?
```
kossnocorp, line 117:
```suggestion
```

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,238 @@
...
+ * intlFormatDistance(
+ *   new Date(1986, 3, 4, 11, 30, 0),
+ *   new Date(1986, 3, 4, 10, 30, 0),
+ *   { unit: 'minute', locale: 'es' }
+ * )
+ * //=> 'dentro de 60 minutos'
+ *
+ * @example
+ * // The option represents the numeric format which will be used for the result representation.
+ * // What is the distance in numeric format between the dates when the fist date is after the second?
```
kossnocorp, line 137:
```suggestion
```

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,238 @@
...
+ * // What is the distance in Spanish in minutes between the dates when the fist date is after the second?
+ * intlFormatDistance(
+ *   new Date(1986, 3, 4, 11, 30, 0),
+ *   new Date(1986, 3, 4, 10, 30, 0),
+ *   { unit: 'minute', locale: 'es' }
+ * )
+ * //=> 'dentro de 60 minutos'
+ *
+ * @example
+ * // The option represents the numeric format which will be used for the result representation.
```
kossnocorp, line 136: Please use this example to update the option examples comment above and below:

```suggestion
 * // Use the numeric option to force the function to use numeric values. Without setting it, the example would return "tomorrow". 
```

kossnocorp (changes requested):

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,215 @@
...
+ * | 2-6 days               | X days ago     | in X days       |
+ * | 7 days                 | last week      | next week       |
+ * | 8 days-1 month         | X weeks ago    | in X weeks      |
+ * | 1 month                | last month     | next month      |
+ * | 2-3 months             | X months ago   | in X months     |
+ * | 1 quarter              | last quarter   | next quarter    |
+ * | 2-3 quarters           | X quarters ago | in X quarters   |
+ * | 1 year                 | last year      | next year       |
+ * | 2+ years               | X years ago    | in X years      |
+ *
```
kossnocorp, line 59:
```suggestion
```

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,214 @@
...
+ * @example
+ * // Use the style option to force the function to use short values. Without setting it, the example would return "in 2 years".
+ * intlFormatDistance(
+ *   new Date(1988, 3, 4, 11, 30, 0),
+ *   new Date(1986, 3, 4, 11, 30, 0),
+ *   { style: 'short' }
+ * )
+ * //=> 'in 2 yr'
+ */
+
```
kossnocorp, line 131:
```suggestion
```

kossnocorp (approved):
👏

orhangazi:
Hi, I can't pass this [error](https://github.com/date-fns/date-fns/pull/2173#issuecomment-821888470). I am using react native version is 0.68.2, react version is 17.0.2 and using date-fns version is 2.29.3.  Code reference doc [here](https://date-fns.org/v2.29.3/docs/intlFormatDistance#examples).
```
lastMessageDeliveryDate = intlFormatDistance(
			new Date(1986, 3, 4, 11, 30, 0),
			new Date(1986, 3, 4, 10, 30, 0)
		)
```
I am getting this error:
image

src/intlFormatDistance/index.ts
```diff
@@ -0,0 +1,213 @@
...
+ * @throws {RangeError} `date` must not be Invalid Date
+ * @throws {RangeError} `baseDate` must not be Invalid Date
+ * @throws {RangeError} `options.unit` must not be invalid Unit
+ * @throws {RangeError} `options.locale` must not be invalid locale
+ * @throws {RangeError} `options.localeMatcher` must not be invalid localeMatcher
+ * @throws {RangeError} `options.numeric` must not be invalid numeric
+ * @throws {RangeError} `options.style` must not be invalid style
+ *
+ * @example
+ * // What is the distance between the dates when the fist date is after the second?
```
hrdyjan1, line 81: Typo - first

JestemTrzcinska:
Hi, I also encountered [this](https://github.com/date-fns/date-fns/pull/2173#issuecomment-821888470) problem. The same result as [here](https://github.com/date-fns/date-fns/pull/2173#issuecomment-1262715438).
It works good on newer iOS simulators, but it doesn't on Android simulators and old iPhone simulators like iPhone 6s with iOS 13.7. 

Any workaround? I need something exactly like intlFormatDistance with its "... ago" etc. 

Making if statement is not good way for me, because I need it to work on every devices.

```
"date-fns": "^2.29.3",
"react": "18.2.0",
"react-native": "0.71.6",
```

PR #2185, opened 2021-02-03
Fix differenceInMonths

tan75:
To make this function cater for:
- end of February dates
- cases with 1 full calendar month if the end of month of the left date is bigger than the end of month of the right date


kossnocorp (approved):
Well done!

PR #2200, opened 2021-02-14
Add accept offsets positive and negative

tan75:
This will allow parseJSON to accept formats with positive and negative offsets. 
Documentation pending.
WIP


leshakoss (approved):
🙏 👏

PR #2205, opened 2021-02-16
Fix date formatting in Latvian

tan75:
As per
https://unicode.org/cldr/cldr-aux/charts/32/summary/lv.html#1882


src/locale/lv/_lib/formatLong/index.js
```diff
@@ -1,41 +1,41 @@
 import buildFormatLongFn from '../../../_lib/buildFormatLongFn/index'
 
 var dateFormats = {
-  full: "y. 'gada' M. MMMM., EEEE",
-  long: "y. 'gada' M. MMMM",
+  full: "y. 'gada' d. MMMM, EEEE",
```
gkuznets, line 4: Actually, according to this https://unicode.org/cldr/cldr-aux/charts/32/summary/lv.html#1882
format should be "EEEE, y. 'gada' d. MMMM".

tan75: > Actually, according to this https://unicode.org/cldr/cldr-aux/charts/32/summary/lv.html#1882
> format should be "EEEE, y. 'gada' d. MMMM".

Yes, you are right. Will fix it shortly! Thanks!

tan75: @gkuznets
should be good now

gkuznets: Thanks!

gkuznets (approved):

kossnocorp (approved):
🙏

PR #2213, opened 2021-02-19
Fix short date separator for Romanian locale

razvanmitre:
The correct and most used separator in Romanian language is . (period) not / (slash)


leshakoss:
Hi @date-fns/i18n-ro! Could you please confirm that this change is correct?

jsergiu (approved):

leshakoss (approved):
Thanks!

PR #2214, opened 2021-02-19
Add next mon tue

tan75:
Add nextTuesday() with minor changes


src/nextMonday/index.ts
```diff
@@ -0,0 +1,45 @@
...
+      return addDays(convertedDate, 5)
+    case 4:
+      return addDays(convertedDate, 4)
+    case 5:
+      return addDays(convertedDate, 3)
+    case 6:
+      return addDays(convertedDate, 2)
+    default:
+      return convertedDate
+  }
```
kossnocorp, line 44:
```suggestion
  return addDays(date, map[getDay(date)])
```

src/nextMonday/index.ts
```diff
@@ -0,0 +1,45 @@
+import requiredArgs from '../_lib/requiredArgs/index'
+import getDay from '../getDay'
+import addDays from '../addDays'
+import toDate from '../toDate/index'
+/**
```
kossnocorp, line 5:
```suggestion

const map = [1, 7, 6, 5, 4, 3, 2]

/**
```

kossnocorp (changes requested):
The logic is great, but I would simplify it a bit for build size sake. I suggest applying the same approach as in `nextMonday` everywhere.

src/nextDay/index.ts
```diff
@@ -0,0 +1,51 @@
...
+ * const result = nextMonday(new Date(2020, 2, 22), 1)
+ * => new Date(2020, 2, 23)
+ *
+ * @example
+ * When is the next Tuesday after Mar, 22, 2020?
+ * const result = nextMonday(new Date(2020, 2, 22), 2)
+ * => new Date(2020, 2, 24)
+ **/
+
+export default function nextDay(date: Date, day: number): Date {
```
kossnocorp, line 32:
```suggestion
export default function nextDay(date: Date | number, day: number): Date {
```

src/nextDay/index.ts
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * @name nextDay
+ * @category Weekday Helpers
+ * @summary When is the next day of Week?
+ *
+ * @description
+ * When is the next day of week?
+ * {0|1|2|3|4|5|6} the day of week, 0 represents Sunday
+ *
+ * @param {Date} date - the date to check
```
kossnocorp, line 16:
```suggestion
 * @param {Date | number} date - the date to check
```

src/nextDay/index.ts
```diff
@@ -0,0 +1,51 @@
...
+const baseMap = [7, 6, 5, 4, 3, 2, 1]
+
+/**
+ * @name nextDay
+ * @category Weekday Helpers
+ * @summary When is the next day of Week?
+ *
+ * @description
+ * When is the next day of week?
+ * {0|1|2|3|4|5|6} the day of week, 0 represents Sunday
```
kossnocorp, line 14:
```suggestion
 * When is the next day of the week? 0-6 the day of the week, 0 represents Sunday.
```

src/nextDay/index.ts
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @example
+ * When is the next Monday after Mar, 22, 2020?
+ * const result = nextMonday(new Date(2020, 2, 22), 1)
+ * => new Date(2020, 2, 23)
+ *
+ * @example
+ * When is the next Tuesday after Mar, 22, 2020?
+ * const result = nextMonday(new Date(2020, 2, 22), 2)
+ * => new Date(2020, 2, 24)
```
kossnocorp, line 29: Please update the examples

kossnocorp (changes requested):

src/nextDay/index.ts
tan75: Updated as below:

```
 * @example
 * When is the next Monday after Mar, 20, 2020?
 * const result = nextDay(new Date(2020, 2, 20), 1)
 * => new Date(2020, 2, 23)
 *
 * @example
 * When is the next Tuesday after Mar, 21, 2020?
 * const result = nextDay(new Date(2020, 2, 21), 2)
 * => new Date(2020, 2, 24)
```

src/nextDay/index.ts
```diff
@@ -0,0 +1,56 @@
...
+ * const result = nextDay(new Date(2020, 2, 20), 1)
+ * => new Date(2020, 2, 23)
+ *
+ * @example
+ * When is the next Tuesday after Mar, 21, 2020?
+ * const result = nextDay(new Date(2020, 2, 21), 2)
+ * => new Date(2020, 2, 24)
+ **/
+
+export default function nextDay(date: Date | number, day: number): Date {
```
kossnocorp, line 33: Let's instead of checking the type in runtime and throwing `throw new RangeError('Invalid Day')` use an alias for day: 
```suggestion
export default function nextDay(date: Date | number, day: Day): Date {
```

...and add to `types.ts`:

```ts
export type Day = 0 | 1 | 2 | 3 | 4 | 5 | 6
```

src/nextDay/index.ts
```diff
@@ -0,0 +1,56 @@
...
+  requiredArgs(2, arguments)
+  // Throw an exception `Invalid Day` if the day is out or range
+  if (!baseMap.includes(day + 1)) {
+    throw new RangeError('Invalid Day')
+  }
```
kossnocorp, line 38: Instead of adding runtime checks and adding extra bytes, let's give users a choice, use TypeScript/Flow or getting obscure errors when they misuse the API.

src/nextDay/index.ts
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @example
+ * When is the next Monday after Mar, 20, 2020?
+ * const result = nextDay(new Date(2020, 2, 20), 1)
+ * => new Date(2020, 2, 23)
+ *
+ * @example
+ * When is the next Tuesday after Mar, 21, 2020?
+ * const result = nextDay(new Date(2020, 2, 21), 2)
+ * => new Date(2020, 2, 24)
```
kossnocorp, line 30: Please make the example consistent with other functions (see for example `addDays`: https://github.com/date-fns/date-fns/blob/master/src/addDays/index.ts#L25).

Just use `toString` on the date and cut off the timezone (to make it universal):

```js
new Date(2020, 2, 24).toString()
//=> "Tue Mar 24 2020 00:00:00 GMT+0800 (Singapore Standard Time)"
```

src/nextDay/index.ts
```diff
@@ -0,0 +1,56 @@
...
+import addDays from '../addDays'
+import toDate from '../toDate'
+import isValid from '../isValid/index'
+
+const baseMap = [7, 6, 5, 4, 3, 2, 1]
+
+/**
+ * @name nextDay
+ * @category Weekday Helpers
+ * @summary When is the next day of Week?
```
kossnocorp, line 12:
```suggestion
 * @summary When is the next day of the week?
```

src/nextMonday/index.ts
```diff
@@ -0,0 +1,31 @@
...
+import toDate from '../toDate/index'
+import isValid from '../isValid/index'
+
+/**
+ * @name nextMonday
+ * @category Weekday Helpers
+ * @summary When is the next Monday?
+ *
+ * @description
+ *  When is the next day of the week? 1 represents Monday.
```
kossnocorp, line 12:
```suggestion
 * When is the next Monday?
```

kossnocorp (approved):
Please make examples in both `nextMonday` and `nextDay` consistent with other functions and remove extra runtime checks. Just to reiterate, we don't remove `requiredArgs` because other functions use that, but in v3, we'll remove it.

src/nextDay/index.ts
tan75: Done

kossnocorp (approved):
👏

PR #2221, opened 2021-02-21
feat: add custom businessDays option to addBusinessDays

taneba:
This PR adds custom businessDays option to addBusinessDays (and subBusinessDays).

Typescriptifed subBusinessDays as well https://github.com/date-fns/date-fns-typescriptify/issues/139

```
addBusinessDays(new Date(), 1, {
  businessDays: [1, 2, 3] // <- here
})
```

API design:
* default is Monday to Friday which means it doesn't affect current behavior.
* options now accepts only businessDays, but there could be other options in the future (say holidays).

A Question for maintainers:
I couldn't build docs because of the error below but it happens on master branch as well. I missed something?

```
TypeError: Cannot read property 'customTags' of undefined
    at /Users/nakamurayoshihiro/oss/date-fns/scripts/build/docs.js:43:13
    at Array.map (<anonymous>)
    at generateDocsFromSource (/Users/nakamurayoshihiro/oss/date-fns/scripts/build/docs.js:41:6)
```

### DIFF
--- a/src/addBusinessDays/index.ts
+++ b/src/addBusinessDays/index.ts
@@ -1,9 +1,6 @@
-import isWeekend from '../isWeekend/index'
 import toDate from '../toDate/index'
 import toInteger from '../_lib/toInteger/index'
 import requiredArgs from '../_lib/requiredArgs/index'
-import isSunday from '../isSunday/index'
-import isSaturday from '../isSaturday/index'
 
 /**
  * @name addBusinessDays
@@ -15,6 +12,8 @@ import isSaturday from '../isSaturday/index'
  *
  * @param {Date|Number} date - the date to be changed
  * @param {Number} amount - the amount of business days to be added. Positive decimals will be rounded using `Math.floor`, decimals less than zero will be rounded using `Math.ceil`.
+ * @param {Object} [options] - an object with options.
+ * @param {Number[]} [options.businessDays=[1, 2, 3, 4, 5]] - the business days. default is Monday to Friday.
  * @returns {Date} the new date with the business days added
  * @throws {TypeError} 2 arguments required
  *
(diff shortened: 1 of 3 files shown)


taneba:
@codinsonn @kossnocorp 
Hi, do you have any thoughts on this?

src/addBusinessDays/index.ts
```diff
@@ -25,40 +24,59 @@ import isSaturday from '../isSaturday/index'
  */
 export default function addBusinessDays(
   dirtyDate: Date | number,
-  dirtyAmount: number
+  dirtyAmount: number,
+  dirtyOptions?: {
+    businessDays?: number[]
+  }
```
codinsonn, line 30: Maybe I'm not entirely following how the options work, but I'm not sure the array of weekdays is the most DX-friendly way to go to exclude holidays next to weekends, but it could work as one version of doing so.

@kossnocorp Would an api with an array of holidays as a second way to exclude them be a good idea?

e.g.

```suggestion
  dirtyOptions?: {
    businessDays?: number[],
    excludeDates?: (Date | number)[],
  }
```

codinsonn (commented):
Some minor api suggestions:
- `dirtyOptions.businessDays` can work as one solution to the problem, but may not be the best DX-friendly way to solve this if it is the only option
- Could be worth exploring a `dirtyOptions.excludeDates` (or `holidays`, since it would be an array of those) ?

PS: No idea why the Docs won't build (have only done it myself once, a few years back 😅 )

src/addBusinessDays/index.ts
taneba: Thanks for your feedback and suggestion!
Having excludeDates would be nice, but for parameter type `Date[]` makes it clearer when to use it imo.
So businessDays, I was thinking of a case when business days are not from Monday to Friday (e.g. A store is open only on Monday, Wednesday, Friday, then businessDays would be `[1, 3, 5]` ).
But excludeDates (or holidays?) wants to solve the cases like excluding certain dates, whereas businessDays is for days of the week.

codinsonn: Right, that makes tot sense 🙃 Had not considered that scenario. It is a good api for solving that specific usecase. I still have to get around to testing the code, but I’m definitely a fan of the api now.

(I think I’ll try to add the holidays solution on dirtyOptions once this is merged, but I’m hesitant to sign off on this without the docs building or one of the core contributors giving their approval ofcourse)

taneba: > I think I’ll try to add the holidays solution on dirtyOptions once this is merged
cool 👍

AsemK:
Any plans to get this merged?

codyfrisch:
Clearly a project that doesn't care about non-US developers (or those developing for non-US markets). How sad.

PR #2222, opened 2021-02-22
move isEqual to ts

maximtop:
This pr fixes https://github.com/date-fns/date-fns-typescriptify/issues/67

### DIFF
--- a/src/isEqual/index.ts
+++ b/src/isEqual/index.ts
@@ -13,8 +13,8 @@ import requiredArgs from '../_lib/requiredArgs/index'
  *
  * - [Changes that are common for the whole library](https://github.com/date-fns/date-fns/blob/master/docs/upgradeGuide.md#Common-Changes).
  *
- * @param {Date|Number} dirtyLeftDate - the first date to compare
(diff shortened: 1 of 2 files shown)


src/isEqual/index.ts
```diff
@@ -26,10 +26,10 @@ import requiredArgs from '../_lib/requiredArgs/index'
...
  * //=> false
  */
-export default function isEqual(dirtyLeftDate, dirtyRightDate) {
+export default function isEqual(dirtyLeftDate: Date | number, dirtyRightDate: Date | number): boolean {
   requiredArgs(2, arguments)
 
-  var dateLeft = toDate(dirtyLeftDate)
-  var dateRight = toDate(dirtyRightDate)
+  const dateLeft: Date = toDate(dirtyLeftDate)
+  const dateRight: Date = toDate(dirtyRightDate)
```
kossnocorp, line 33: I don't think it's required, could you please confirm?

```suggestion
  const dateLeft = toDate(dirtyLeftDate)
  const dateRight = toDate(dirtyRightDate)
```

kossnocorp (changes requested):
Please check if my change can be safely applied.

PR #2230, opened 2021-02-22
move isMonday to ts

maximtop:
this pr fixes https://github.com/date-fns/date-fns-typescriptify/issues/75

### DIFF
--- a/src/isMonday/index.ts
+++ b/src/isMonday/index.ts
@@ -13,7 +13,7 @@ import requiredArgs from '../_lib/requiredArgs/index'
  *
  * - [Changes that are common for the whole library](https://github.com/date-fns/date-fns/blob/master/docs/upgradeGuide.md#Common-Changes).
  *
- * @param {Date|Number} date - the date to check
+ * @param {Date|Number} dirtyDate - the date to check
  * @returns {Boolean} the date is Monday
  * @throws {TypeError} 1 argument required
  *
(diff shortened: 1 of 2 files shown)


src/isMonday/index.ts
```diff
@@ -13,7 +13,7 @@ import requiredArgs from '../_lib/requiredArgs/index'
  *
  * - [Changes that are common for the whole library](https://github.com/date-fns/date-fns/blob/master/docs/upgradeGuide.md#Common-Changes).
  *
- * @param {Date|Number} date - the date to check
+ * @param {Date|Number} dirtyDate - the date to check
```
kossnocorp, line 16: Please leave it as is. In the code, the name makes sense because it highlights that the date is not converted yet (and not everywhere, here for example it could be simplidied), but in the docs, is just noise.

```suggestion
 * @param {Date|Number} date - the date to check
```

src/isMonday/index.ts
```diff
@@ -22,7 +22,7 @@ import requiredArgs from '../_lib/requiredArgs/index'
  * var result = isMonday(new Date(2014, 8, 22))
  * //=> true
  */
-export default function isMonday(dirtyDate) {
+export default function isMonday(dirtyDate: Date | number): boolean {
```
kossnocorp, line 25:
```suggestion
export default function isMonday(date: Date | number): boolean {
```

kossnocorp (changes requested):
Please address my comments. Besides that, good job!

kossnocorp (approved):
👍

PR #2235, opened 2021-02-22
🚧 Housekeeping - TypeScript configuration and development environment

nonewcode:
TBC


tan75:
hi @nonewcode 
Thank you for this PR however, unfortunately, we can not accept it as there are too many changes involved. 
We prefer a separate PR for every change - it makes it easier for us to review them.
And the `prettier` configuration shouldn't be in the package.json. 
Closing this PR and thanks again!  🙏

PR #2172, opened 2021-01-23
Add intlFormat function

rikkalo:
- Add intlFormat with Intl.DateTimeFormat
- Add tests for intlFormat

PR #2236, opened 2021-02-22
Update `isMonday` to use correct argument name

codecog:
A regression was introduced in https://github.com/date-fns/date-fns/pull/2230 where the argument name was changed but the name passed to `toDate()` was missed. This change addresses that regression.


codecog:
Please see https://github.com/date-fns/date-fns/pull/2172 for the cause of the failing tests.

kossnocorp (approved):
Appreciate it a lot! It's my bad that I missed that.

PR #2238, opened 2021-02-23
Migrate `getWeekOfMonth` to typescript

codecog:
This change moves `getWeekOfMonth` over to typescript. A few key points:
- Removed test to assert that the method will throw if the required amount of arguments are not passed as this can be checked with types
- added `@babel/plugin-proposal-nullish-coalescing-operator` to take advantage of setting a default to `weekStartsOn` of `0` by using `??`
- Updated parts of the JSDoc comments to use tsdoc


codecog:
@kossnocorp I had issues running `./scripts/build/build.sh` as it throws an error to do with the docs. Have you run into this problem before?

src/getWeekOfMonth/test.js
```diff
@@ -71,10 +71,6 @@ describe('getWeekOfMonth', function() {
     assert(isNaN(result))
   })
 
-  it('throws TypeError exception if passed less than 1 argument', function() {
-    assert.throws(getWeekOfMonth.bind(null), TypeError)
-  })
-
```
kossnocorp, line 77: Please keep this and the check in the code and also don't forget to keep `@throws` in the docs. I know that TypeScript solves the problem and we'll remove all type checks in v3, but this is a breaking change and we're not ready for a major release.

src/getWeekOfMonth/index.ts
```diff
@@ -0,0 +1,55 @@
...
+    startWeekDay >= weekStartsOn
+      ? weekStartsOn + 7 - startWeekDay
+      : weekStartsOn - startWeekDay
+
+  return dayOfMonth > lastDayOfFirstWeek
+    ? weekNumber + Math.ceil((dayOfMonth - lastDayOfFirstWeek) / 7)
+    : weekNumber
+}
+
+export default getWeekOfMonth
```
kossnocorp, line 55: Please rewrite to a normal function `export default function getWeekOfMonth`.

src/getWeekOfMonth/index.ts
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @param date - The given date
+ * @param options - An object with options
+ * @param [options.locale=defaultLocale] - The locale object. See {@link https://date-fns.org/docs/Locale Locale}
+ * @param [options.weekStartsOn=0] - The index of the first day of the week (0 - Sunday)
+ * @returns The week of month
+ *
+ * @example // Which week of the month is 9 November 2017?
+ * const result = getWeekOfMonth(new Date(2017, 10, 9));
+ */
```
kossnocorp, line 30: Please preserve the original JSDoc, until v3 when all code is in TypeScript we're forced to use it.

src/getWeekOfMonth/index.js
```diff
@@ -1,72 +0,0 @@
...
-  }
-
-  var weekNumber = 1
-
-  if (currentDayOfMonth > lastDayOfFirstWeek) {
-    var remainingDaysAfterFirstWeek = currentDayOfMonth - lastDayOfFirstWeek
-    weekNumber = weekNumber + Math.ceil(remainingDaysAfterFirstWeek / 7)
-  }
-  return weekNumber
-}
```
kossnocorp, line 72: Could you please rework your commit so git understands that it is a rename to preserve the git history?

src/getWeekOfMonth/index.ts
```diff
@@ -0,0 +1,55 @@
...
+const getWeekOfMonth = (
+  date: Date | number,
+  { locale, weekStartsOn } = defaultOptions
```
kossnocorp, line 33: Please assign type `LocaleOptions & WeekStartOptions` from https://github.com/date-fns/date-fns/blob/master/src/types.ts. Also `= defaultOptions` breaks when you pass any object, so please get rid of it and keep it in the code or `{ locale, weekStartsOn = 0 }`.

kossnocorp (changes requested):
Thank you for the pull request! Please address my comments. Docs generation will fix once you restore the original JSDoc. Also, please rebase from the master branch, where we fixed tests.

PR #2251, opened 2021-02-24
migrate set function

maximtop:
This PR fixes: https://github.com/date-fns/date-fns-typescriptify/issues/120


rikkalo (approved):
Thank you for your contribution 👍

src/set/index.ts
```diff
@@ -3,6 +3,16 @@ import setMonth from '../setMonth/index'
...
+interface Values {
+  year?: number;
+  month?: number;
+  date?: number;
+  hours?: number;
+  minutes?: number;
+  seconds?: number;
+  milliseconds?: number;
+}
+
```
kossnocorp, line 15: Could you please move it to `src/types.ts`

kossnocorp (changes requested):
Thank you a lot, please move the interface to shared types cc @rikkalo

src/set/index.ts
maximtop: I've moved

kossnocorp (approved):
🙏

PR #2256, opened 2021-02-25
Add check for Feb 28 and 29 of the same year

tan75:
Add extra check to test if dates are either Feb 28 or Feb 29 of the same year
Add extra tests for intervalToDuration() to test the same

### DIFF
--- a/src/differenceInMonths/test.js
+++ b/src/differenceInMonths/test.js
@@ -5,6 +5,41 @@ import assert from 'power-assert'
 import differenceInMonths from '.'
 
 describe('differenceInMonths', function () {
+  it('returns the number of full months between the given dates - end of Feb', function () {
+    const start = new Date('Feb 29 2012 09:00:00')
+    const end = new Date('Feb 28 2012 10:00:00')
+    var result = differenceInMonths(start, end)
+    assert(result === 0)
+  })
+
+  it('returns the number of full months between the given dates - end of Feb', function () {
+    const start = new Date('Feb 28 2012 09:00:00')
+    const end = new Date('Feb 29 2012 10:00:00')
+    var result = differenceInMonths(start, end)
+    assert(result === 0)
+  })
+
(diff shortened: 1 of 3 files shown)


kossnocorp (approved):
Looks good!

src/differenceInMonths/test.js
```diff
@@ -5,6 +5,41 @@ import assert from 'power-assert'
...
+    var result = differenceInMonths(start, end)
+    assert(result === 0)
+  })
+
+  it('returns the number of full months between the given dates - end of Feb', function () {
+    const start = new Date('Feb 29 2012 09:00:00')
+    const end = new Date('Feb 29 2012 10:00:00')
+    var result = differenceInMonths(start, end)
+    assert(result === 0)
+  })
```
kossnocorp, line 34: Please do the following:

1. Join these tests into a single one, as all of them have the same description
2. Move edge case tests to the end of the file, you might even wrap them into `describe('edge cases', ...` or `describe('regression tests', ...`

src/intervalToDuration/test.js
```diff
@@ -4,8 +4,50 @@
...
+    const result = intervalToDuration({ start, end })
+    assert.deepEqual(result, {
+      years: 0,
+      months: 0,
+      days: 0,
+      hours: 1,
+      minutes: 0,
+      seconds: 0,
+    })
+  })
```
kossnocorp, line 48: Same here, join tests and move to the bottom.

Could you also fix the `Returns` -> `returns`.

kossnocorp (changes requested):

src/differenceInMonths/test.js
tan75: Done

src/intervalToDuration/test.js
tan75: Done

kossnocorp (approved):
👍

PR #2266, opened 2021-03-02
Fix formatDuration week translation in de locale

helmut-lang:
Some _week_ related output produced by `formatDuration` for the `de` locale is not correct.
This PR fixes those issues.

It was a bit hard for me to tell, if my changes produced the desired output.
So I've added some `formatDuration` output to the `locale-snapshot` files.

Additionally, there was an example for formatDuration given in the sources:
```
formatDuration({ years: 0, months: 9 }, null, { zero: true })
```
I think this example is not correct, since the function actually only takes two arguments. I've updated the example to match the actual function signature.


src/locale/de/_lib/formatDistance/index.js
```diff
@@ -2,172 +2,172 @@ var formatDistanceLocale = {
...
+      one: 'etwa einer Woche',
+      other: 'etwa {{count}} Wochen',
+    },
   },
 
   xWeeks: {
     standalone: {
-      one: 'ein Woche',
-      other: '{{count}} Wochen'
+      one: 'eine Woche',
```
leshakoss, line 97: Hi @date-fns/i18n-de could you please confirm that this is correct?

pex: LGTM

pex (approved):

leshakoss (approved):
The main change of `formatDuration` patterns was already overridden by #2576. But we're merging this PR for snapshot improvements anyway. Thanks a lot!

PR #2270, opened 2021-03-02
Update README.md

sarkiroka:
minimal description correction: "npm install" saves the changes by default - doesn't need "--save" option. and nobody use npm with "install" instead of "i".

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ The library is available as an [npm package](https://www.npmjs.com/package/date-
 To install the package run:
 
 ```bash
-npm install date-fns --save
+npm i date-fns
 # or with yarn
 yarn add date-fns
 ```


sarkiroka:
Sorry guys, my change not affect the code, I don't know what happened

tan75 (approved):
Thank you!

tan75:
Hi @sarkiroka 

unfortunately we will not be able to proceed with merging your PR as it turns out that not everyone uses modern versions of NPM right now. 
Thanks again for your contribution however I am closing this PR right now.

sarkiroka:
@tan75 I understand it. By what statistics do I know this? I ask so I can avoid the problem later if I create another PR.

kossnocorp:
@sarkiroka, there are no statistics behind the reasoning but just our subjective notion. The code examples must be as explicit and compatible as possible. There's no value in making it short-n-sweet but confusing for new developers. Also, you might notice that `--save` on the end so you can easily copy without it.

Issue #2255, opened 2021-02-25
intervalToDuration bug with Feb 28


mraballard:
I think I found an interesting bug here that occurs with Feb 28, no matter the year.

```
let start = new Date("Feb 28 2012 09:00:00")
let end = new Date("Feb 28 2012 10:00:00")
fns.intervalToDuration({start, end})
```

returns 
```
{
  years: 0,
  months: 1,
  days: 28,
  hours: 1367,
  minutes: 164040,
 }
```
  
```
start = new Date("Feb 28 2021 09:00:00")
end = new Date("Feb 28 2021 10:00:00")
fns.intervalToDuration({start, end})
```

returns
```
{
  years: 0,
  months: 1,
  days: 27,
  hours: 1320,
  minutes: 158400,
  seconds: 19008000
}
```

tan75:
hi @mraballard,

yes you are right! This is going to be fixed shortly - WIP.

PR #2287, opened 2021-03-04
Switch dateLeft and dateRight in intervalToDuration

lorenzvanherwaarden:
Functions like differenceInMonths, differenceInDays, ... , which are used in intervalToDuration, expect dateLeft to be the later date and dateRight the earlier date. Currently, intervalToDuration sets its dateLeft to its earlier date and dateRight to its later date. This has been fixed in this PR: by setting dateLeft to the end date and dateRight to the start date.

While the dateLeft and dateRight arguments can normally be switched in these functions, because they work in both cases as they use Math.abs, a sign variable, ..., meaning that it shouldn't matter in what order intervalToDuration passes dateLeft and dateRight to these functions, there are two reasons why I still think this change is useful:
- It's more logical that if the dateLeft is the later date in these 'difference' functions, this is also the case in the intervalToDuration function
- In my case, the intervalToDuration bug with Feb 28th #2255, would not have occurred if dateLeft would have been set as the later date in intervalToDuration

### DIFF
--- a/src/intervalToDuration/index.js
+++ b/src/intervalToDuration/index.js
@@ -37,14 +37,14 @@ import sub from '../sub/index'
 export default function intervalToDuration({ start, end }) {
   requiredArgs(1, arguments)
 
-  const dateLeft = toDate(start)
-  const dateRight = toDate(end)
+  const dateLeft = toDate(end)
+  const dateRight = toDate(start)
 
   if (!isValid(dateLeft)) {
-    throw new RangeError('Start Date is invalid')
+    throw new RangeError('End Date is invalid')
   }
   if (!isValid(dateRight)) {
-    throw new RangeError('End Date is invalid')
+    throw new RangeError('Start Date is invalid')
   }
 
   const duration = {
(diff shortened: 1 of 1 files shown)


leshakoss:
I feel like it's the differenceInX arguments that are named incorrectly, they should be something like `dateMinuend` and `dateSubtrahend`. Or just `laterDate` and `earlierDate`

lorenzvanherwaarden:
@leshakoss I agree, using laterDate and earlierDate is more logical. I can adapt the PR if needed. I just tackled it this way because there are many differenceInX functions and I thought that was done on purpose and the error was in the intervalToDuration function.

tan75 (dismissed):
Thank you!

tan75:
@lorenzvanherwaarden 
unfortunately we can not accept this PR right now as this will introduce a breaking change. 
However we might re-visit this in future.
@leshakoss what do you think?

tan75:
@kossnocorp I looked into this PR and I think it makes sense. Even if we swap the end date and the start date, it will produce a correct result as it applies the `sign` during the computation process: 
`const sign = compareAsc(dateLeft, dateRight)`

PR #2290, opened 2021-03-05
i18n support for Albanian language

arditdine:
Following the guide here: https://date-fns.org/v2.18.0/docs/I18n-Contribution-Guide

I added support for the Albanian language.


src/locale/sq/_lib/localize/index.js
```diff
@@ -0,0 +1,184 @@
...
+    switch (rem100 % 10) {
+      case 1:
+        return number
+      case 2:
+        return number
+      case 3:
+        return number
+    }
+  }
+  return number
```
tan75, line 147: Hi @arditdine 
Thanks a lot for your contribution - support for Albanian language would be a great addition to this library!
We are currently reviewing your PR and here are our suggestions: 

It looks like this code returns `number` in any case. Hence would that be better just to return `number` here?

tan75 (changes requested):

src/locale/sq/_lib/localize/index.js
arditdine: I've updated this with some examples I managed to find online. 
https://sq.wikipedia.org/wiki/How_I_Met_Your_Mother_(sezoni_i_1-r%C3%AB)
https://sq.wikipedia.org/wiki/How_I_Met_Your_Mother_(sezoni_i_4-t)

For hours we don't use suffix so I return only the number.

src/locale/sq/index.js
```diff
@@ -0,0 +1,27 @@
...
+var locale = {
+  code: 'sq',
+  formatDistance: formatDistance,
+  formatLong: formatLong,
+  formatRelative: formatRelative,
+  localize: localize,
+  match: match,
+  options: {
+    weekStartsOn: 1 /* Monday */,
+    firstWeekContainsDate: 1
```
tan75, line 23: @arditdine could you please double-check this value: 
`firstWeekContainsDate: 1`
Wikipedia says its 4 in the EU and most of other European countries.
Have you got any sources to confirm if it is 1 or 4?

tan75 (changes requested):

src/locale/sq/index.js
arditdine: Yes i've confirmed that it is 1. In date-fns documentation:
https://date-fns.org/v2.19.0/docs/I18n-Contribution-Guide#index.js

You have this link:
http://www.pjh2.de/datetime/weeknumber/wnd.php?l=en

tan75 (approved):
Much appreciated!

src/locale/sq/_lib/match/index.js
```diff
@@ -0,0 +1,137 @@
+import buildMatchPatternFn from '../../../_lib/buildMatchPatternFn/index'
```
kossnocorp, line 1: @arditdine please update patterns below, so it would fix `parse` (the third column should not have that mane `Invalid Date`: https://github.com/date-fns/date-fns/pull/2290/files#diff-b1642ead537bf233bdd80d82877ee7a3e3ff81641532ea7b4947ce6e242df01f)

kossnocorp (changes requested):
Please add support for `parse`. It should be straightforward but feel free to ask any questions.

src/locale/sq/_lib/match/index.js
arditdine: Sorry my bad, didn't really understand why the last column was always `Invalid Date` but I've fixed the matching regex

arditdine: @kossnocorp any updates on this?

kossnocorp: @arditdine thank you for updating the PR, it will come with the next version of the library.

kossnocorp (approved):

PR #2297, opened 2021-03-08
Migrate startOfWeek function to TS

rikkalo:
This pr fixes date-fns/date-fns-typescriptify#136

### DIFF
--- a/src/startOfWeek/index.ts
+++ b/src/startOfWeek/index.ts
@@ -3,16 +3,6 @@ import toDate from '../toDate/index'
 import toInteger from '../_lib/toInteger/index'
 import requiredArgs from '../_lib/requiredArgs/index'
 
-export enum Day {
-  SUNDAY,
-  MONDAY,
-  TUESDAY,
-  WEDNESDAY,
-  THURSDAY,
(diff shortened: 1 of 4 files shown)


src/startOfWeek/index.ts
```diff
@@ -1,7 +1,18 @@
...
 
+export enum Day {
+  SUNDAY,
+  MONDAY,
+  TUESDAY,
+  WEDNESDAY,
+  THURSDAY,
+  FRIDAY,
+  SATURDAY,
+}
```
kossnocorp, line 14: This is clever, but there's a problem. 

First of all it will require importing enum everywhere, for example:  `nextDay(day, Dat.SUNDAY)`. Not very handy, yet explicit. Should we allow using enums (`export type Day = DayIndex | DayEnum`) as well, for type nerds? 

Secondly, as I can see it's not used and you changed your mind, but just in case, there's Day alias in types.

kossnocorp (changes requested):
Please remove redundant `Day`

kossnocorp (approved):
🙏

PR #2299, opened 2021-03-08
Migrate parseJSON function to TS

rikkalo:
This pr fixes date-fns/date-fns-typescriptify#118


src/parseJSON/index.ts
```diff
@@ -35,21 +35,22 @@ import requiredArgs from '../_lib/requiredArgs/index'
  * @returns {Date} the parsed date in the local time zone
  * @throws {TypeError} 1 argument required
  */
-export default function parseJSON(argument) {
+export default function parseJSON(argument: Date | string | number): Date {
```
kossnocorp, line 38: Let's make it `string`. I know that originally it was like that, but having this kind of type is demising the whole idea of types 😉

```suggestion
export default function parseJSON(argument: string): Date {
```

kossnocorp (changes requested):
Let's simplify the argument type. It will not affect the actual version, so it's safe and future-proof.

src/parseJSON/index.ts
rikkalo: But we can use dates and timestamps.
``` it('clones a date object', () => {
    const date = new Date(2000, 2, 15, 5, 20, 10, 20)
    const parsedDate = parseJSON(date)
    assert.deepEqual(parsedDate, date)
    assert.notEqual(parsedDate, date)
  })

  it('assumes a number is a timestamp', () => {
    const date = new Date(2000, 2, 15, 5, 20, 10, 20)
    const timestamp = date.getTime()
    const parsedDate = parseJSON(timestamp)
    assert.deepEqual(parsedDate, date)
  })```

kossnocorp: @rikkalo, yeah, I see, but does it make sense? I suggest removing the tests as well.

If I were a user, I would assume that the input would be a string and expect that I'll be warned if I accidentally pass a number or a date that does not require parsing. In the best-case scenario, it would be just a waste. In the worst-case, a random number might be passed, which will create a bug.

Also, `Date.parse` also accepts all kind of arguments, but in types is the string:

image

kossnocorp (approved):
🙏

PR #2308, opened 2021-03-10
Migrate isValid function to TS

rikkalo:
This pr fixes date-fns/date-fns-typescriptify#101


src/isValid/index.ts
```diff
@@ -58,9 +58,14 @@ import requiredArgs from '../_lib/requiredArgs/index'
...
  */
-export default function isValid(dirtyDate) {
+export default function isValid(
+  dirtyDate: unknown
+): dirtyDate is Date | number {
   requiredArgs(1, arguments)
 
-  var date = toDate(dirtyDate)
-  return !isNaN(date)
+  if (!(dirtyDate instanceof Date) && typeof dirtyDate !== 'number') {
```
fturmel, line 66: I think it would be better to use `isDate` here since it handles an iframe passing edge case not covered by instanceof alone.

fturmel:
@tan75 I think that last change was a mistake. The predicate `is Date | number` works great. The problem I had pointed out a while ago in other comments was that `is Date` was not sufficient.

With predicate, we narrow the result down to Date or number:
Screen Shot 2021-09-15 at 8 20 43 AM

With boolean return, type is still unknown:
Screen Shot 2021-09-15 at 8 20 08 AM

fturmel:
As I pointed out here https://github.com/date-fns/date-fns/pull/2551#issuecomment-878429081, the remaining problem is with the jsdocs / type generator system for these type guards in `isValid` and `isDate` to be correctly carried to `typings.d.ts`.

How to write the jsdocs part: https://github.com/microsoft/TypeScript/issues/25127#issuecomment-587937146

src/isValid/index.ts
kossnocorp: Good point! 👍

PR #2317, opened 2021-03-11
Migrate fromUnixTime function

jooola:
Fixes https://github.com/date-fns/date-fns-typescriptify/issues/43


tan75 (changes requested):
Could you please change `var` to `const`?

jooola:
Just rebased and replaced var with const

PR #2318, opened 2021-03-11
Migrate getWeek function

jooola:
Fixes https://github.com/date-fns/date-fns-typescriptify/issues/58


src/getWeek/index.ts
```diff
@@ -3,7 +3,13 @@ import startOfWeekYear from '../startOfWeekYear/index'
...
 import requiredArgs from '../_lib/requiredArgs/index'
 
-var MILLISECONDS_IN_WEEK = 604800000
+const MILLISECONDS_IN_WEEK = 604800000
+
+export interface GetWeekOptions {
+  locale: Locale,
+  weekStartsOn: 0 | 1 | 2 | 3 | 4 | 5 | 6,
+  firstWeekContainsDate: 1 | 2 | 3 | 4 | 5 | 6 | 7,
+}
```
tan75, line 12: Could you please use the existing types from [types.ts](https://github.com/date-fns/date-fns/blob/master/src/types.ts)

tan75 (changes requested):
Thank you for your contribution! Could you please address my comments? Many thanks!

src/getWeek/index.ts
jooola: Ok, it should be fixed now.

src/getWeek/index.ts
```diff
@@ -35,24 +36,24 @@ var MILLISECONDS_IN_WEEK = 604800000
...
- * var result = getISOWeek(new Date(2005, 0, 2), {
+ * const result = getISOWeek(new Date(2005, 0, 2), {
  *   weekStartsOn: 1,
  *   firstWeekContainsDate: 4
  * })
  * //=> 53
  */
 
-export default function getWeek(dirtyDate, options) {
+export default function getWeek(dirtyDate: Date | number, options?: LocaleOptions & WeekStartOptions & FirstWeekContainsDateOptions): number {
```
jooola, line 52: @tan75 I was wondering, should I default the options to an empty object or is it better to allow it to be undefined ?

tan75: I don't think so.
I defaulted the options to an empty object during the migration of `getWeekYear` and surprisingly introduced a breaking change - you can see the fix in this PR - 
[PR](https://github.com/date-fns/date-fns/pull/2423/files#diff-f4381187a2ef5a29864dc44990590030b44dedb7d9772ed895724a7abd11b2bbL56)

tan75 (approved):

tan75:
@jooola thanks a lot for your work!
I spotted you changed the `FirstWeekContainsDate` values in `types.ts` and just wanted to mention that according to 
[Wikipedia Week Numbering](https://en.wikipedia.org/wiki/Week#Week_numbering) is is always either 1 or 4 (and our Locales can prove it) so I changed it back to `1 | 4`

src/types.ts
```diff
@@ -27,14 +29,12 @@ export interface LocaleOptions {
   locale?: Locale
 }
 
+export type FirstWeekContainsDate = 1 | 4
+
```
kossnocorp, line 33: In the future, please avoid introducing random changes like this. You probably had good intentions and logic behind those, but they aren't evident and produce unnecessary noise and potential merge conflicts for maintainers.

jooola: Haha, this wasn't random, I followed the documentation, which state that it could have been 1-7 and I didn't know what @tan75 just said. But yeah, my bad, sorry for the noise !

I would suggest adding a comment for the type, as it should be documented anyway, and reduce unawareness from dumb contributors.

PR #2319, opened 2021-03-11
Upgrade huksy and lint-staged dependencies

jooola:
Upgraded huksy and lint-staged dependencies with some extra changes :
- Added subfolders and .ts files to the lint-staged glob pattern
- New husky layout with .huksy folder and hooks inside


package.json
```diff
@@ -20,22 +20,17 @@
   "main": "index.js",
   "module": "esm/index.js",
   "scripts": {
+    "postinstall": "husky install",
```
kossnocorp, line 23: What's the point of this? Won't it will affect all users installing date-fns?

I can see they recommend using pinst: https://typicode.github.io/husky/#/?id=yarn-v2

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
```
kossnocorp, line 2: I am not sure if you can verify that it works, but could you please add support for Windows? (There're few paragraphs about that in the documentation).

kossnocorp (changes requested):
Thank you for the contribution. Could you please describe in a paragraph what does this update provides and also address my comments?

package.json
jooola: This project seem to use Yarn v1, I replace the `postinstall` to the `prepare` lifecycle.

If yarn v2 is used, we will need to change this.

.husky/pre-commit
jooola: I added the small helper for windows, it should be working.

jooola:
The main pros of the new versions are that husky is really lightweight and allow to run any kind of scripts (bash, node, python).

Other than that, it simply keep the project dependencies up to date.

package.json
kossnocorp: I mean users that install date-fns will have this command run and throw an error.

kossnocorp:
@jooola got you, thanks, please fix the issue with `postinstall`.

jooola:
@kossnocorp Fixed the postinstall issue. I guess we cannot enforce the use of husky on the developer side.

jooola:
Closing in favor of #2899

PR #2331, opened 2021-03-16
Fix formatDuration function example

fgottschalk:
Executing the given example throws an error, because the function expects an object with options as second argument


tan75 (approved):
Thank you!

PR #2333, opened 2021-03-17
Fix Day type generation and add missing next functions

kossnocorp:
The current type generator is becoming a mess, can't wait until we complete TypeScript migration and get rid of it.


docs/Day.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * @category Types
+ * @summary Day type
+ *
+ * @description
+ * Day type.
```
leshakoss, line 6: This is going to be in the docs, so I think it would be useful to explain here what day values are (i.e. 0 is always Sunday, 6 is always Saturday). Basically this: https://github.com/date-fns/date-fns/pull/2066#issuecomment-799054491

leshakoss (dismissed):

docs/Day.js
kossnocorp: Good point 👍

tan75 (approved):
Thanks a lot!

PR #2334, opened 2021-03-17
optimize isValid, remove unnessesary warnings

rifler:


package.json
```diff
@@ -23,6 +23,7 @@
     "test": "karma start config/karma.js",
     "lint": "eslint .",
     "lint-types": "eslint --config=.ts-eslintrc.js typings.d.ts",
+    "lint-prettier": "prettier --list-different './**/*.js'",
```
rifler, line 26: fyi now prettier doesn't fix typescript files, I think it is not good

src/isValid/index.ts
```diff
@@ -1,6 +1,35 @@
-import toDate from '../toDate/index'
 import requiredArgs from '../_lib/requiredArgs/index'
 
+type ValidationResult = {
+  valid: boolean;
+  type?: 'date' | 'string' | 'number';
+}
+
+export function validate(dirtyDate: any): ValidationResult {
```
rifler, line 8: maybe it is not good to implement this function in `isValid` module

PR #2339, opened 2021-03-18
Fix for package.json file generator (and ESM tree-shaking issues)

### LINKED ISSUE #2207: Tree shaking issue in 2.17
I'm using date-fns in two projects (without additional locales) and when I upgraded to 2.17 the bundle size went to ~200Kb. 

Checked with the webpack bundle analyzer, and all the functions seem to be there. I'm actually using just 7 functions with import syntax (ESM), so nothing changed from my side. If I rollback to 2.16.1, everything is OK again.

Does it have to do with the Typescript addition? My projects are plain js.

fturmel:
Hoping this fixes #2207.

None of the generated package.json files had their `module` property written because the async function `getInitialPackages()` was used synchronously and not waiting for its promise to resolve.

Before:
`{
  "sideEffects": false,
  "typings": "../typings.d.ts"
}`

After:
`{
  "sideEffects": false,
  "module": "../esm/locale/index.js",
  "typings": "../typings.d.ts"
}`

### DIFF
--- a/scripts/build/packages.js
+++ b/scripts/build/packages.js
@@ -18,18 +18,23 @@ const rootPath =
 
 const extraModules = [
   { fullPath: './src/fp/index.js' },
-  { fullPath: './src/locale/index.js' }
+  { fullPath: './src/locale/index.js' },
 ]
 
-const initialPackages = getInitialPackages()
+writePackages()
 
-listAll()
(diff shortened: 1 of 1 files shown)


kossnocorp (approved):
Thank you for that, I will ship the fix soon.

PR #2344, opened 2021-03-18
Migrate isThisWeek function

maximtop:
This pr fixes: https://github.com/date-fns/date-fns-typescriptify/issues/95


src/isThisWeek/index.ts
```diff
@@ -1,5 +1,6 @@
 import isSameWeek from '../isSameWeek/index'
 import requiredArgs from '../_lib/requiredArgs/index'
+import { LocaleOptions, WeekStartOptions } from '../types';
```
kossnocorp, line 3:
```suggestion
import { LocaleOptions, WeekStartOptions } from '../types'
```

kossnocorp (approved):

PR #2360, opened 2021-03-22
Translate min function to TS

tan75:
#### Changes
Min function translated to TS

#### Related Issue
https://github.com/date-fns/date-fns-typescriptify/issues/115


kossnocorp (approved):
👌

PR #2362, opened 2021-03-22
add de-AT locale (and typings for missing 'mn' locale)

cstenglein:
This PR adds the 'de-AT' locale (German locale for Austria).

The (only) difference to the 'de' locale is that January is not 'Januar', but 'Jänner'

In addition, the 'mn' (mongolian) typings were missing and were automatically generated by the provided script. 

EDIT: Since the checks were now failing due to the correct typing of the 'mn' locale, I fixed that too


tan75 (approved):

kossnocorp:
Thank you a lot!

PR #2365, opened 2021-03-23
Migrate de locale

maximtop:
This pr fixes: https://github.com/date-fns/date-fns-typescriptify/issues/154


src/locale/de/_lib/formatDistance/index.ts
```diff
@@ -1,4 +1,6 @@
-var formatDistanceLocale = {
+import { FormatDistanceFn } from '../../../types';
```
kossnocorp, line 1: Your Prettier seems off, please setup to automatically format the code using the repo settings.

I also suggest always using `import { FormatDistanceFn } from '../../../types'` when importing only types.

src/locale/de/_lib/formatRelative/index.ts
```diff
@@ -0,0 +1,16 @@
...
+const formatRelativeLocale = {
+  lastWeek: "'letzten' eeee 'um' p",
+  yesterday: "'gestern um' p",
+  today: "'heute um' p",
+  tomorrow: "'morgen um' p",
+  nextWeek: "eeee 'um' p",
+  other: 'P'
+}
+
+const formatRelative: FormatRelativeFn = function formatRelative(token, _date, _baseDate, _options) {
```
kossnocorp, line 12: While I prefer using `function name() {...` in this case (given that we verify `FormatRelativeFn`) it's better using `() => {...`:

```suggestion
const formatRelative: FormatRelativeFn = formatRelative(token, _date, _baseDate, _options) => {
```

src/locale/de/_lib/localize/index.ts
```diff
@@ -129,13 +131,13 @@ var formattingDayPeriodValues = {
   }
 }
 
-function ordinalNumber(dirtyNumber, _dirtyOptions) {
-  var number = Number(dirtyNumber)
+const ordinalNumber: LocalizeFn<number> = function ordinalNumber(dirtyNumber, _dirtyOptions) {
```
kossnocorp, line 134: Here too, I think it's better to not mix `() =>` and `function ()`, it's just confusing.

src/locale/de/_lib/localize/index.ts
```diff
@@ -146,7 +148,7 @@ var localize = {
   quarter: buildLocalizeFn({
     values: quarterValues,
     defaultWidth: 'wide',
-    argumentCallback: function(quarter) {
+    argumentCallback: function(quarter: unknown) {
```
kossnocorp, line 151: Can't we actually ensure that `quarter` is always a number and get rid of `Number(quarter)`?

kossnocorp: I would also modernize the function while we at it: `argumentCallback: (quarter) =>quarter - 1`

src/locale/de/_lib/match/index.ts
```diff
@@ -107,7 +108,7 @@ var match = {
     defaultMatchWidth: 'wide',
     parsePatterns: parseQuarterPatterns,
     defaultParseWidth: 'any',
-    valueCallback: function(index) {
+    valueCallback: function(index: number) {
```
kossnocorp, line 111: I hoped that `valueCallback`'s argument could be inferred from `Match`. Curious why it doesn't work?

kossnocorp (changes requested):
Great job! Please format using Prettier config, use `import type` where possible, modernize the code, simplify types, and get rid of extra type conversions (`Number()`; of course if it's possible).

src/locale/de/_lib/match/index.ts
maximtop: actually I do not know, maybe because buildMatchFn is not typed yet

src/locale/de/_lib/formatDistance/index.ts
maximtop: > I also suggest always using import { FormatDistanceFn } from '../../../types' when importing only types.

To use import type I had to update babel dependencies because babel introduced support of import type only from [version 7.9 ](https://babeljs.io/blog/2020/03/16/7.9.0#typescript-38-type-only-imports-and-exports-11171httpsgithubcombabelbabelpull11171)

maximtop:
> Great job! Please format using Prettier config, use `import type` where possible, modernize the code, simplify types, and get rid of extra type conversions (`Number()`; of course if it's possible).

Prettier generated extra load for the reviewer)

kossnocorp (approved):
Let's go!

cheeZery:
Just wanted to say that although I like the extra punctuation in German months abbreviations (e.g. "Feb." instead of "Feb.", see https://github.com/date-fns/date-fns/blob/master/src/locale/de/_lib/localize/index.ts#L22), it makes working with different languages in one code base harder because of the inconsistency.
So if you want a date with an abbrivated month, thanks to this PR working with German dates got a whole lot easier: `2021-02-01` becomes "01. Feb. 21" with `dd. MMM yy`, `2021-05-01` on the other hand results in "01. Mai 21", so without a dot. But if you work with English dates as well, it will be "01. Feb 21" and "01. May 21", both without a dot.

PR #2372, opened 2021-03-28
Migrate arMA locale

maximtop:
This pr fixes: https://github.com/date-fns/date-fns-typescriptify/issues/144

It should be merged after https://github.com/date-fns/date-fns/pull/2365 is merged


leshakoss:
Merged as a part of #2840

PR #2383, opened 2021-03-29
Round seconds

Glandos:
When using `differenceInSeconds`, the value is truncated: 1.9 second become 1 second. However, until a rounding option is introduced, it is more accurate to round the remaining seconds instead of truncating it.

### DIFF
--- a/src/intervalToDuration/index.js
+++ b/src/intervalToDuration/index.js
@@ -4,7 +4,7 @@ import differenceInMonths from '../differenceInMonths/index'
 import differenceInDays from '../differenceInDays/index'
 import differenceInHours from '../differenceInHours/index'
 import differenceInMinutes from '../differenceInMinutes/index'
-import differenceInSeconds from '../differenceInSeconds/index'
+import differenceInMilliseconds from '../differenceInMilliseconds/index'
 import isValid from '../isValid/index'
 import requiredArgs from '../_lib/requiredArgs/index'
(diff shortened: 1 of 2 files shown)


tan75 (approved):
Thank you!

kossnocorp (changes requested):
Thank you for your pull request, but I have few concerns that we should address before shipping it.

First of all, this is a breaking change, and we can't apply it like that in the current major release. Given that, could you please add `roundingMethod` with possible values `floor` (default), `ceil`, or `round`. 

Secondly, why `round` should be the default rounding method (which we might change in v3)? Could you please tell me your rationale and a little about your use case? This will help us to make an educated decision.

Glandos:
> why round should be the default rounding method (which we might change in v3)?

Well, my personal use case is to display images from CCTV based on user requested timestamp. Some images might not be exactly at the requested time, so I display the difference between the requested timestamp and the actual image timestamp. When the user requested `2021-04-02 12:42:40.000` and the image is at `2021-04-02 12:42:41.950`, the current implementation will give an interval of one second.

I would much prefer a version with (optionally) milliseconds in the interval, but the result of 1 second is more misleading than 2 seconds: the image is nearer to 2 seconds late than 1 second late.

To be honest… I totally agree that this is a breaking change. So, I'll try to make a `roundingMethod` parameter as soon as possible with the default using the current behavior (`floor`).

PR #2384, opened 2021-03-29
fix: i18n example

ctamisier:
In i18n example, shouldn't it be `window.__localeId__ = 'enGB'` instead of `window.__localeId__ = 'en'`.
`window.__localeId__ = 'en'` is working but just because it uses `enUS` as a fallback.
(like it would have been the same if we had set `window.__localeId__ = 'load-english-plz'`)


tan75 (approved):
Thank you!

PR #2391, opened 2021-04-01
Add check for 12 months to formatDistanceStrict

tan75:
#### Changes
This PR will allow `formatDistanceStrict` and `formatDistanceToNowStrict` to return **'1 year'** rather than **'12 months'** if
` dstNormalizedMinutes < MINUTES_IN_YEAR &&  months === 12`


kossnocorp (approved):
👍

Issue #2398, opened 2021-04-08
Use of flow type annotation that does not exist


khamiltonuk:
Thanks for `date-fns` 🥰 I love using it

When using `date-fns` in a code base with flow-js as a type checker
in the recent version v2.20.0 the type definitions contains a type called `unknown`
Flow js doesn't have an `unknown` type, I think that is a typescript convention but flow js does have `mixed` or `any` type.

**Before**
```
// node_modules/date-fns/index.js.flow:469
isDate: (value: any) => boolean,
````

**Now**
```
// node_modules/date-fns/index.js.flow:478
isDate: (value: unknown) => boolean,
```
I can see the change in the code diff of the new version https://github.com/date-fns/date-fns/compare/v2.19.0...v2.20.0 

I assume this was a manual change, I am not aware of a way to automate flow type definitions.

This is causing our pipeline to fail, so we can't merge
![Screenshot 2021-04-08 at 15 14 12](https://user-images.githubusercontent.com/4013283/114033906-fc876e00-987d-11eb-865b-089816340531.png)

tan75:
hi @khamiltonuk 
thank you so much for reporting this.
We will fix it ASAP

@kossnocorp 
could you please have a look into https://github.com/date-fns/date-fns/pull/2399

kossnocorp:
Fixed it in v2.20.1, thank you for reporting it.

PR #2399, opened 2021-04-08
fix: correction in flow types `unknown` is not a type in flowjs

khamiltonuk:
Fixes: https://github.com/date-fns/date-fns/issues/2398

**What has changed:**
revert change that introduced a flow type that does not exist


kossnocorp:
Thank you for the effort, but we generate Flow from JSDoc, so your change will disappear after the next build. I'll take care of it.

PR #2400, opened 2021-04-08
Fix Docs URL

tan75:
#### Changes
Fixed docs URLs in README.md as the ones were leading to 
`This page is not available for this version or this submodule`


kossnocorp (approved):
Wow 👏

PR #2407, opened 2021-04-11
Add "@babel/plugin-proposal-object-rest-spread" plugin

maximtop:
This would fix error occurring in the tests failing because "mn" locale uses spread operator.


kossnocorp (approved):
🙏

PR #2411, opened 2021-04-13
Add options.unit check

tan75:
#### Changes
Now if the distance is 12 month, it returns `1 year.`
This change will allow to force a `month` unit if specified in options.


kossnocorp (approved):
🙏

PR #2413, opened 2021-04-14
Migrate constants to Typescript

jooola:
I propose the following changes :

- Migrated the constants package to Typescript
- Grouped constants in a single file `constants.ts`
- Deprecated maxTime and minTime consants in favor of MAX_TIME and MIN_TIME in `constants.ts`


kossnocorp:
Sorry for closing the PR. Your changes conflict with ongoing work in https://github.com/date-fns/date-fns/pull/2173.

Please don't introduce breaking changes like renaming `maxTime` to `MAX_TIME` without consulting with the team.

jooola:
Yes, my bad, should have posted a proposal.

The idea is still a good one, in my opinion. A constant should be named with uppercase letters, if we follow the general conventions.

PR #2417, opened 2021-04-14
Format code

jooola:
This PR simply runs prettier on js, ts and md files.

Since the CI never enforced the prettier code formatting this PR is huge and will touch a lot of file.

I will propose a PR to add some CI jobs to enforce the code formatting is right before merging.


kossnocorp:
There's no way we can review a PR like this, but we can't blindly merge code without going through every change, so I have to close it.

Also, formatting the whole codebase will break many PR and create a lot of work for us. Formatting changed files works well for us, and there's no rush in migrating to Prettier.

PR #2422, opened 2021-04-14
Migrate _lib/format/*

jooola:
I started working on the migration of the `_lib/format/*` tools.

All the options objects are still not typed, since I need to track down what it might be.

I might have made some bold changes, so I wanted to push this here to have a little review, before I put too much effort in something wrong.

I'd be happy to have any pointers for some guidelines, codes styles that you folks agreed on ? For now I try to follow as much as I can the general coding conventions.


kossnocorp:
Hey thank you so much for your effort, but I must stop you as your changes conflict with the migration that we do in https://github.com/date-fns/date-fns/pull/2365.

jooola:
@kossnocorp No problem

Issue #2377, opened 2021-03-29
Rename firstWeekContainsDate -> firstWeekContainsNthOfJanuary


leshakoss:
Nth of January is easier to understand. However it is a breaking change (and so we can only ship it in v3).
Is it worth it?

jooola:
I'm still not sure to understand what firstWeekContainsDate or firstWeekContainsNthOfJanuary means, where is it used ?

Could you explain it, it will give a better idea of how to name it ?!

jooola:
Never mind, just found this in the code https://en.wikipedia.org/wiki/Week#Week_numbering

We need to add the link to Wikipeda in the typescript `FirstWeekContainsDateOptions` definition, so everyone can open the link from there IDE.

Regarding the name, I would first replace Date with Day, since we are talking about a day in the week, also replace Week with WeekNumber.

So maybe something like firstWeekNumberContainsDay ?

PR #2427, opened 2021-04-16
Add comment on FirstWeekContainsDate type

jooola:
Since I've been confused by this type in my last PR, I propose to add a comment to give some more context and explanation about it. See https://github.com/date-fns/date-fns/pull/2318#issuecomment-820729615

Also it seems the field is not easy to understand without a proper explanation from Wikipedia. See #2377


tan75 (approved):
Thank you!

Issue #732, opened 2018-04-23
Unit conversion functions


ctf0:
is it possible to have like a helper function that can convert a date to another ?

ex.
```js
convert(2, 'hrs', 'milli')
convert(1, 'month', 'hrs')
etc...
```

edorivai:
What is it exactly that you want to accomplish? In case you want to convert durations, this can be done with math!

```js
// 2 hours in milliseconds
2 * 60 * 60 * 1000; // 7200000

// 1 month in hours (approximately, because it depends on your definition of the length of a month)
const averageDaysInMonth = 365.25 / 12;
1 * averageDaysInMonth * 24; // 730.5
```

ctf0:
exactly that but for ppl that suck in math like myself, the **helper** will the perfect answer

kossnocorp:
This is a good idea, however, I don't like the proposed API. I suggest making it explicit, like `secondsToMinutes(45)`, `hoursToDays(123)`, etc.

PRs are welcome.

edorivai:
Oh my, should this be a cross map of all different time units? Also, we should probably make a table of definitions. Like how many days is a month? A year? Do we do gap years?

Lakston:
I can try do do something for this request but we would have to define the API and what conversions we would plan to support because some can get a bit tricky as mentioned by @edorivai , if you want to convert minutes to months, would you take the current month length ? allow for a month argument (that would not fit with a simple, explicit API) ?

The easiest ones to start with would be : 

- seconds => minutes
- seconds => hours
- seconds => days
- minutes => seconds
- minutes => hours
- minutes => days

When getting into months / years territory it gets more complicated with months duration and gap years.

Maybe an optional argument such as `minutesToYears(15962, true)` where `true` defines if the year is a gap year or not and `secondsToMonths(15472, 30)` where `30` would be an optional month duration.

ctf0:
imho using the explicit api will be hard to maintain while using the one from the main topic is easier as it can be done using switches.

on -a-side-note apparently this is hard to create because of the indifferences in month days & year months

leshakoss:
For the record, without any ambiguity it is possible to convert between these groups of units:
- milliseconds ↔ seconds ↔ minutes ↔ hours
- days ↔ weeks
- months ↔ quarters ↔ years

edorivai:
@leshakoss hours :left_right_arrow: days?

edorivai:
Damn, you're right

[...]

PR #2433, opened 2021-04-20
creating unit conversions functions

LucasHFS:
Attempt to solve #732 . This is my first PR to the lib, any changes needed let me know.


src/daysToWeeks/test.ts
```diff
@@ -0,0 +1,12 @@
+// @flow
+/* eslint-env mocha */
+
+import assert from 'assert'
+import daysToWeeks from '.'
+
+describe('daysToWeeks', function() {
+  it('converts days to weeks', function() {
+    const result = daysToWeeks(14);
```
wenderjean, line 9: You could exercise your function a bit more I mean, giving it different values like `5`, `13`, `8`

tan75:
@LucasHFS 
Thanks a lot for your work!
Could you please fix all the examples in your files as per this commit below: 
[Commit](https://github.com/date-fns/date-fns/pull/2433/files#diff-91ef18f1c7c67a10134478750473345429e9746a3ec075f0d07ce69ad107c83aR16-R19)

As well, could you please add more tests to your test files like 
` daysToWeeks(13);` - see the comment from @wenderjean 
From my point of view, I would expect the result of the above call to be `1`

tan75 (changes requested):
Thank you for your PR! Please, address my comments.

LucasHFS:
> From my point of view, I would expect the result of the above call to be `1`

@tan75 I get it,  can I use this logic in similar situations, e.g. minutesToHours & monthsToYears?

tan75:
> > From my point of view, I would expect the result of the above call to be `1`
> 
> @tan75 I get it, can I use this logic in similar situations, e.g. minutesToHours & monthsToYears?

Yes, you can. 
I think it makes sense.

src/daysToWeeks/test.ts
LucasHFS: Done. 😃

tan75 (approved):
Many thanks for your PR!

src/millisecondsToMinutes/test.ts
```diff
@@ -0,0 +1,21 @@
...
+  it('uses floor rounding', () => {
+    assert(millisecondsToMinutes(60001) === 1)
+    assert(millisecondsToMinutes(59999) === 0)
+  })
+
+  it('handles border values', () => {
+    assert(millisecondsToMinutes(60000.5) === 1)
+    assert(millisecondsToMinutes(0) === 0)
+  })
+})
```
kossnocorp, line 21: Please take a look at this test file. I want you to rewrite tests in the PR to make them look like this.

Instead of focusing on what test does `converts 700 milliseconds to seconds` I want you describe behavior and test it. It's enough to have a couple of similar examples that test the same thing:

```ts
assert(millisecondsToMinutes(60000) === 1)
assert(millisecondsToMinutes(120000) === 2)
```

Also, when a function does rounding, I want you to add an example for that (please make sure that the first test doesn't check for rounding).

And finally, test border values (float and zero).

kossnocorp (changes requested):
Thank you so much for your great work! 

The only problem I found in the PR is that test examples focus on what a test does instead of the function behavior. Please rewrite those using the example I added for `millisecondsToMinutes`.

Please rewrite tests

tan75:
hi @LucasHFS 

Hope all is well. Just wondering if you are still working on this PR?

LucasHFS:
@tan75 Yep, I'll send the requested changes today.

kossnocorp (approved):

PR #2459, opened 2021-05-01
Fix docs generation perf issue by limiting processes to machine cpu count

fturmel:
I never managed to successfully run the `build.sh` script on my machine before, it always completely destroyed my MacBook Air M1 16GB to the point where I had to force quit VS Code. CPU, RAM were pegged and it was swapping like crazy.

After taking a closer look, it looks like the jsdocs to markdown task is spawning 207 node processes at the same time!

Here’s a proposed fix which simply limits the amount of async tasks to the number of cores on the machine. In my case, I can see that the CPU usage looks optimal without making the computer unresponsive, RAM usage is stable and there is no swapping. The `docs.js` execution actually completes in about 50 seconds.

While editing this file, I also replaced the `fs-promise` package (deprecated) with the native `fs/promises`. It was only used here, so I removed it from the project's dev dependencies.


fturmel:
More supporting evidence of the problem, after over 4 minutes running `build.sh`, my computer is dying and we're not anywhere close to being done with the first step (in fact it never seems to be able to finish and I have to force quit VS Code).

Screen Shot 2021-05-13 at 12 18 02 PM

With this PR, `build.sh` executes completely in under 53 seconds and memory usage is stable.

Screen Shot 2021-05-13 at 12 31 04 PM

Screen Shot 2021-05-13 at 12 25 00 PM

kossnocorp (approved):
Thank you so much for debugging and fixing it! I'm curious how it managed to work on my machine before, but somehow I shipped quite a few releases with this version of the script. It's worth noting that I worked with it on PC and never tested it on my MacBook.

fturmel:
Interesting, maybe it's an architecture thing or you have more RAM on that machine. Happy to help.

PR #2476, opened 2021-05-16
Lower tsc error count

fturmel:
This PR lowers the error count of the `yarn tsc --noEmit` command from 54 to 16 (all from test.ts files).

14 of the remaining 16 errors are caused by `lint-staged v7`'s dependency tree bringing in `date-fns` itself: `lint-staged -> listr -> listr-verbose-renderer -> date-fns:"^1.27.2"`. The duplicate type definitions clash with the project's `typings.d.ts`. Could that also be potentially dangerous? (ex: accidental wrong import path in a unit test might silently run against an older version of the library)

I can think of two approaches to resolve this, would love to get feedback so I can put together a follow-up PR. I tend to prefer the first one.

1. Bump `lint-staged` to the latest version, which would get rid of that `date-fns` nested dependency altogether
2. Add `skipLibCheck` to `tsconfig.json`. This could be suppressing important type feedback though, see https://www.typescriptlang.org/tsconfig#skipLibCheck

### DIFF
--- a/src/areIntervalsOverlapping/test.ts
+++ b/src/areIntervalsOverlapping/test.ts
@@ -3,7 +3,6 @@
 
 import assert from 'assert'
 import areIntervalsOverlapping from '.'
-import context from 'assert'
 
 describe('areIntervalsOverlapping', function () {
   const initialIntervalStart = new Date(2016, 10, 10, 13, 0, 0)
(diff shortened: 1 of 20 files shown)


fturmel:
I thought of a 3rd option this morning, migrating to `pnpm` would resolve these potential phantom dependency problems altogether. https://pnpm.io/motivation#creating-a-non-flat-node_modules-directory

I tested it quickly and not only does it remove these tsc errors, but it surfaced that `tslib` and `@babel/plugin-proposal-class-properties` are currently phantom dependencies in the project. It also pointed out that `babel-eslint` has been deprecated and was moved to `@babel/eslint-parser`.

fturmel:
Going to close this PR, #2901 will cover the same grounds and much more. Also #2899 would get rid of the transitive dep mentioned if merged.

PR #2493, opened 2021-05-28
typings issue fixed.

developer0794:
Typings issue fixed.
Removed the duplicate keys from the typings.d.ts.
i,e., daysInWeek, minTime.


developer0794:
Already fixed.

PR #2495, opened 2021-05-28
Add Bosnian locale

BranislavLazic:
This PR adds Bosnian locale and resolves this issue https://github.com/date-fns/date-fns/issues/2435.


src/locale/bs/_lib/formatRelative/index.js
```diff
@@ -0,0 +1,44 @@
...
+  },
+  yesterday: "'juče u' p",
+  today: "'danas u' p",
+  tomorrow: "'sutra u' p",
+  nextWeek: function (date) {
+    var day = date.getUTCDay()
+
+    switch (day) {
+      case 0:
+        return "'sledeće nedjelje u' p"
```
sake92, line 24: I think that `sljedeće` is correct one here.

sake92: Also `sedmica` instead of `nedjelja`.

BranislavLazic: Fixed. Thank you!

src/locale/bs/index.js
```diff
@@ -0,0 +1,28 @@
...
+var locale = {
+  code: 'bs',
+  formatDistance: formatDistance,
+  formatLong: formatLong,
+  formatRelative: formatRelative,
+  localize: localize,
+  match: match,
+  options: {
+    weekStartsOn: 1 /* Monday */,
+    firstWeekContainsDate: 1,
```
tan75, line 24: hi @BranislavLazic 
Could you please provide a reference that confirms the value of 1 of `firstWeekContainsDate`? 
According to Wikipedia, it is 4 for the majority of European countries. 
Do you possibly know a source that we can refer to?

tan75 (changes requested):

src/locale/bs/index.js
BranislavLazic: Seems like you are right. Also, some of them share the same flaw. E.g. sr-Latn (which was my reference locale) or hr.

tan75:
@BranislavLazic 
could you please re-generate snapshots by running `yarn locale-snapshots` and commit the results?

tan75 (changes requested):
Please run `yarn locale-snapshots` and commit the results.

tan75 (approved):
Much appreciated! 👍  👍  
I will schedule this for our next release given the go-ahead by @kossnocorp

PR #2496, opened 2021-06-03
Migrate English locales to TS

kossnocorp:
I've migrated all English locales, including `en-US`.


src/locale/en-US/_lib/localize/index.ts
```diff
@@ -124,11 +126,11 @@ var formattingDayPeriodValues = {
...
     evening: 'in the evening',
-    night: 'at night'
-  }
+    night: 'at night',
+  },
 }
 
-function ordinalNumber(dirtyNumber, _dirtyOptions) {
+const ordinalNumber: LocalizeFn<number, undefined> = (dirtyNumber) => {
   var number = Number(dirtyNumber)
```
maximtop, line 134: var -> const

maximtop, line 134:
```suggestion
  const number = Number(dirtyNumber)
```

src/locale/en-US/_lib/formatLong/index.ts
```diff
@@ -1,41 +1,42 @@
...
 
 var timeFormats = {
   full: 'h:mm:ss a zzzz',
   long: 'h:mm:ss a z',
   medium: 'h:mm:ss a',
-  short: 'h:mm a'
+  short: 'h:mm a',
 }
 
 var dateTimeFormats = {
```
maximtop, line 18:
```suggestion
const dateTimeFormats = {
```

src/locale/en-US/_lib/formatLong/index.ts
```diff
@@ -1,41 +1,42 @@
...
-var dateFormats = {
+const dateFormats = {
   full: 'EEEE, MMMM do, y',
   long: 'MMMM do, y',
   medium: 'MMM d, y',
-  short: 'MM/dd/yyyy'
+  short: 'MM/dd/yyyy',
 }
 
 var timeFormats = {
```
maximtop, line 11:
```suggestion
const timeFormats = {
```

maximtop (dismissed):
Looks good to me.
There are left few var declarations instead of const in the code.
Also, tests are red.
Update of jest should resolve this issue

fturmel:
@tan75 I took a look at the failing tests. PR #2608

This is the offending function, it always seems to return undefined: https://github.com/date-fns/date-fns/blob/e5fa99340c846bbe37b8a4d586f4575356d7097d/src/intlFormat/test.ts#L20-L33

Why did passing an undefined locale not make Intl calls fail before? That's pretty puzzling to me... Maybe it only broke for Node 14? That would be the blind spot for me because it works locally on 16.8 and the CI's previous tests were covering Node 8-12.

fturmel:
I was curious and did some logs during the CI tests, it's not related to undefined return values after all.

`getOperationSystemLocale()` returned 'C' because `process.env.LANG` is 'C.UTF-8' in that environment.

```ts
new Intl.DateTimeFormat('C').format(new Date())
// Uncaught RangeError: Incorrect locale information provided
```

PR #2498, opened 2021-06-04
Add clamp function

Liam-Tait:
Add a function to clamp a date based on an interval.

This function ensures that a date is bounded by a minimum and maximum


Liam-Tait:
For the curried version of this function.  Currying from the right makes sense for my use case. I have an interval that I want to clamp multiple values to.

```typescript
const contractInterval = {start, end}
const clampToContract =  clamp(allowed)

const queryStartDate = new Date();
const queryEndDate = new Date();

const start = clampToContract(queryStartDate)

```

Liam-Tait:
Sorry @tan75, I accidentally closed the PR

src/clamp/index.ts
```diff
@@ -0,0 +1,46 @@
...
+ * When date is less than start, start is returned
+ * When date is greater than end, end is returned
+ * Otherwise the date is returned
+ *
+ *
+ * Bounds the date
+ *
+ * @example
+ * ```javascript
+ * var interval = {
```
tan75, line 22: Could you please change `var` to `const` in your examples?

tan75 (changes requested):
Hi @Liam-Tait
this function is a great idea and it might be very useful! Thank you for your PR! 👍 
Could you please make the examples consistent with our current example convention - e.g. see this one below:
[Next Monday example](https://github.com/date-fns/date-fns/blob/master/src/nextMonday/index.ts#L18-L21)
Now the convention is to use `const` and please delete the 'javascript' mention.
As well, please format the examples and the 'Docs' part of the file - it's good to have them consistent with the `nextMonday` example. Let us know if anything is unclear.

tan75:
@Liam-Tait 
could you please provide any use cases for `clamp`?

src/clamp/index.ts
Liam-Tait: Can do.
Is it preferable to write the example as typescript?

tan75: > Can do.
> Is it preferable to write the example as typescript?

No, JS is ok in case if a user doesn't use TS.

Liam-Tait: Resolved with 68d41b2

Liam-Tait:
> Hi @Liam-Tait
> this function is a great idea and it might be very useful! Thank you for your PR! 👍

Thanks!

> Could you please make the examples consistent with our current example convention - e.g. see this one below:
> [Next Monday example](https://github.com/date-fns/date-fns/blob/master/src/nextMonday/index.ts#L18-L21)
> Now the convention is to use `const` and please delete the 'javascript' mention.
> As well, please format the examples and the 'Docs' part of the file - it's good to have them consistent with the `nextMonday` example. Let us know if anything is unclear.

I think I have matched the convention. Let me know if this correct

- use `const` instead of `var`
- removed javascript mention
- removed backtick code section
- inline parameters instead of using seperate variables
- Add comment for the question the function is answering
- Add comment for the returned value

Liam-Tait:
> @Liam-Tait
> could you please provide any use cases for `clamp`?

My use case is for querying a list of data based on a contract. 

The query has a minDate and maxDate
The contract has a start and end
The data has a date 

I only want to retrieve the data within the period of a contract.
In order to do this i want to clamp the query minDate  and max Date to the contract interval.

Ideally this would be done at the database layer, however that option is not available to me.
```javascript
const minDate = clamp(requestedMin, contractInterval)
const maxDate = clamp(requestedMax, contractInterval)
const result = query(minDate, maxDate)
```

---
Useful for when there is an allowed period of time (interval) and a user input of date and instead of causing an error bounding to the allowed range

Liam-Tait:
Updated! Please take a look

tan75 (approved):
@Liam-Tait Thank you for this PR!  👍 
Assigned to @kossnocorp for release

Issue #2078, opened 2020-11-11
Month and DayOfWeek constants/enums


mfulton26:
instead of writing

```js
import { format } from 'date-fns'

format(new Date(2014, 1, 11), 'MM/dd/yyyy')
//=> '02/11/2014'
```

I would like to be able to write

```js
import { format, Month } from 'date-fns'

format(new Date(2014, Month.February, 11), 'MM/dd/yyyy')
//=> '02/11/2014'
```

and instead of writing

```js
import { startOfWeek } from 'date-fns'

// If the week starts on Monday, the start of the week for 2 September 2014 11:55:00:
startOfWeek(new Date(2014, 8, 2, 11, 55, 0), { weekStartsOn: 1 })
//=> Mon Sep 01 2014 00:00:00
```

I would like to be able to write

```js
import { DayOfWeek, Month, startOfWeek } from 'date-fns'

// If the week starts on Monday, the start of the week for 2 September 2014 11:55:00:
startOfWeek(new Date(2014, Month.September, 2, 11, 55, 0), { weekStartsOn: DayOfWeek.Monday })
//=> Mon Sep 01 2014 00:00:00
```

I know these may be specific to the Gregorian calendar but is there some way to include them in `date-fns` somehow? If not, is there some well adopted library out there that already defines these that I am overlooking?

Thank you.

mfulton26:
I've defined these enums myself in some projects and it isn't hard but it would be nice to have them already included in `date-fns` and not have to redefine them or create my own package, etc.

PR #2506, opened 2021-06-07
adding month enum

LucasHFS:
attempt to solve #2078 , any suggestions on how should I implement this?


fturmel:
My 2 cents (caveat: not a maintainer)

I like the idea for these personally, even if it's minor syntactic sugar. I noticed a lot of the internal tests have inlined comments to identify months more easily, so this is one scenario that could see a slight improvement in adopting enums instead.

ex:
https://github.com/date-fns/date-fns/blob/ea3f05069605388af0bb6a27f8cf4539f3d70db6/src/add/test.ts#L22-L25

Some thoughts on implementation:
- I think this should go in the constants file instead of an additional package
- Depending on what it transpiles too, you could probably just make these actual TypeScript enums. They are zero-based auto-incremented integers by default too so even simpler.
- The JSDocs type should be either enum of numbers, or object with string keys and number values
- Unit tests for constant values don't serve any purpose imo, so I would remove that
- Up to debate wether a 3 letter abbreviation could be prefered over the full word for conciseness (`Month.Sep` vs `Month.September`)
- Not in your PR (yet?), but per #2078 should the week day enum be called `DayOfWeek` or simply `Day` to match the internal type already established? I like `DayOfWeek`, but there is a precedent in the codebase to consider.

https://github.com/date-fns/date-fns/blob/ea3f05069605388af0bb6a27f8cf4539f3d70db6/src/types.ts#L22-L24

https://github.com/date-fns/date-fns/blob/ea3f05069605388af0bb6a27f8cf4539f3d70db6/src/types.ts#L38-L40

fturmel:
@LucasHFS Cool, no problem.

Yeah, objects are probably fine. That enum error is a false positive from eslint. The project's linting approach needs a revamp now that there is more and more TypeScript code. I'm planning on looking into it.

One more thing to look at in your PR: your day of week numbering sequence is wrong, 0 should represent Sunday not Monday. You can see an example in the docs here: https://date-fns.org/docs/getWeeksInMonth

LucasHFS:
@fturmel fixed.

fturmel:
I've identified a blocker just now... The build step generates TypeScript and Flow types based on jsdocs. It doesn't know what to do with objects or enums in the constants file, so it will emit broken definitions.

This PR should be on hold until it can address this type generation hurdle, or until the date-fns project finishes its TypeScript migration and uses `tsc` to generate the type definitions per the migration plan #1932

src/constants/index.ts
```diff
@@ -1,3 +1,45 @@
...
+  Apr : 3,
+  May : 4,
+  Jun : 5,
+  Jul : 6,
+  Aug : 7,
+  Sep : 8,
+  Oct : 9,
+  Nov : 10,
+  Dec : 11,
+}
```
Liam-Tait, line 23: Suggest adding a const assertion here and for DayofWeek

```suggestion
} as const
```

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

Issue #2421, opened 2021-04-14
Add roundingMethod parameter in differenceInSeconds


tan75:
See 
[PR 2383](https://github.com/date-fns/date-fns/pull/2383)

PR #2511, opened 2021-06-09
adding roundingMethod parameter in differenceInSeconds

LucasHFS:
attempt to implement #2421 . I made it using the floor as roundingMethod default.

### DIFF
--- a/src/differenceInSeconds/index.ts
+++ b/src/differenceInSeconds/index.ts
@@ -21,18 +21,32 @@ import requiredArgs from '../_lib/requiredArgs/index'
  * @example
  * // How many seconds are between
  * // 2 July 2014 12:30:07.999 and 2 July 2014 12:30:20.000?
+ * // with default floor roundingMethod
  * const result = differenceInSeconds(
  *   new Date(2014, 6, 2, 12, 30, 20, 0),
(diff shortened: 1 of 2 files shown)


src/differenceInSeconds/index.ts
```diff
@@ -21,18 +21,32 @@ import requiredArgs from '../_lib/requiredArgs/index'
...
+    roundingMethodFn = Math.floor
+  } else if (roundingMethod === 'ceil') {
+    roundingMethodFn = Math.ceil
+  } else if (roundingMethod === 'round') {
+    roundingMethodFn = Math.round
+  } else {
+    throw new RangeError("roundingMethod must be 'floor', 'ceil' or 'round'")
+  }
+  return roundingMethodFn(diff);
+}
```
pabrodez, line 52: What do you think of doing this to look up the method instead?

```javascript
const roundingMethodFns = {
  'floor': Math.floor,
  'ceil': Math.ceil,
  'round': Math.round
}

if (!roundingMethodFns.hasOwnProperty(roundingMethod)) throw new RangeError("roundingMethod must be 'floor', 'ceil' or 'round'");

return roundingMethodFns[roundingMethod](diff);
```

PR #2515, opened 2021-06-14
Fix - add and subtract when duration has undefined params

komyg:
Hi,

I've been working with `date-fns`  and I saw that the `add` and `subtract` functions fail when they receive a `Duration` argument that contains one or more params that were set explicitly as `undefined`.

Below is an example
```ts
import { add, sub } from 'date-fns';

const now = new Date();
const tomorrow = add(now, { days: 1, months: undefined }); // Tomorrow is invalid date
const yesterday = sub(now, { days: 1, months: undefined }); // Yesterday is invalid date
```

This happens, because of this check:

```ts
const months = 'months' in duration ? toInteger(duration.months) : 0
```

And since `months` is inside the `duration`, but is not an integer, the function fails.

I've created a PR with a proposed fix and some unit tests to validate that everything is working.

Thanks,
Komyg

### DIFF
--- a/src/add/index.ts
+++ b/src/add/index.ts
@@ -20,7 +20,7 @@ import { Duration } from '../types'
  * |----------------|------------------------------------|
  * | years          | Amount of years to be added        |
  * | months         | Amount of months to be added       |
- * | weeks          | Amount of weeks to be added       |
+ * | weeks          | Amount of weeks to be added        |
  * | days           | Amount of days to be added         |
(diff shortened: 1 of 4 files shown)


tan75 (approved):
Thank you for this PR! Much appreciated! 👍

komyg:
Hi @tan75, could we merge this PR? I am not authorized to do so.

Thanks,
Komyg

PR #2523, opened 2021-06-19
Convert fromUnixTime function to TypeScript

fturmel:
Also added a note to jsdocs about timestamp format being in seconds (see #2206)


fturmel:
I just noticed there is another similar open PR #2317, commenting to link them.

PR #2522, opened 2021-06-18
adding previousDay and its variations

LucasHFS:
[WIP] Implements #2507 request, and similar functions.

PR #2524, opened 2021-06-19
Improve nextDay implementation

fturmel:
This is roughly 50% faster than the current implementation while being simpler. Also see my comment on PR #2522 that implements `previousDay`.


tan75 (approved):
Wow, that's great!
Thanks for this PR!  :)  🥇

fturmel:
I realized we don't need to convert the date inside `nextDay`, which gives us a bit more perf.

Benchmark on local machine:

| Version        | ops/sec           |
| ------------- |------------:|
| revised PR      | 2,048,456 |
| initial PR      | 1,792,947 |
| current | 1,306,011 |

PR #2525, opened 2021-06-20
Generated fp and indices TypeScript conversion

fturmel:
(Sorry for the large amount of files modified here, it's simpler than it looks at first glance.)

This PR is essentially making the build system (fp.js and indices.js) output TypeScript files instead of JS, with a few additional related TypeScript conversions.

Overview:

- required changes to `fp.js` and `indices.js` + adding 'index.ts' and 'test.ts' filenames to a few ignore arrays
- `fp/test.ts` converted to TS
- `fp/_lib/convertToFP` internal function and its test file converted to TS
- the 3 indices generated by `indices.js` are now TS files, no content change (`index.ts`, `fp/index.ts`, `locale/index.ts`)
- 200+ fp index files converted to TS and simplified a bit (all generated by `fp.js` script). It's all the same pattern, so hopefully not too hard to quickly review.

All tests are passing, full build and package steps run fine. I think we should be good. 🤞


leshakoss (approved):
Awesome!

PR #2527, opened 2021-06-22
Allow flattening overflowing duration during format

paroxp:
## What

Allow the `formatDuration`, to convert overflowing values into its higher units.

`90 seconds` -> `1 minute 30 seconds`
`3024075 seconds` -> `1 month 5 days 1 minute 15 seconds`


Lemmmy:
Is there any pre-existing function to achieve this before this is merged?

Issue #2083, opened 2020-11-14
"Format string contains an unescaped latin alphabet character `v`" using "formatRelative" and "sk" locale


MartinMikusat:
Here's a CodeSandbox with the issue
https://codesandbox.io/s/date-fns-bug-qf1db?file=/src/index.js

Doesn't happen with all dates, only some. I wasn't able to find any clues to what exactly is causing this.

fturmel:
Hi, I just stumbled on your issue.

This definitely is a bug. This error message comes from the `format` method. I tracked it down to format string in the sk locale that is not escaped properly.

nextWeek + `štvrtok` string would throw the error in your example. Should be an easy fix, I'll work on a PR.

https://github.com/date-fns/date-fns/blob/ea3f05069605388af0bb6a27f8cf4539f3d70db6/src/locale/sk/_lib/formatRelative/index.js#L44

https://github.com/date-fns/date-fns/blob/ea3f05069605388af0bb6a27f8cf4539f3d70db6/src/locale/sk/_lib/formatRelative/index.js#L4-L12

PR #2540, opened 2021-06-29
Fix unescaped formating string in sk locale

fturmel:
See #2083, there is an edge case formatting in the sk locale that can return an improperly escaped string and cause a runtime error.


tan75 (approved):
Thank you!

PR #2545, opened 2021-07-04
Deno-related improvements and fixes

fturmel:
This brings a number of improvements and fixes related to the new Deno package build, which is partially broken at the moment. See https://github.com/date-fns/deno/pull/2

## Fixes
- ~remove the broken locales from output~ (not relevant anymore)
- remove typings.d.ts and *.d.ts from output (they are incompatible with Deno when referenced using `@deno-types`)
- addDenoExtensions build:
  - correctly resolve the file path for imports that reference a directory instead of a file
  - stricter path replace to avoid transforming an import multiple times (ex: `../../types.ts` and `../../../types.ts` in same file)
  - script will now fail if it can't locate a file for an import/export path
  - early exit check in core loop was always a miss because it was using `importPath` instead of `fullPath`
  - minor optimizations
- fix all bad imports in the project (missing types, typos)

## Tests
- added a Deno test which simple runs a few assertions and imports the whole library
- added a Deno github action that builds and tests the Deno package on pull requests/push
- Jest config ignore `test/deno` path

### DIFF
--- a/scripts/build/_lib/addDenoExtensions.ts
+++ b/scripts/build/_lib/addDenoExtensions.ts
@@ -1,25 +1,20 @@
-import { resolve, dirname } from 'path'
-import fs from 'fs'
+import { existsSync } from 'fs'
+import { readFile, stat, writeFile } from 'fs/promises'
 import globby from 'globby'
+import { dirname, resolve } from 'path'
 import ts, {
   ExportDeclaration,
   ImportDeclaration,
   StringLiteral,
 } from 'typescript'
 
-const { readFile, writeFile, stat } = fs.promises
-
-const pattern = /\.(ts|js)$/
-const ignore = [/\.d\.ts$/]
+// *.ts and *.js, but NOT *.d.ts files
+const pattern = /(?<!\.d)\.(ts|js)$/
 
 const resolvedExtensions: Record<string, string> = {}
 
 globby('deno')
-  .then((files) =>
-    files.filter(
-      (file) => pattern.test(file) && !ignore.find((p) => p.test(file))
-    )
-  )
+  .then((files) => files.filter((file) => pattern.test(file)))
   .then((files) =>
     Promise.all(
       files.map((file) =>
@@ -48,14 +43,33 @@ globby('deno')
 
           await Promise.all(
             imports.map(async (importPath) => {
-              if (resolvedExtensions[importPath]) return
               const fullPath = resolveFullPath(file, importPath)
-              let isTs = false
-              try {
-                await stat(fullPath + '.ts')
-                isTs = true
-              } catch (_) {}
-              resolvedExtensions[fullPath] = isTs ? '.ts' : '.js'
+
+              if (resolvedExtensions[fullPath]) return
+
+              // identify directory imports that should resolve to their index file in Deno
+              let isDirectory = false
+              if (existsSync(fullPath)) {
+                const st = await stat(fullPath)
+                isDirectory = st.isDirectory()
+              }
+
+              const suffixes = isDirectory
+                ? ['/index.ts', '/index.js']
+                : ['.ts', '.js']
+
+              for (const suffix of suffixes) {
+                if (existsSync(fullPath + suffix)) {
+                  resolvedExtensions[fullPath] = suffix
+                  break
+                }
+              }
+
+              if (!resolvedExtensions[fullPath]) {
+                console.error("Can't locate import path:", fullPath)
+                console.error('In file:', file)
+                process.exit(1)
+              }
             })
           )
 
@@ -63,8 +77,10 @@ globby('deno')
             file,
             imports.reduce((acc, importPath) => {
               const fullPath = resolveFullPath(file, importPath)
+
+              // only replace when path is surrounded by single or double quotes
               return acc.replace(
-                new RegExp(importPath, 'g'),
+                new RegExp('(?<=\'|")' + importPath + '(?=\'|")', 'g'),
                 importPath + resolvedExtensions[fullPath]
               )
             }, content)
--- a/.github/workflows/deno_tests.yaml
+++ b/.github/workflows/deno_tests.yaml
@@ -1,0 +1,36 @@
+name: Deno tests
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - '**'
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+
+    strategy:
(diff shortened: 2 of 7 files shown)


tan75:
@fturmel 
thank you! @kossnocorp can you please have a look?

scripts/build/_lib/addDenoExtensions.ts
```diff
@@ -1,25 +1,20 @@
...
   ExportDeclaration,
   ImportDeclaration,
   StringLiteral,
 } from 'typescript'
 
-const { readFile, writeFile, stat } = fs.promises
-
-const pattern = /\.(ts|js)$/
-const ignore = [/\.d\.ts$/]
+// *.ts and *.js, but NOT *.d.ts files
```
andykais, line 11: why are we excluding `.d.ts` files?

fturmel: They were already excluded by this import transformation script, I simply rewrote the matching logic as a single regex instead of two.

However, I _did_ end up excluding all *.d.ts file from the rsync step, see `scripts/build/deno.sh`.

Based on my tests working on this PR, the `typings.d.ts` file is not in a usable shape for Deno when you reference it with a type directive (`// @deno-types="../../deno/typings.d.ts"`). All other .d.ts in the project rely on module declarations from that file.

Import and export statements inside ambient modules seem to be the problem. I think the approach used by date-fns to reuse function type declarations may not be allowed in Deno applications.

Here's a simple example with error messages. I can't figure out a way to make it work. Any insights?

```ts
declare module 'date-fns' {
  function addDays(date: Date | number, amount: number): Date
  namespace addDays {}
}

declare module 'date-fns/addDays' {
  import { addDays } from 'date-fns' // <- error: Relative import path "date-fns" not prefixed with / or ./ or ../
  export default addDays
}
```

```ts
declare module 'index.js' {
  function addDays(date: Date | number, amount: number): Date
  namespace addDays {}
}

declare module 'addDays/index.js' {
  import { addDays } from 'index.js' // <- error: Relative import path "index.js" not prefixed with / or ./ or ../ 
  export default addDays
}
```

```ts
declare module 'index.js' {
  function addDays(date: Date | number, amount: number): Date
  namespace addDays {}
}

declare module 'addDays/index.js' {
  import { addDays } from './index.js' // <- error: TS2439 [ERROR]: Import or export declaration in an ambient module declaration cannot reference module through relative module name.
  export default addDays
}
```

vizcay:
@fturmel any chance to use this from deno without waiting to be merged?

fturmel:
@vizcay not directly from a CDN, but you could clone my branch, build the Deno files and copy them to your app temporarily if you can't wait for an official release.

Here are the steps to follow (assuming you have Node and Deno installed on that machine):

```bash
# clone & checkout the PR branch
git clone git@github.com:fturmel/date-fns.git
cd date-fns
git checkout PR/deno-fixes

# build & test Deno version
yarn install
./scripts/build/deno.sh
./scripts/test/deno.sh
```

If the build and test shell scripts run successfully, you can copy the content of the `date-fns/deno` directory in your project and import the lib with a relative path. Here's a quick example:

```ts
import * as dateFns from './date-fns-deno/index.js';

const res = dateFns.parseISO('2010-03-14', {});
console.log(res); // 2010-03-14T05:00:00.000Z
```

fturmel:
Also, depending on which date-fns functions you want to use, you might still be able to use the published version on deno.land. You can't bulk-import the whole library for now but can still use the functions that are not missing import statements if you import them individually.

Every module entry point is an `index.js` or `index.ts` file inside a directory that matches the function name. You can conveniently browse the file system structure on the deno site if you need to verify the correct path to use: https://deno.land/x/date_fns@v2.22.1

```ts
import parseISO from 'https://deno.land/x/date_fns@v2.22.1/parseISO/index.js';

const res = parseISO('2010-03-14', {});
console.log(res); // 2010-03-14T05:00:00.000Z
```

fturmel:
@kossnocorp @leshakoss @tan75 Anything else I can do to make this PR go through?

Was there a change of heart in supporting Deno altogether? (last published release was 2.22.1, and it looks like this PR was expected to be merged in 2.23 and 2.24).

fturmel:
Rebased and added some additional fixes to address recent library changes (mostly type and import issues). The PR should be in a mergeable state again.

fturmel:
@tan75 saw that you self-assigned - I'm working on a rebase and fixes following all the things that got merged to master recently. Hold on a bit.

fturmel:
@tan75 OK, we're in a happy place again.

It's astounding the amount of type issues that snuck in. The eslint configuration in the project is not adequate at all.

This Deno smoke test is catching a lot of things, which is nice. But these errors should surface more naturally with the linter. I might have a look later this week and work on a PR.

I think one of the reasons these missing imports are happening is because the `date-fns` project itself has `date-fns` v1 in its deps through lint-staged (pointed out in #2476) and the ambient type defs of v1 are leaking. That seems like such a dangerous setup, this transitive dependency should really be removed.

Screen Shot 2021-12-28 at 2 14 54 PM

fturmel:
Another thing, I think it would make sense that the `/scripts/build/deno.sh` script publish to /tmp/deno by default instead of /deno. That way it would match the `/scripts/build/package.sh` behavior a bit more closely. Maybe it also needs a destination output cli arg.

tan75:
hi @fturmel 
thanks a lot for all your amazing work that you have put in Deno version but we are currently not sure if we are going to continue supporting it at all as it looks like a lot of extra work involved when it comes to release and support. 

I have self-assigned this as I am going to discuss this with @kossnocorp and @leshakoss again but this will happen in late January. Until then, I am not sure what will happen with the Deno version of date-fns.

fturmel:
@tan75 Thanks, really appreciate the note.

I understand what you're saying. It's an emergent platform (although maybe already stagnating a bit?), and it would be nice for date-fns to be represented, but it's not my call to make.

My counterpoints would be:

- Most (all?) of the work is done already
- The errors caught with this little smoke test I wrote are actual TypeScript type errors, not necessarily Deno-centric so it's helping the overall project's code quality.
- Once the Deno tests run against new PRs, there would be no more catching up to do. Errors would be addressed before they are merged.
- Couldn't the publishing on deno.land be fully automated with github actions, and trigger on new release/tag?

fturmel:
Soft reset and only kept the few type issues left in master.

Deno tests will still fail until we integrate the `uk` locale type fixes in #2911, so this is back in draft mode

PR #2551, opened 2021-07-12
Make `isValid` a type predicate

Krisztiaan:
Simple change to help TS know the typing on the supplied argument in conditionals

### DIFF
--- a/typings.d.ts
+++ b/typings.d.ts
@@ -785,7 +785,7 @@ declare module 'date-fns' {
   function isTuesday(date: Date | number): boolean
   namespace isTuesday {}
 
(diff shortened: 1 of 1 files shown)


fturmel:
Unfortunately it's not that easy. `typings.d.ts` is generated by `/scripts/build/typings.js` and uses the jsdocs in each functions. They need to remain compatible with JS, TS and Flow. Lots of changes required.

Also, if you look at the docs the correct predicate for `isValid` would be `is Date | number`. #2308 is a move in the right direction but will not change the output of `typings.d.ts` until either the build system is updated or the project switches to use `tsc` to generate `.d.ts` files.

PR #2575, opened 2021-08-05
Use the number 1 instead of the word "ein" for formatDuration in German locale

### LINKED ISSUE #2505: Using the number 1 instead of the word "ein" for formatDuration in German locale
Hello. I tried using `formatDuration` to calculate a duration based on an interval. When I try it without a locale (i.e. in English) the result that I get is e.g. "1 year 1 month 15 days". However, when I add the German locale and try to format the same interval, I get "ein Jahr ein Monat 15 Tage". I am not a native German speaker, but it's strange that it does it only to number 1. It doesn't happen to other numbers, e.g. "2". Is there a reason why this is the behavior or is it a bug?

EDIT: My current workaround is to do a replace `formattedString.replace(/ein/g, '1')`

janziemba:
Fixes #2505


janziemba:
Sorry, moved to https://github.com/date-fns/date-fns/pull/2576

PR #2577, opened 2021-08-05
Use the word "sekunda" instead of "vteřina" for second in Czech locale

### LINKED ISSUE #1488: Incorrect Czech translation
The translation for _seconds_ in Czech locale is not correct. It should be “sekunda” (correct unit of time) and its inflections whereas “vteřina” (colloquial, however, an angular unit in reality) is used at the moment.

Was there any strong reasoning behind that @JozefBiros?

https://github.com/date-fns/date-fns/blob/v2.5.1/src/locale/cs/_lib/formatDistance/index.js#L4

janziemba:
Fixes #1488

### DIFF
--- a/src/locale/cs/_lib/formatDistance/index.js
+++ b/src/locale/cs/_lib/formatDistance/index.js
@@ -1,37 +1,37 @@
 var formatDistanceLocale = {
   lessThanXSeconds: {
     one: {
-      regular: 'méně než vteřina',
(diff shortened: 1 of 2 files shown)


janmyler (commented):
As for the Czech language, looking good! 👏

tan75 (approved):
Thank you for this PR! 🙏

PR #2578, opened 2021-08-05
Add the missing number 1 to formatDuration in Czech locale

janziemba:


leshakoss:
Hi @date-fns/i18n-cs, could you please confirm that this is correct and sounds natural in Czech?

fturmel:
@tan75 tests are failing because there is an unresolved merge conflict with the snapshot.md file

leshakoss:
I don't think that number 1 is missing on purpose. `SK` locale doesn't have it either for example: https://github.com/date-fns/date-fns/blob/master/src/locale/sk/_lib/formatDistance/index.ts

PR #2585, opened 2021-08-10
ja i18n: use ヶ not か for month counts

osdiab:
While they're pronounced the same, when written Japanese month counts usually are written with the ヶ character, not the hiragana か. I checked with my native Japanese speaking wife, she says that while か isn't strictly wrong it's not what she would use naturally, and i've never seen it not written with ヶ in my time living in Japan.

https://en.wikipedia.org/wiki/Small_ke

### DIFF
--- a/src/locale/ja/_lib/formatDistance/index.js
+++ b/src/locale/ja/_lib/formatDistance/index.js
@@ -51,13 +51,13 @@ var formatDistanceLocale = {
   },
 
   aboutXMonths: {
-    one: '約1か月',
-    other: '約{{count}}か月'
+    one: '約1ヶ月',
+    other: '約{{count}}ヶ月'
   },
 
   xMonths: {
-    one: '1か月',
-    other: '{{count}}か月'
+    one: '1ヶ月',
+    other: '{{count}}ヶ月'
   },
 
   aboutXYears: {


JhonySpark (approved):

tan75:
hi @date-fns/i18n-ja 
could you please have a look and approve this PR?

tan75 (changes requested):
hi @osdiab 
thank you for your PR! 
Could you please run `yarn locale-snapshots` and commit the results? 
This will help to pass the tests.

ykzts:
Although "1か月" and "1ヶ月" have the same meaning, "ヶ" is an abbreviation of "箇" and is not often used in official situations.

see also https://en.wiktionary.org/wiki/%E3%83%B6

mesqueeb:
I guess we should make it configurable by the developer in some way.

tan75:
hi @date-fns/i18n-ja 
is this PR to go ahead? Or should I close it?

mesqueeb:
@tan75  か is more correct so should be the default. In case the user wants ヶ they can simply do a text replacement.

I feel like making this configurable in date-fns can be considered but will require tremendous effort thinking of implementation syntax. 

You have my vote to close this PR without changes.

PR #2332, opened 2021-03-17
Migrate `getWeekOfMonth` to TypeScript

varwasabi:
I've addressed previous feedback here and have re-opened the PR. @kossnocorp please let me know if there's anything else for this. I also note that after running `./scripts/build/build.sh`, there's quite a lot of changes. I've opened this under the assumption that those files will be committed at a later time once the major version bump takes place.

PR #2146, opened 2021-01-06
migrate getWeekOfMonth function to TS

pabrodez:
Fixes https://github.com/date-fns/date-fns-typescriptify/issues/59

PR #2598, opened 2021-08-17
convert `getWeekOfMonth` to TypeScript

fturmel:
Closes https://github.com/date-fns/date-fns-typescriptify/issues/59

Oddly, there has been 3 abandoned PR for this TS conversion (#2238, #2146, #2332). Hopefully this is the right one!

I tried to simplify the code a bit, and added a test to raise coverage to 100%.


src/getWeekOfMonth/index.ts
```diff
@@ -1,8 +1,9 @@
 import getDate from '../getDate/index'
 import getDay from '../getDay/index'
 import startOfMonth from '../startOfMonth/index'
-import toInteger from '../_lib/toInteger/index'
+import { LocaleOptions, WeekStartOptions } from '../types'
```
tan75, line 4:
```suggestion
import type { LocaleOptions, WeekStartOptions } from '../types'
```

src/getWeekOfMonth/index.ts
```diff
@@ -22,51 +23,38 @@ import requiredArgs from '../_lib/requiredArgs/index'
...
-    locale && locale.options && locale.options.weekStartsOn
-  var defaultWeekStartsOn =
-    localeWeekStartsOn == null ? 0 : toInteger(localeWeekStartsOn)
-  var weekStartsOn =
-    options.weekStartsOn == null
-      ? defaultWeekStartsOn
-      : toInteger(options.weekStartsOn)
+  const defaultWeekStartsOn = options?.locale?.options?.weekStartsOn || 0
+
+  const weekStartsOn =
```
tan75, line 41: It looks like you changed the logic here. Could you please restore the previous logic?

tan75 (changes requested):
Hi! Thank you for this PR! 🙏
Could you please address my comments?

src/getWeekOfMonth/index.ts
fturmel: All tests are passing, the behavior looks the same to me even though it's been simplified. Could you point out specifically where the problem is?

src/getWeekOfMonth/index.ts
fturmel: Done

fturmel:
Rebased + suppress a TS error in the test file that I had missed before

src/getWeekOfMonth/index.ts
leshakoss: Ok I tested this manually and seems like the only difference is `{ weekStatsOn: null }`. Dropping this is fine for v3 but it would be a breaking change for v2. Fixed in the commit: https://github.com/date-fns/date-fns/pull/2598/commits/e081fcf0b6dbaf2b40d348fa45cc26e4bcc6cb04

leshakoss (approved):
Thanks a lot!

src/getWeekOfMonth/index.ts
leshakoss: BTW @tan75 👀

PR #2599, opened 2021-08-17
remove unnecessary `toDate` call in next week day functions (`nextMonday`, etc)

fturmel:
As discussed https://github.com/date-fns/date-fns/pull/2522#issuecomment-890246854


leshakoss (approved):
Thanks a lot!

PR #2602, opened 2021-08-19
`de` and `de-AT` abbreviated stand-alone months support

### LINKED ISSUE #2601: Short month format string ('MMM') doesn't work for german locale anymore as of >= 2.23.0
See https://codesandbox.io/s/focused-silence-oh3ku?file=/src/index.js

It shows "März 2021" twice, both in short and long form. After changing date-fns to 2.22.1, it shows "Mär 2021" for the short form correctly.

Could be related to https://github.com/date-fns/date-fns/commit/ea5c30925daeb2caded8609355b7c29405d2942a#diff-9a0077df644a04c6d103f64e94f2124d02fd48e5b5439abd1042e1080e0fe2ef

Thanks for this very nice library by the way! 👍

fturmel:
Closes #2601

Based on the Unicode CLDR survey, this adds support for abbreviated stand-alone months in the `de` and `de-AT` locales.
https://st.unicode.org/cldr-apps/v#/de/Gregorian/
https://st.unicode.org/cldr-apps/v#/de_AT/Gregorian/

This also fixes the match regex so localized `parse` works properly with the `MMM` and `LLL` tokens.

The `de-AT` locale was converted to TypeScript in the process.


alexbfr (approved):
Appears to work, thanks!

tan75 (approved):
Thank you for this PR! 
Will include in the next release. Subject to approval by @kossnocorp

PR #2603, opened 2021-08-20
Fix eachMinuteOfInterval interval validation

nodeadtree:
Allows eachMinuteOfInterval to use intervals shorter than a minute. Currently throws an `RangeError` if a one second long interval is given.


leshakoss (approved):
Thanks!

Issue #2470, opened 2021-05-08
intervalToDuration error?


diegomvh:
Hi, I'm not sure what is happening, but I have tried different configurations and the function does not return the difference consistently

```javascript
const result = intervalToDuration({
  start: new Date(Date.parse("2021-03-01T00:00:00Z")), 
  end: new Date(Date.parse("2021-09-01T00:00:00Z"))
});
/*
result = { days: 34,
  hours: 0,
  minutes: 0,
  months: 5,
  seconds: 0,
  years: 0 }
*/
```

and

```javascript
const result = intervalToDuration({
  start: new Date(Date.parse("2021-09-01T00:00:00Z")), 
  end: new Date(Date.parse("2021-03-01T00:00:00Z"))
});
/*
result = { days: 0,
  hours: 0,
  minutes: 0,
  months: 6,
  seconds: 0,
  years: 0 }
*/
```

and

```javascript
const result = intervalToDuration({
  start: new Date(Date.parse("2021-02-01T00:00:00Z")), 
  end: new Date(Date.parse("2021-08-01T00:00:00Z"))
});
/*
result = { days: 0,
  hours: 0,
  minutes: 0,
  months: 6,
  seconds: 0,
  years: 0 }
*/
```

and

```javascript
const result = intervalToDuration({
  start: new Date(Date.parse("2021-08-01T00:00:00Z")), 
  end: new Date(Date.parse("2021-02-01T00:00:00Z"))
});
/*
result = { days: 0,
  hours: 0,
  minutes: 0,
  months: 6,
  seconds: 0,
  years: 0 }
*/
```

I do not fully understand what happens with the first result and I have found other similar behaviors where days that go around the month are taken but it does not count as one. `from: "2021-01-01T00:00:00Z" to: "2021-07-01T00:00:00Z"`

Thanks for the wonderful work on the library.

fturmel:
Dates can be so hard! I feel like you got a perfect storm here of feb 28 / leap year, time zones, and defining what the value of a month even means.

These UTC dates get parsed using your system's time zone. If you're in a "negative" time zone relative to UTC you will end up with a Date object that is in the previous day. So your interval is not March 1st to Sept 1st, but actually Feb 28th to Aug 31st.

For example, I'm in eastern time:

```js
console.log( new Date('2021-03-01T00:00:00Z') )
// Sun Feb 28 2021 19:00:00 GMT-0500 (Eastern Standard Time)
```

[...]

Issue #1778, opened 2020-05-17
intervalToDuration and adding Duration to leap-year date give different result


sunil-shrestha:
Hello all,
I am trying to calculate duration between two dates which includes leap-year on the start-date.
On the other-hand, i also need to calculate the end-date based on start-date and duration.
In theory, they are just reverse of each other. But the results are not same. Any idea how we can resolve this?
For example:
```javascript
const startDate = new Date('2020-02-29');
const endDate = new Date('2021-03-29');
const duration = { years: 1, months:1, days: 0 };

add(startDate, duration)
result1 = '2021-03-29' // endDate

intervalToDuration({ start: startDate, end: endDate });
result2 = { years: 1, months:1, days: 1 } 
```

Expected: result2 to be equal to duration

nsafai:
Seeing the same issue between 
```ts
      const timeBefore = new Date("2021-08-31");
      const timeAhead = new Date(timeBefore);
      timeAhead.setFullYear(timeBefore.getFullYear() + 1);
      timeAhead.setMonth(timeBefore.getMonth() + 1);
      
      const { years, months, days, hours, minutes } = intervalToDuration({
         start: currTime,
         end: aheadTime,
      });
      
      console.log(years, months, days, hours, minutes) // Outputs: 1 1 1 0 0, but expected: 1 1 0 0 0
```

PR #2616, opened 2021-09-03
`intervalToDuration` fix for end of month calculations

### LINKED ISSUE #2611: intervalToDuration is off by 1
calculating difference between two dates shows that `intervalToDuration` is off by 1:

```ts
const currTime = new Date();
const timeAhead = new Date(currTime);
timeAhead.setFullYear(currTime.getFullYear() + 1);
timeAhead.setMonth(currTime.getMonth() + 1);
      
const { years, months, days, hours, minutes } = intervalToDuration({
  start: currTime,
  end: aheadTime,
});
      
console.log(years, months, days, hours, minutes) 
// Outputs: 1 1 1 0 0, but expected: 1 1 0 0 0
```

Creating a new issue here since this was first reported in May 2020 but hasn't been resolved yet: https://github.com/date-fns/date-fns/issues/1778

fturmel:
Closes #2611, #1778, #2470

By using `add` internally instead of `sub`, the `intervalToDuration` function gives better results for end of month interval start dates while still passing all existing unit tests. It also means better interop with `add`, as discussed in #2611.

Before:
```js
intervalToDuration({start: new Date(2021, 7, 31), end: new Date(2021, 8, 30)})
// {years: 0, months: 0, days: 30, hours: 0, minutes: 0, seconds: 0}
```

With this PR:
```js
intervalToDuration({start: new Date(2021, 7, 31), end: new Date(2021, 8, 30)})
// {years: 0, months: 1, days: 0, hours: 0, minutes: 0, seconds: 0}
```

This also brings a few improvements:
- new RangeError thrown if interval start date is greater than end date. This is standard in all other functions handling intervals and per https://date-fns.org/docs/Interval
- additional unit tests covering new edge cases and thrown exceptions
- leaner implementation: removed dependencies to `compareAsc` and `isValid`, removed sign and Math.abs operations that were not required.

### DIFF
--- a/src/intervalToDuration/index.ts
+++ b/src/intervalToDuration/index.ts
@@ -1,14 +1,13 @@
-import compareAsc from '../compareAsc/index'
-import differenceInYears from '../differenceInYears/index'
-import differenceInMonths from '../differenceInMonths/index'
+import add from '../add/index'
 import differenceInDays from '../differenceInDays/index'
 import differenceInHours from '../differenceInHours/index'
 import differenceInMinutes from '../differenceInMinutes/index'
+import differenceInMonths from '../differenceInMonths/index'
 import differenceInSeconds from '../differenceInSeconds/index'
-import isValid from '../isValid/index'
-import requiredArgs from '../_lib/requiredArgs/index'
+import differenceInYears from '../differenceInYears/index'
 import toDate from '../toDate/index'
-import sub from '../sub/index'
+import { Duration, Interval } from '../types'
+import requiredArgs from '../_lib/requiredArgs/index'
 
 /**
  * @name intervalToDuration
@@ -24,6 +23,7 @@ import sub from '../sub/index'
  * @throws {TypeError} Requires 2 arguments
  * @throws {RangeError} `start` must not be Invalid Date
  * @throws {RangeError} `end` must not be Invalid Date
+ * @throws {RangeError} The start of an interval cannot be after its end
  *
  * @example
  * // Get the duration between January 15, 1929 and April 4, 1968.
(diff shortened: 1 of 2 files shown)


src/intervalToDuration/index.ts
```diff
@@ -33,49 +33,36 @@ import sub from '../sub/index'
  * })
  * // => { years: 39, months: 2, days: 20, hours: 7, minutes: 5, seconds: 0 }
  */
-
-export default function intervalToDuration({ start, end }: Interval): Duration {
+export default function intervalToDuration(interval: Interval): Duration {
```
nsafai, line 36: what's the benefit of changing the func signature here? can we avoid a breaking change to syntax?

src/intervalToDuration/index.ts
```diff
@@ -33,49 +33,36 @@ import sub from '../sub/index'
...
   requiredArgs(1, arguments)
 
-  const dateLeft = toDate(start)
-  const dateRight = toDate(end)
+  const start = toDate(interval.start)
+  const end = toDate(interval.end)
```
nsafai, line 40: if you really want to keep the internal variable names the same (while keeping the func signature the same as before), you could do something like this:
```suggestion
  start = toDate(start)
  end = toDate(end)
```

src/intervalToDuration/index.ts
```diff
@@ -33,49 +33,36 @@ import sub from '../sub/index'
...
 
-  const duration = {
-    years: 0,
-    months: 0,
-    days: 0,
-    hours: 0,
-    minutes: 0,
-    seconds: 0,
+  const duration: Duration = {
+    years: differenceInYears(end, start),
```
nsafai, line 49: nice cleanup here 🥇

src/intervalToDuration/index.ts
```diff
@@ -33,49 +33,36 @@ import sub from '../sub/index'
...
   }
 
-  const sign = compareAsc(dateLeft, dateRight)
-
-  duration.years = Math.abs(differenceInYears(dateLeft, dateRight))
-
-  const remainingMonths = sub(dateLeft, { years: sign * duration.years })
-  duration.months = Math.abs(differenceInMonths(remainingMonths, dateRight))
+  const remainingMonths = add(start, { years: duration.years })
+  duration.months = differenceInMonths(end, remainingMonths)
```
nsafai, line 53: this is a nice cleanup as well 👏

src/intervalToDuration/test.ts
```diff
@@ -1,15 +1,16 @@
...
-describe('intervalToDuration', function () {
-  it('returns correct duration for arbitrary dates', function () {
+describe('intervalToDuration', () => {
+  it('returns correct duration for arbitrary dates', () => {
     const start = new Date(1929, 0, 15, 12, 0, 0)
     const end = new Date(1968, 3, 4, 19, 5, 0)
     const result = intervalToDuration({ start, end })
 
-    assert.deepEqual(result, {
+    assert.deepStrictEqual(result, {
```
nsafai, line 13: even safer, nice

src/intervalToDuration/test.ts
```diff
@@ -49,9 +50,36 @@ describe('intervalToDuration', function () {
...
+  })
+
+  it("throws a RangeError if interval's end date invalid", () => {
+    const interval = {
+      start: new Date(2020, 0, 1),
+      end: new Date(NaN),
+    }
+
+    assert.throws(intervalToDuration.bind(null, interval), RangeError)
+  })
```
nsafai, line 78: nice additions

src/intervalToDuration/test.ts
```diff
@@ -112,5 +140,147 @@ describe('intervalToDuration', function () {
...
+          months: 1,
+          days: 1,
+          hours: 0,
+          minutes: 0,
+          seconds: 0,
+        }
+
+        assert.deepStrictEqual(duration, expectedDuration)
+      })
+    })
```
nsafai, line 284: nice these are all great edge cases to test

nsafai (commented):
i think it's worth leaving the func signature as-is, other than that this looks great, tysm for following up on this!

nsafai:
lmk what you think of my suggestion of keeping func signature as-is and feel free to tag me for a +1

src/intervalToDuration/index.ts
fturmel: Correct me if I'm wrong, but there are no signature changes here. It's not destructuring the object anymore but the only argument is still an Interval object. As I think you noticed, the change was to allow the use of `start` and `end` variable names in the function body for readability.

src/intervalToDuration/index.ts
fturmel: Personal preference, but I don't love assignments to function arguments. I tend to treat them as read only. In this case, they are also not quite the same type. Interval start and end are `Date | number` but the return value of `toDate` is `Date`.

fturmel:
@nsafai Thanks for taking the time and all the nice notes! As I replied in my comments I think the method signature is intact but let me know if I'm missing something.

nsafai (approved):
ah you're entirely right, func signature is still backwards compatible. LGTM!

tan75 (approved):
Thank you!

hiendaovinh:
Thanks for the fix but it's a breaking change because of the Math.abs removal. Please bump a major version next time.

leshakoss:
@hiendaovinh sorry, that was not supposed to happen :[ A patch with a revert coming soon

leshakoss:
@hiendaovinh actually, I don't think this is a breaking change. If interval.start was after interval.end the function would've thrown a RangeError anyway, so there's no way to get negative values from the function. If I'm wrong, please provide a test case that we can use to verify this 🙇

fturmel:
>  If interval.start was after interval.end the function would've thrown a RangeError anyway

@leshakoss The RangeError was introduced in this PR, so if there are any breaking changes that would be it. However that error should've always been there if you look at the behavior of every other functions accepting an Interval in the library, per my PR notes.

leshakoss:
Ah, I see. Honestly, this is a tough call. I would count this as a bug fix rather than a breaking change, but that's maybe because I already have "interval.start is not supposed to be after interval.end" in mind, but I understand why one might think the opposite

hiendaovinh:
I do understand and I share the same thinking without regarding the "interval.start is not supposed to be after interval.end" stuff. Actually one of my colleges had a bug switching their places so I knew :laughing:. It's certainly a bug fix but also a breaking change in its own way. That' interesting.

leshakoss:
@hiendaovinh reverted the change in v2.29.2: https://github.com/date-fns/date-fns/releases/tag/v2.29.2

PR #2617, opened 2021-09-05
improved `clamp` implementation

fturmel:
This is an improved implementation of the `clamp` function.

- approx. 45% faster perf (1.8MM vs 2.6MM ops/sec)
- slightly smaller overall file size by removing min/max imports
- fix missing import of Interval type
- RangeErrors for date and interval arguments
- added `eslint-env mocha` to test file to remove eslint errors

### DIFF
--- a/src/clamp/index.ts
+++ b/src/clamp/index.ts
@@ -1,5 +1,5 @@
 import toDate from '../toDate/index'
-import { Interval } from '../types'
+import type { Interval } from '../types'
 import requiredArgs from '../_lib/requiredArgs/index'
 
 /**
--- a/src/clamp/index.ts
+++ b/src/clamp/index.ts
@@ -1,6 +1,5 @@
-import max from '../max/index'
-import min from '../min/index'
-import type { Interval } from '../types'
+import toDate from '../toDate/index'
+import { Interval } from '../types'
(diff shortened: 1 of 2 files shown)


src/clamp/index.ts
```diff
@@ -27,11 +27,29 @@ import requiredArgs from '../_lib/requiredArgs/index'
...
+
+  const date = toDate(dirtyDate)
+  const start = toDate(interval.start)
+  const end = toDate(interval.end)
+
+  if (isNaN(date.getTime())) throw new RangeError('Date is invalid')
+  if (isNaN(start.getTime())) throw new RangeError('Start Date is invalid')
+  if (isNaN(end.getTime())) throw new RangeError('End Date is invalid')
+  if (start > end) {
+    throw new RangeError('The start of an interval cannot be after its end')
```
leshakoss, line 49: This is a good change, but it is a breaking change too. We'll have to wait with this until v3

fturmel: Fair enough. If you like the idea for v3, maybe there should be a utility method to validate Interval arguments (similar to `requiredArgs`). There are at least 13 functions in `date-fns` that take intervals as input.

Issue #2484, opened 2021-05-20
eachDayOfInterval incorrect start date


VasilyVP:
Hi, my next code:
`
const end = new Date();
const start = addDays(new Date(), -7);

console.log('start: ', start);
console.log('end: ', end);
console.log(eachDayOfInterval({ start, end }));
`
returns
![image](https://user-images.githubusercontent.com/31712700/118985454-222c8a80-b987-11eb-8900-a50f4e20c1ee.png)

so, the first day is one day before start

rikkalo:
Hi
Thanks for the issues!
Which version of the date-fns you are using? Which version of the node or browser where you run the example?

VasilyVP:
Hi,
I checked it on
"date-fns": "^2.21.3",
and nodejs v14.17.0

Vasily.

вс, 23 мая 2021 г. в 22:47, Anastasia Kobzar ***@***.***>:

rikkalo:
Can you tell me more about the timezone?

VasilyVP:
new Date() returns Date object which represents time in UTC,
so it's UTC timezone.

пн, 24 мая 2021 г. в 15:43, Anastasia Kobzar ***@***.***>:

brugos:
Same here! this function is localising the output!
`start 2021-05-01T00:00:00.000Z
end 2021-05-05T00:00:00.000Z
queryDates [
  2021-04-30T23:00:00.000Z,
  2021-05-01T23:00:00.000Z,
  2021-05-02T23:00:00.000Z,
  2021-05-03T23:00:00.000Z,
  2021-05-04T23:00:00.000Z
]`
I guess it's removing one hour because I'm on BST (GMT+1) at the moment.

fturmel:
`date-fns` is doing its operations with native date objects in the system's time zone. https://date-fns.org/docs/Time-Zones

So while confusing I think it's the expected behavior for native JS Date objects. When you create a Date instance, it creates it in your system's time zone.

For example, I'm in eastern time:

```js
const d = new Date(2021, 0, 1)

d.toString() // Fri Jan 01 2021 00:00:00 GMT-0500 (Eastern Standard Time)
d.getMonth() // 0
d.getDate() // 1

// but as UTC it's 5 hours later
d.toISOString() // 2021-01-01T05:00:00.000Z
```

If you give the constructor a UTC string, it will convert that UTC date into your local time zone, and as you can see it's not the same day anymore.

```js
[...]

Issue #2151, opened 2021-01-12
formatISO doesn't use UTC (Z) timezone


scscgit:
When making a sanity check of `now.toISOString() === formatISO(now)` (with `const now = new Date()`), it fails due to a timezone difference, as the **formatISO** doesn't match the browser's [**toISOString**](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString), which explicitly claims that `The timezone is always zero UTC offset, as denoted by the suffix "Z".`. The [formatISO docs \@v2.16.1](https://date-fns.org/v2.16.1/docs/formatISO) doesn't mention anything about this - it actually doesn't even seem to support configuring UTC to be used - and it has an example, which is thus incorrect (non-deterministic):
```
// Represent 18 September 2019 in ISO 8601 format (UTC):
const result = formatISO(new Date(2019, 8, 18, 19, 0, 52))
//=> '2019-09-18T19:00:52Z'
```
Actual result:
<table>
<tr><td>formatISO(new Date())</td><td>2021-01-12T02:57:09+01:00</td></tr>
<tr><td>formatISO(new Date(2019, 8, 18, 19, 0, 52))</td><td>2019-09-18T19:00:52+02:00</td></tr>
<tr><td>new Date(2019, 8, 18, 19, 0, 52).toISOString()</td><td>2019-09-18T17:00:52.000Z</td></tr>
</table>

Tested on `v2.16.1` in:
* Chrome 87.0.4280.88
* Firefox 84.0.2

devuxer:
The documentation is confusing on this. The examples for `formatISO` imply that it yields a date string in UTC time, but on my Windows machine at least, it always yields a data string in local time (i.e., the operating system's time zone setting).

Conversely, the built-in `toISOString` appears to always yield a date string in UTC regardless of the OS's time zone setting.

Taking a quick look at the [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators) standard, either of these approaches appear to be acceptable:

> Time zones in ISO 8601 are represented as local time (with the location unspecified), as UTC, or as an offset from UTC.

[...]

Issue #1604, opened 2020-01-17
Should startOfISOWeek be Local Timezone?


sbwrege2z:
"The result will be in the local timezone."

This has caused me some problems.  I am currently Eastern Standard Time (GMT-0500).

So new Date('2020-01-13T00:00:00.000Z') is "Sun Jan 13 2020 00:00:00 GMT-0000" or "Sun Jan 12 2020 19:00:00 GMT-0500".  This is not ambiguous.

The thing that is causing me problems is that the startOfISOWeek has 2 different answers (a week apart) depending on the local time zone you are in.  In the example above, the start of the ISO week will be "Sun Jan 13 2020 00:00:00 GMT-0000" if you are at or east of UTC, and it will be "Sun Jan 6 2020 00:00:00 GMT-0000" if you are west of of UTC.

Is this the intent?  I though the point of using ISO date/times was to eliminate ambiguity.

Issue #376, opened 2017-01-06
UTC functions


leshakoss:
Make UTC versions of functions and make non-UTC versions (regular ones) be based on UTC-versons.

Example:

```javascript
// date-fns/utc/addDays

import toDate from '../toDate'

export default function addDays (dirtyDate, amount, options) {
  const date = toDate(dirtyDate, options)
  date.setUTCDate(date.getDate() + amount)
  return date
}
```

```javascript
// date-fns/utc/subDays

import addDays from '../addDays'

export default function subDays (dirtyDate, amount, options) {
  return addDays(dirtyDate, -amount, options)
}
```

```javascript
// date-fns/subDays

import utcSubDays from '../utc/subDays'
import toUTC from '../_lib/toUTC'
import fromUTC from '../_lib/fromUTC'

export default function subDays (dirtyDate, amount, options) {
  const utcDate = toUTC(dirtyDate)
  const utcResult = utcSubDays(utcDate, amount, options)
  return fromUTC(utcResult)
}
```

The problem with non-UTC functions is that they are not pure and can produce the different results in the different timezones. Moreover, in some timezones they can produce the wrong result. That happens when the date just doesn't exist in a timezone:

```
$ env TZ="America/New_York" node
> new Date(2017, 2 /* March */, 12, 2).toString()
'Sun Mar 12 2017 03:00:00 GMT-0400 (EDT)'
> new Date(2017, 2 /* March */, 12, 3).toString()
'Sun Mar 12 2017 03:00:00 GMT-0400 (EDT)'
```

In `America/New_York` timezone the date `2017-03-12T02:00:00` just doesn't exist because DST.
So when in chain of functions some intermediate result should be `2017-03-12T02:00:00`, the result of the chain will be wrong in `America/New_York` timezone. Sadly, it can also happen without chaining, because some of our functions can depend on the other functions. So every non-UTC function is not pure and produces side effects and subtle bugs. The way to resolve this issue is to make non-UTC operations minimal (is the best case, it should be done only twice: to convert local date to UTC and to convert the result back to the local timezone)

jbrantly:
We just ran into a situation where we needed to `format` dates in UTC but were unable to using date-fns since it only formats in local time. I'm assuming that feature would be bundled into this issue? Very much looking forward to this!

hlehmann:
Same issue with startOfDay in UTC

hlehmann:
Is this coming with v2 ?

leshakoss:
[...]

PR #2619, opened 2021-09-07
Fix timezone support in isSameMonth

marcbachmann:
Related Issues:
- https://github.com/date-fns/date-fns/issues/376
- https://github.com/date-fns/date-fns/issues/2484
- https://github.com/date-fns/date-fns/issues/2151
- https://github.com/date-fns/date-fns/issues/1604

The current implementation of `isSameMonth` didn't respect the timezones.

```js
const january = new Date('2015-02-01T00:00:00.000+01:00')
const february = new Date('2015-02-01T00:00:00.000Z')

// Current behavior 
_.isSameMonth(january, february)
true

// Expected behavior
_.isSameMonth(january, february)
false
```

### DIFF
--- a/src/isSameMonth/index.ts
+++ b/src/isSameMonth/index.ts
@@ -29,7 +29,7 @@ export default function isSameMonth(dirtyDateLeft: Date | number, dirtyDateRight
   const dateLeft = toDate(dirtyDateLeft)
   const dateRight = toDate(dirtyDateRight)
   return (
-    dateLeft.getFullYear() === dateRight.getFullYear() &&
-    dateLeft.getMonth() === dateRight.getMonth()
+    dateLeft.getUTCFullYear() === dateRight.getUTCFullYear() &&
+    dateLeft.getUTCMonth() === dateRight.getUTCMonth()
   )
 }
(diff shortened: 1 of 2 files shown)


leshakoss:
This doesn't look correct.

`Date` object doesn't store the timezone it has been created with. In fact `Date` is basically a wrapper over a single value, the date's timestamp. `isSameMonth(january, february)` does return `true` for me, but that is because in my timezone both dates happen to refer to the same day.

```
const january = new Date('2015-02-01T00:00:00.000+01:00')
const february = new Date('2015-02-01T00:00:00.000Z')
isSameMonth(january, february) //=> true
january.toString() //=> "Sun Feb 01 2015 00:00:00 GMT+0100 (CET)"
february.toString() //=> "Sun Feb 01 2015 01:00:00 GMT+0100 (CET)"
```

marcbachmann:
Yes, this should be covered by implementing utc support for `isSameMonth` similar to https://github.com/date-fns/date-fns/blob/master/src/_lib/isSameUTCWeek/index.ts

The docs mention something about that, but it's not really working.
`date-fns/utc` doesn't exist in the module.

image

What are the plans for that and [#1458 ](https://github.com/date-fns/date-fns/issues/376)?

PR #2623, opened 2021-09-09
fix: returned type of parseISO function should returns a Date or the string Invalid Date.

### LINKED ISSUE #2622: Wrong return type for parseISO
Hi, 

I use the parseISO function like this : 

```
      const date = parseISO(val)
      if (date === "Invalid Date") return false
```

TS compiler returns an error because parseISO is supposed to return a Date. 
But in fact, it can return a string with the value "Invalid Date".

In my idea, it would be better to return an exception.
But at least, the definition should be : `parseInfo(...): Date | string`.

pom421:
Fix #2622

### DIFF
--- a/typings.d.ts
+++ b/typings.d.ts
@@ -10050,7 +10050,7 @@ declare module 'date-fns/esm' {
     options?: {
       additionalDigits?: 0 | 1 | 2
     }
-  ): Date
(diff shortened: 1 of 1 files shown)


pom421:
My bad, parseISO returns a Date. See the issue.

PR #2630, opened 2021-09-13
Remove bad jsdocs and fp functions for `endOfDecade` and `lastDayOfQuarter`

fturmel:
The jsdocs did not reflect the correct function signature, which generated wrong documentation and `...WithOptions` fp functions that should not exist.


fturmel:
Fixed already with changes to jsdocs/documentation in v3

PR #2658, opened 2021-10-02
Fix/change abbreviation locale id follow momentjs

mandaputtra:
I notice when I migrate from momentjs, the abbreviation for Agustus aren't the same.

```
Ags => Agt
```


tan75:
hi @date-fns/i18n-id 
could you please have a look into this change and approve it?

PR #2672, opened 2021-10-07
Convert format duration to typescript

Liam-Tait:
Replace regex replace with lookup on object. This adds type safety that regex does not give
Added `count` variable for type safety, this allows type narrowing
Added trailing commas by running prettier

Closes https://github.com/date-fns/date-fns-typescriptify/issues/36


Liam-Tait:
Hey @tan75, do I need to do anything to this to get it ready for review/merged?

tan75:
Hi @Liam-Tait 
Thanks a lot for your help however, unfortunately, we can not accept this PR as there are some changes to the logic in it. At the moment our focus is to translate the functions to TS without any logic changes so we could release v3. 
I will convert this issue to discussion so we could discuss the logic changes you proposed above and work on it once the v3 is out.
Thanks again 🙏

PR #2683, opened 2021-10-10
Update getWeek documentation to reference the correct function

### LINKED ISSUE #2100: getWeek documentation is misleading
Landing from a google search, https://date-fns.org/v2.16.1/docs/getWeek documentation is not accurate, the code examples are for getISOWeek which doesn't accept a second parameter.

image

Zacharias3690:
Closes #2100


leshakoss (approved):
Thanks!

PR #2709, opened 2021-10-20
Monday as first day of the week for korean

dogfootruler-kr:
We call Saturday and Sunday as 주말(週末). 주(週) means "week" and 말(末) means "end".
So you can see that Sunday is the last day of a week in Korean language.

### DIFF
--- a/src/locale/ko/index.js
+++ b/src/locale/ko/index.js
@@ -22,7 +22,7 @@ var locale = {
   localize: localize,
   match: match,
   options: {
-    weekStartsOn: 0 /* Sunday */,
+    weekStartsOn: 1 /* Monday */,
     firstWeekContainsDate: 1
   }
 }


leshakoss:
Hi @date-fns/i18n-ko could you please confirm that this is correct? `weekStartsOn` affects calculation of week-related functions like `isSameWeek` for example.

iamssen:
https://korean.go.kr/front/onlineQna/onlineQnaView.do?mn_id=216&qna_seq=218209&pageIndex=1

The official answer of the National Institute of Korean Language is that the start of the week is Monday.

However, I think changing that `weekStartsOn` will change all the week-related calculations. I think this is likely to create too many breaking changes. 

image

Koreans think that the start of the week is 54% on Mondays and 43% on Sundays according to the survey, so apps such as calendars generally start Sunday and then add options to change to Monday. I think there may be a risk when the result values of functions such as `isSameWeek()` become different from before.

I think I'll have to look more at how much this change will affect.

iamssen (approved):

tribela:
NIKL is not an organisation that standardize specs on this kind.
Should look at KS(Korean Standard) spec (KS X ISO8601).
https://standard.go.kr/KSCI/standardIntro/getStandardSearchView.do?menuId=503&topMenuId=502&ksNo=KSXISO8601&tmprKsNo=KSXISO8601&reformNo=03
Which also specifying that Monday is the first day of the week.
> 2.2.8
> 역주 (Calendar week)
> 역년 중의 서수에 의하여 지정되는 특정한 7일의 기간으로 월요일부터 시작된다.

lifthrasiir:
> NIKL is not an organisation that standardize specs on this kind. Should look at KS(Korean Standard) spec (KS X ISO8601).

KS X ISO 8601 is, uh, just a localized version of ISO 8601. The ISO 8601 week calendar is distinct to other widely used week calendars; technically speaking it's a [leap week calendar](https://en.wikipedia.org/wiki/Leap_week_calendar) with a different year grouping compared to other calendars which limits its uses (it even triggered a [Twitter bug](https://www.theguardian.com/technology/2014/dec/29/twitter-2015-date-bug) back in 2014). As such it is irrelevant for this discussion.

KS X ISO 8601은... 그냥 ISO 8601의 지역화된 버전입니다. ISO 8601 주간 달력은 다른 널리 쓰이는 주간 달력과는 별개의 체계인데, 기술적으로 말하자면 [윤주 달력](https://en.wikipedia.org/wiki/Leap_week_calendar)이라 해서 한 해의 시작과 끝이 다른 달력과 다르게 잡혀 있기 때문에 쓸모가 제한됩니다(심지어 2014년에는 이것 때문에 [트위터 버그](https://www.theguardian.com/technology/2014/dec/29/twitter-2015-date-bug)가 나기도 했고요). 그러니 이 표준은 현 논의에서는 무관합니다.

dittos:
[Unicode CLDR](https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/territory_information.html) has "First day of week" data. For South Korea, it is Sunday.

I know nothing about date-fns project's policy, but relying on the external (authorative?) source would be more productive than discussing about each locale's case in this project.

Maybe related: https://github.com/tc39/ecma402/issues/6

leshakoss:
In that case, we'll keep it the same as in CLDR (i.e. Sunday). Thanks!

PR #2712, opened 2021-10-20
Clarify start vs. end parameters in differenceIn* methods

### LINKED ISSUE #2179: difference (differenceInMinutes etc.) parameter naming
Hello, I find the naming of arguments (dateLeft, dateRight) in difference functions very confusing. 

I always imagine them on a timeline, where dateLeft would be on the left, ie. earlier, and dateRight on the right, ie. later. But for some reason it's the other way around.
I'm not necessarily asking for a change, but could you please provide a rationale for choosing those argument names?

Thank you very much
Jiri Richter
A very thankful admirer of date-fns.

gibson042:
Fixes #2179

I'm open to alternative approaches if preferred, but this should at least get the ball rolling.

### DIFF
--- a/src/differenceInBusinessDays/index.ts
+++ b/src/differenceInBusinessDays/index.ts
@@ -10,17 +10,17 @@ import requiredArgs from '../_lib/requiredArgs/index'
 /**
  * @name differenceInBusinessDays
  * @category Day Helpers
- * @summary Get the number of business days between the given dates.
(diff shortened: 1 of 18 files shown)


tan75:
I see the point of this PR however it is hard to accept that many changes in one request.
@leshakoss what do you think? One of my concerns is variable names.

PR #2730, opened 2021-10-28
docs: fix grammar in i18nContributionGuide.md

wingclover:
The most obvious one was probably a typo in `Could take of these strings as the argument`. It was not clear whether it meant `one of` or `some of`.

All the other things are minor.


leshakoss (approved):
Thanks a lot!

PR #2753, opened 2021-11-07
formatRelative singular on locale/es package fix #2710

### LINKED ISSUE #2710: Wrong `formatRelative` for the  `locale/es` package.
Hi, I have noticed an error in the code for the `locale/es` package.

https://github.com/date-fns/date-fns/blob/f8fb243c55853f90d04efd8933fc291a24402b27/src/locale/es/_lib/formatRelative/index.js#L1-L8

https://github.com/date-fns/date-fns/blob/f8fb243c55853f90d04efd8933fc291a24402b27/src/locale/es/_lib/formatRelative/index.js#L10-L17

The `last week` key on the singular object shows the month (`L`) and the milliseconds timestamp (`T`), when it should be displaying the Long localized time (`p`) as its plural counterpart.

I cannot guarantee it is the only error, as I have only tested this particular function.

jhonatanmacazana:


tan75 (approved):
Thank you!

PR #2784, opened 2021-11-24
Updated Swedish short date format

Yrlish:
Changed Swedish short date format from `y-MM-dd` to `yyyy-MM-dd`, which is the correct Swedish official date format.
Sweden basically follows ISO-8601 as official date format.

### DIFF
--- a/src/locale/sv/_lib/formatLong/index.js
+++ b/src/locale/sv/_lib/formatLong/index.js
@@ -4,7 +4,7 @@ var dateFormats = {
   full: 'EEEE d MMMM y',
   long: 'd MMMM y',
   medium: 'd MMM y',
-  short: 'y-MM-dd'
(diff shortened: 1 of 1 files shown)


leshakoss:
Hi @date-fns/i18n-sv could you please confirm that this is correct?

alexandernanberg:
CLDR says `y-MM-dd` is correct https://www.unicode.org/cldr/cldr-aux/charts/32/summary/sv.html#1939, it also makes little difference in a real world case. E.g. `2021-11-30` is formated the same way, it only makes a difference for dates like `0001-12-24` etc. So I would be in favour of keep using the CLDR suggested format

leshakoss:
@alexandernanberg thanks!
@Yrlish thanks a lot for your contribution! We try to keep our formats as close to CLDR as possible, so I am closing this.

PR #2796, opened 2021-11-29
Fixed `do` formatter for `fr`, `fr-CA` and `fr-CH` locales (closes #1391)

### LINKED ISSUE #1391: Wrong translation for French.
Hi!

When we use the format `iiii do MMMM yyyy`, we should have this translation:

"Jeudi 1er août 2019" and "Jeudi 29 août 2019"
but we have:
"Jeudi 1er août 2019" and "Jeudi 29[**ème**](#) août 2019"
                                                           
There is an extra "ième" when the day is greater than 1, this should not happen.

[src/locale/fr/_lib/formatters/index.js](https://github.com/date-fns/date-fns/blob/master/src/locale/fr/_lib/formatters/index.js) is not included / loaded and use the wrong format (`Do` instead of `do`). I didn't know how to do it. `formatters` is not used when I place it in the `locale`... and the engine seems to ignore it...

It does not break the functionalities but it's a bad "experience" for users... I can replace it if French, but... 😄 

Regards

**EDIT**:
New link in v3: https://github.com/date-fns/date-fns/blob/main/src/_lib/format/formatters/index.ts#L317

leshakoss:


leshakoss:
Hi @date-fns/i18n-fr. Tried to make ordinal day of the month formatter more useful and fix #1391. I feel like this should cover more `format` use cases than not having this fix at all. Am I correct? This is how moment.js behaves too.

fturmel:
@leshakoss I'm on the fence personally.

As pointed out [here](https://github.com/date-fns/date-fns/issues/1391#issuecomment-530344774) there are some formatting scenarios where the current behavior is fine, and this PR could be a breaking change for many.

`le 2ème jour du mois` vs `le 2 novembre`

In v1, this conditional logic was only applied if `do` was immediately followed by `MMMM`.

https://github.com/date-fns/date-fns/blob/67a0329d663f82393261f38c4c47ad65c038bd35/src/locale/fr/build_format_locale/index.js#L80-L98

PR #2801, opened 2021-11-30
add: diff year example on isSameMonth

### LINKED ISSUE #2760: docs: add example for isSameMonth with different year
https://date-fns.org/v2.25.0/docs/isSameMonth#examples

currently the examples say:

```js
// Are 2 September 2014 and 25 September 2014 in the same month?
var result = isSameMonth(new Date(2014, 8, 2), new Date(2014, 8, 25))
//=> true
```

but would be better if it was:
```js
// Are 2 September 2014 and 25 September 2014 in the same month?
var result = isSameMonth(new Date(2014, 8, 2), new Date(2014, 8, 25))
//=> true

// same month different year
var result = isSameMonth(new Date(2015, 8, 2), new Date(2014, 8, 25))
//=> false
```

BTW: Will this be `false` if it's a different year?

hg-pyun:
fix #2760

Add diff year example on isSameMonth.


leshakoss (approved):
Thanks!

PR #2556, opened 2021-07-22
added uz-cryl locale

kamronbek28:
#2550

PR #2811, opened 2021-12-02
added uz-Cyrl locale

tan75:
A copy of this PR https://github.com/date-fns/date-fns/pull/2556


fturmel:
@tan75 possible typo for the locale? I think it might be `uz_Cyrl` https://unicode-org.github.io/cldr-staging/charts/40/summary/uz_Cyrl.html

tan75:
> @tan75 possible typo for the locale? I think it might be `uz_Cyrl` https://unicode-org.github.io/cldr-staging/charts/40/summary/uz_Cyrl.html

Yes, you are right! Thanks for the heads-up!

leshakoss (approved):
👍

PR #2814, opened 2021-12-03
`parse` fix: allow `K` token to be used with `a` or `b`

### LINKED ISSUE #1789: Parse string result in incompatible token K and a
When parse a string using following format:

```
dd/MM/yyyy KK:mm a
```

It will throw the following error:

```
The format string mustn't contain `KK` and `a` at the same time
```

However, using `hh` does not throw the error.

The only different between `K` and `h` is `K` is `[0-11]` and `h` is `[1-12]`.

Am I misunderding the `K` pattern?

---

Patterns Table:

image

From: https://date-fns.org/v2.14.0/docs/format

fturmel:
Closes #1789

Since `K` parses hours with patterns 0-11 and 00-11, its usage with the `a` and `b` tokens (day periods) should be allowed.


leshakoss (approved):
Thanks, that's awesome!

PR #2835, opened 2021-12-09
Feat(docs)/update examples

mesqueeb:
As a continuation of the original issue at https://github.com/date-fns/date-fns/issues/2760 I have updated all the similar functions with more clear descriptions to prevent any confusion around the topic.


leshakoss (approved):
Thanks!

PR #2844, opened 2021-12-13
minor TypeScript type fixes

fturmel:
This fixes some missing type imports and adds explicit return types where they were missing.

It also fixes two additional issues I stumbled upon:
- `setWeek` options wasn't optional in the function definition (missing a `?`)
- `eachWeekOfInterval` was missing `LocaleOptions` as a type definition to its options argument. This options argument gets passed down to `startOfWeek` which accepts `LocaleOptions`.


leshakoss (approved):
Thanks a lot!

PR #2669, opened 2021-10-06
Migrate hi locale

minitesh:
This PR if for issue date-fns/date-fns-typescriptify#176

Fixes https://github.com/date-fns/date-fns-typescriptify/issues/176

PR #2845, opened 2021-12-13
fix `hi` locale type issues

fturmel:
See my comment on the merged PR for `hi`: [#2669](https://github.com/date-fns/date-fns/pull/2669#issuecomment-992592632)

~The ordinalNumber function in localize still has issues, so this is in draft mode.~

Out of draft mode, but would still like a review / confirmation that the `ordinalNumber` function is OK @minitesh

### DIFF
--- a/src/locale/hi/_lib/formatDistance/index.ts
+++ b/src/locale/hi/_lib/formatDistance/index.ts
@@ -1,6 +1,5 @@
-import type { FormatDistanceLocale, FormatDistanceToken } from '../../../types'
-import { FormatDistanceFnOptions } from '../../../types'
-import localize from '../localize/index'
+import type { FormatDistanceFn, FormatDistanceLocale } from '../../../types'
+import { numberToLocale } from '../localize/index'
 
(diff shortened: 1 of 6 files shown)


fturmel:
@minitesh There are type issues remaining with the ordinalNumber function. Can you help me fix these?

- localeToNumber function takes in a string, not a number
- The function sometimes returns undefined, so that needs to be addressed. What should be returned when `rem10` is not between 0 and 9?
 
Screen Shot 2021-12-13 at 12 41 26 PM

fturmel:
> The function sometimes returns undefined, so that needs to be addressed. What should be returned when rem10 is not between 0 and 9?

Err, answering my own question here. Obviously any positive integer modulo 10 will return a value between 0 and 9... So am I wrong in thinking this whole function should only be the following?

```ts
const ordinalNumber: LocalizeFn<number, undefined> = (
  dirtyNumber,
  _options
) => {
  const number = Number(dirtyNumber)
  return numberToLocale(number)
}
```

The snapshot.md file remains the same.

tan75 (approved):
Thanks a lot for this!

Issue #2666, opened 2021-10-06
Add isISOMatch function


Liam-Tait:
Suggest adding `isISOMatch` function, used for validating a date string is in ISO 8601 format.

Options similar to the output of `formatISO` would be used to further restrict the validateion

The usages of this function is similar to the `isMatch` function, except where stringent ISO formatting should be used.
In my personal use case I would use the 
```ts
isISOMatch(dateString, { representation: 'date' })
```
to validate ISO date strings
as parameters to an API.

This is in line with other ISO specific functions, such as:
- `parse` -> `parseISO`
- `format` -> `formatISO`
- `formatDuration` -> `formatISODuration`

In line with `formatDuration` -> `formatISODuration`, it is more natural to go with 
`isISOMatch` than `isMatchISO` another option could be `isISO` though I believe this would cause more confusion and is not in line with the existing api.

### Usage

```ts
// CommonJS
var isISOMatch = require('date-fns/isISOMatch')
// ES 2015
import isISOMatch from 'date-fns/isISOMatch'
// ESM
import { isISOMatch } from 'date-fns'
```

### Syntax

```ts
isISOMatch(dateString, [options]);
```

### Arguments

| Name       | Type   | Description               |
| ---------- | ------ | ------------------------- |
| dateString | String | The date string to verify |
| options    | Object | An object with options    |

#### Options

| Name                                                   | Type                           | Description                               |
| ------------------------------------------------------ | ------------------------------ | ----------------------------------------- |
| representation <br /> _(optional, default='complete')_ | 'complete' \| 'date' \| 'time' | match date, time with time zone, or both |

### Returns

| Type    | Description                           |
| ------- | ------------------------------------- |
| Boolean | is the date string an ISO date string |

lielfr:
Is this issue available to work on?

Thanks

Liam-Tait:
> I think this function is a good idea so please go ahead with the implementation.
Excellent, thank you for reviewing

> Re the naming situation - this is an inconsistency that we, unfortunately, didn't spot - apologies for the confusion!

@tan75  I have compiled a list of ISO functions and if they take the suffix approach, based on this it seems the non-suffix approach is more consistent. I would argue that the those functions that use the suffix approach are using the same sentence naming and they just happen to end with a suffix

| ISO function                     | suffix |
| -------------------------------- | ------ |
| format ISO                       | ✅     |
| formatISO9075                    | ✅     |
| formatISODuration                |        |
[...]

PR #2865, opened 2021-12-21
Add isISOMatch function (date-fns#2666)

klokar:
Ref #2666

Added `isISOMatch` function for validating if a date string is in ISO 8601 format.
Implementation is done according to discussion in issue thread #2666 and contains tests for all `'complete' | 'date' | 'time'` representation types.

@tan75 @Liam-Tait


klokar:
Should something additionally be done here? @tan75 @Liam-Tait

Cellule:
Just to mention that I really like that you went with the regex approach. 
I've just hit a bug in production where I mistakenly thought that `parseISO` would only parse ISO strings and was taken aback when `parseISO("50") === "5000-01-01T00:00:00.000Z` which is not at all what I was going for.
I wanted to detect is a string was an ISO date and it was possible for numbers to be in there

xanderknol:
Can this PR maybe get some eyes on or at least the functionality?

I am in need of a function exactly like this only to find this PR not implemented after quite a long time. Not sure if a decision was made to not implement this though, if that is the case just let me know.

PR #2880, opened 2021-12-27
build script: FP and Indices conversion to TypeScript

fturmel:
I tried to simplify things where I could.

The output of these updated scripts wasn't committed because every index file will be re-created (the file name in the generated comment changed to `.ts`). I'll leave it to the maintainers to handle this before the next release, it might be easier to review that way.


leshakoss (approved):
⭐️

PR #2882, opened 2021-12-27
`bs` locale TypeScript conversion

fturmel:
(no `date-fns-typescriptify` issue for this locale TS migration)

Also fixes narrow month parser issue.


tan75 (approved):
Thank you! 🙏

Issue #2831, opened 2021-12-08
GitHub new issue template not active?


fturmel:
@tan75 It looks like there is an issue template in the repo (#2653, https://github.com/date-fns/date-fns/blob/master/.github/ISSUE_TEMPLATE/issue-report.md) but it doesn't seem to be active when a new issue is created from the GitHub UI.

<img width="1374" alt="Screen Shot 2021-12-08 at 10 52 14 AM" src="https://user-images.githubusercontent.com/4009209/145239465-a64895d2-a597-4879-a149-d3404964f0cd.png">

The `js` identifier should also be on the same line as ` ``` ` in the markdown file.
https://github.com/date-fns/date-fns/blob/e18ff095665e386c77ab79137592c36c5e6183a7/.github/ISSUE_TEMPLATE/issue-report.md?plain=1#L7-L9

tan75:
hi @fturmel 

Yes, you are right - currently, this template is not available when creating an issue directly from the Github web app and will need to be fixed with the js identifier.
This template comes into play when you create an issue from the doc site e.g. :
https://date-fns.org/v2.27.0/docs/formatDistanceToNow. => Report a but in the function
[...]

PR #2893, opened 2021-12-30
Fix typo in issue template

aviskarkc10:


fturmel:
The issue template is not active, per #2831. But the typo should be fixed for when it gets activated.

fturmel (approved):

PR #2899, opened 2022-01-06
Upgrade husky and lint-staged

fturmel:
Upgrades `husky`, `lint-staged` and `prettier` to latest versions, adds pre-commit hook and setup instructions.

Similar to #2319 but with newer deps, additional documentation and the `prepare` script to create the hook.

This also has the (huge imo) benefit of removing `date-fns` v1 as a transitive dependency in the project itself, as pointed out in #2476


fturmel:
Doesn't apply to v3

PR #2901, opened 2022-01-07
Test TypeScriptification chores

leshakoss:


fturmel:
@leshakoss if you're wondering why the tests are failing, `this` doesn't refer to the same scope inside a function vs inside an arrow function.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

> [...] does not have its own bindings to **this** or **super**, and should not be used as methods.

https://mochajs.org/#arrow-functions

> Passing arrow functions (aka “lambdas”) to Mocha is discouraged. Lambdas lexically bind **this** and cannot access the Mocha context.

leshakoss:
This PR continues work started in #2476 and #2005

fturmel:
Can all the `// @flow` lines be removed from the test files?

fturmel:
search for `var ` in test files, there are still 38 in there.

fturmel:
If unit tests are not executed with Flow anymore, can the `/flow-typed` directory be removed? It only seems to reference types related to sinon, benchmark and mocha.

fturmel:
Only 16 test.js files left in /src, maybe this PR is a good place to finish converting at least the test files?

Screen Shot 2022-01-07 at 12 17 02 PM

fturmel:
`assert.deepEqual` is being used a lot still (over 600 times), but has been deprecated since node v9.9.0

Screen Shot 2022-01-07 at 12 27 02 PM

fturmel:
Figured out why the tests are failing. That was a tricky one!

`testWithoutLocales.js` is failing to locate any tests at this point because it's only looking for `test.js` patterns and you just converted the remaining ones to TypeScript.

https://github.com/date-fns/date-fns/blob/fa924c5b14661bf6ef13d22a17caf86cfbae3113/testWithoutLocales.js#L1

https://github.com/date-fns/date-fns/blob/fa924c5b14661bf6ef13d22a17caf86cfbae3113/test.js#L1

fturmel:
Also, since the package.json has changed, you should run `yarn install` and commit the lock file

fturmel:
Hope you don't mind, I pushed these last two things quickly to see if that's all that was needed for the tests to pass.

leshakoss:
@fturmel thank you so much for your help! ❤️

fturmel (approved):

fturmel:
You're welcome, that was fun. BTW I noticed with this last set of commits that CI checks were running twice. Just opened a PR that I hope will fix it. #2902

tan75 (changes requested):
All is good but those 2 minor type imports if Im correct

tan75 (approved):
👍  👍  👍

PR #2924, opened 2022-01-18
formatDistance: fixes suffix for AZ

cpapazoglou:
This PR fixes the `before`, `after` suffix for AZ locale.

**Why**
In Azerbaijani the `before` , `after` translations were flipped

**[English Version](https://github.com/date-fns/date-fns/blob/master/src/locale/en-US/_lib/formatDistance/index.ts)**
```
if (options.comparison && options.comparison > 0) {
  return 'in ' + result
} else {
  return result + ' ago'
}
```

**Translation**
SS 2022-01-19 at 00 05 26

### DIFF
--- a/src/locale/az/_lib/formatDistance/index.ts
+++ b/src/locale/az/_lib/formatDistance/index.ts
@@ -100,9 +100,9 @@ const formatDistance: FormatDistanceFn = (token, count, options) => {
 
   if (options?.addSuffix) {
     if (options.comparison && options.comparison > 0) {
-      return result + ' əvvəl'
-    } else {
       return result + ' sonra'
+    } else {
+      return result + ' əvvəl'
     }
   }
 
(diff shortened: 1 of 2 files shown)


fturmel:
I don't speak the language, but this seems to make sense. Could you run `yarn locale-snapshots` and commit the updated snapshot.md file?

Curious why the Unicode CLDR has `öncə` and `ərzində` as past and future for relative units. Any insights there?
https://unicode-org.github.io/cldr-staging/charts/40/summary/az.html

cpapazoglou:
> I don't speak the language, but this seems to make sense. Could you run `yarn locale-snapshots` and commit the updated snapshot.md file?

Done

> Curious why the Unicode CLDR has `öncə` and `ərzində` as past and future for relative units. Any insights there? https://unicode-org.github.io/cldr-staging/charts/40/summary/az.html

I am not sure either but my Azerbaijani colleague confirmed that both `sonra` and `əvvəl` are valid.

fturmel (approved):

cpapazoglou:
@fturmel can you let me know the process for this to be merged?

fturmel:
One of the maintainers need to approve and merge to master, I'm just a contributor with no write permission on the main branch. It's hard to predict when this might happen, so hang tight.

cpapazoglou:
Hey @leshakoss, I hope you are OK! Just pinging you here in case you missed the notification for merging this. Let me know if there something more I should do.

hg-pyun (approved):

PR #2940, opened 2022-02-01
Changed scripts bangs to work without having `ts-node` installed globally

JuanM04:


fturmel (approved):
Good catch! I didn't even realize I had `ts-node` installed globally.

blimmer:
Sounds good! I think this PR might be preferred to the shebang since it's handles different yarn binary paths transparently.

fturmel:
Thoughts on making the shebang `#!/usr/bin/env yarn ts-node` instead?

JuanM04:
I think the shebang works even if you are using NPM (not confirmed)

fturmel:
`npm` and `pnpm` use a `.bin` directory too, same path.

fturmel:
Looks like a fix for this issue snuck into #3069 and got merged already

PR #2941, opened 2022-02-01
Fixed constants TS typings

JuanM04:
Re-added `src/constants/index.d.ts` and added module definitions to `typings.d.ts`

### DIFF
--- a/scripts/build/_lib/typings/typeScript.js
+++ b/scripts/build/_lib/typings/typeScript.js
@@ -192,6 +192,23 @@ function getTypeScriptLocaleIndexModuleDefinition(submodule, locales) {
   }
 }
 
+function getTypeScripttConstantsModuleDefinition(constants, fnSuffix) {
+  const moduleName = `date-fns/constants${fnSuffix}`
+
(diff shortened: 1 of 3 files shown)


scripts/build/_lib/typings/typeScript.js
```diff
@@ -192,6 +192,23 @@ function getTypeScriptLocaleIndexModuleDefinition(submodule, locales) {
   }
 }
 
+function getTypeScripttConstantsModuleDefinition(constants, fnSuffix) {
```
fturmel, line 195: Little typo here, two `t`s in TypeScript

fturmel (changes requested):
Nice! I also had noticed that the constants `.d.ts` file wasn't automatically generated.

scripts/build/_lib/typings/typeScript.js
JuanM04: Nice catch, thanks

fturmel (approved):

leshakoss (approved):
Thank you so much!

PR #2942, opened 2022-02-01
`differenceInBusinessDays` improvements (docs and options)

fturmel:
Following the discussion and troubleshooting here [#1268#issuecomment-1024893923](https://github.com/date-fns/date-fns/issues/1268#issuecomment-1024893923), this PR tries to improve the function's documentation and give better examples.

It also introduces an options object to customize whether the start and end dates should be included in the calculation. Defaults remain the same, so this is not a breaking change.


IgnatG:
> @IgnatG let me know what you think

Beautiful!

PrimulaX:
Can we merge this beauty, guys?

PR #2947, opened 2022-02-03
fix: yarn run ts-node in build.sh

blimmer:
Without this change, when I run the build script, I get this error:

```
 ./scripts/build/build.sh
+ ./scripts/build/docs.js
+ ./scripts/build/fp.ts
env: ts-node: No such file or directory
```

By specifying `yarn ts-node`, it'll enforce usage of the `ts-node` version defined in `package.json` and `yarn.lock`.

### DIFF
--- a/scripts/build/fp.ts
+++ b/scripts/build/fp.ts
@@ -1,4 +1,4 @@
-#!/usr/bin/env ts-node
+#!/usr/bin/env yarn ts-node
 
 /**
  * @file
--- a/scripts/build/indices.ts
+++ b/scripts/build/indices.ts
@@ -1,4 +1,4 @@
-#!/usr/bin/env ts-node
+#!/usr/bin/env yarn ts-node
 
 /**
  * @file


fturmel:
Thanks! There already is a pending PR to fix the issue #2940

fturmel:
> Sounds good! I think this PR might be preferred to the shebang since it's handles different yarn binary paths transparently.

(Adding your comment in the right PR)

Yeah, I kind of agree. That was my initial approach too. I introduced this issue by emulating the shebang in the other JS files. It worked locally, but I didn't realize it was because I had `ts-node` as a transitive global dep 🤦‍♂️

Wish I could merge one of the PRs, but I don't have write permissions on the main branch.

fturmel:
Actually, what if the shebang was `#!/usr/bin/env yarn ts-node`? Wouldn't that address your point while still make the file executable?

blimmer:
@fturmel - whoops, you're right, I put the comment on the wrong PR. I tested out the `#!/usr/bin/env yarn ts-node` and it worked great. I've updated this PR. I that's better than the `./node_modules` approach, personally. That way, `yarn` can resolve the path however it sees fit.

fturmel:
Looks like a fix for this issue snuck into #3069 and got merged already

Issue #2968, opened 2022-02-20
formatISO does not support milliseconds


fregante:
I'm looking to use `formatISO` to preserve the time zone, but it does not match the ISO used by JavaScript itself and there doesn't seem to be an option to enable it:

```js
const date = new Date();

console.log(date.toISOString())
// "2022-02-20T04:00:45.418Z"

console.log(_.formatISO(date))
// "2022-02-20T12:00:45+08:00" <-- missing 418 milliseconds
```

mandaputtra:
I think this is same as https://github.com/date-fns/date-fns/issues/2965

fregante:
This is not a documentation issue. That issue does not mention milliseconds at all

mandaputtra:
Yes from the title, but this sentence are basically the same. 

> Executing that very code in the console comes back with '19:00:52-07:00' and not '19:00:52Z'.

fregante:
Can you reopen this issue while you investigate? Regardless of what it "looks like" this is a feature request / inconsistency report.

fregante:
Terima kasih mas 🙏

mandaputtra:
I'm wrong, after looking up, turns out milliseconds aren't implemented yet. Sorry. Please review my draft if it is okay or not.

PR #2971, opened 2022-02-27
format ISO doesnt support milliseconds

mandaputtra:
Issue #2968 mentioning `formatISO` doesn't support millisecond yet. The JS `Date.toISOString()` return millisecond with `.` (dot)

```
new Date().toISOString() // 2022-02-27T14:26:49.139Z
```

But after reading from Wikipedia https://en.wikipedia.org/wiki/ISO_8601, I can't find any implementation on millisecond. Should we do this by default? Or we can use options?

```
formatISO(date, { withMillisecond: true })
```

### DIFF
--- a/src/_lib/fractionalDigits/index.ts
+++ b/src/_lib/fractionalDigits/index.ts
@@ -1,0 +1,16 @@
+import addLeadingZeros from '../addLeadingZeros'
+
+export default function fractionDigits(
+  fractionDigits: number,
+  originalDate: Date
+): string {
+  let fractionalSecond = ''
+  if (fractionDigits > 0) {
+    const milliseconds = originalDate.getMilliseconds()
+    const fractionalSeconds = Math.floor(
+      milliseconds * Math.pow(10, fractionDigits - 3)
+    )
+    fractionalSecond = '.' + addLeadingZeros(fractionalSeconds, fractionDigits)
+  }
+  return fractionalSecond
+}
--- a/src/_lib/fractionalDigits/test.ts
+++ b/src/_lib/fractionalDigits/test.ts
@@ -1,0 +1,24 @@
+/* eslint-env mocha */
+
+import assert from 'assert'
+import fractionalDigits from '.'
+
+describe('fractionalDigits', () => {
+  it('it adds fractional 4 digits', () => {
+    const date = new Date(2019, 2 /* Mar */, 3, 19, 0, 52, 123)
+    const fraction = fractionalDigits(4, date)
+    assert(fraction === '.1230')
+  })
+
+  it('it only show 2 fractional digits', () => {
+    const date = new Date(2019, 2 /* Mar */, 3, 19, 0, 52, 123)
+    const fraction = fractionalDigits(2, date)
+    assert(fraction === '.12')
+  })
+
+  it('it doesnt show any digits', () => {
+    const date = new Date(2019, 2 /* Mar */, 3, 19, 0, 52, 123)
+    const fraction = fractionalDigits(0, date)
+    assert(fraction === '')
+  })
+})
--- a/src/formatISO/index.ts
+++ b/src/formatISO/index.ts
@@ -24,23 +24,23 @@ export interface FormatISOOptions
  *
  * @example
  * // Represent 18 September 2019 in ISO 8601 format (local time zone is UTC):
- * const result = formatISO(new Date(2019, 8, 18, 19, 0, 52))
- * //=> '2019-09-18T19:00:52Z'
+ * const result = formatISO(new Date(2019, 8, 18, 19, 0, 52, 123))
+ * //=> '2019-09-18T19:00:52.123Z'
  *
  * @example
  * // Represent 18 September 2019 in ISO 8601, short format (local time zone is UTC):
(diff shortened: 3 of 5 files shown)


fregante (approved):

fregante:
> But after reading from Wikipedia [en.wikipedia.org/wiki/ISO_8601](https://en.wikipedia.org/wiki/ISO_8601), I can't find any implementation on millisecond. Should we do this by default? Or we can use options?

It seems that some languages include milliseconds while others do not. It's possible that the milliseconds are implied as part of the seconds: "seconds with decimals".

We should probably rely on EcmaScript’s specification and usage of the standard: https://tc39.es/ecma262/#sec-date-time-string-format

> ECMAScript defines a string interchange format for date-times based upon a simplification of the ISO 8601 calendar date extended format. The format is as follows: `YYYY-MM-DDTHH:mm:ss.sssZ`

imballinst (commented):
Thanks for adding the implementation! When you have the time, could you also update the tests for the IST as well?

With regards to the ISO8601 implementation, I _think_ we can proceed with it -- @fregante mentioned that the ECMAScript specification is there, plus this another w3 one: https://www.w3.org/TR/NOTE-datetime

> The formats are as follows. Exactly the components shown here must be present, with exactly this punctuation. Note that the "T" appears literally in the string, to indicate the beginning of the time element, as specified in ISO 8601.
> 
>    Year:
>       YYYY (eg 1997)
>    Year and month:
>       YYYY-MM (eg 1997-07)
>    Complete date:
>       YYYY-MM-DD (eg 1997-07-16)
>    Complete date plus hours and minutes:
>       YYYY-MM-DDThh:mmTZD (eg 1997-07-16T19:20+01:00)
>    Complete date plus hours, minutes and seconds:
>       YYYY-MM-DDThh:mm:ssTZD (eg 1997-07-16T19:20:30+01:00)
>    Complete date plus hours, minutes, seconds and a decimal fraction of a
> second
>       YYYY-MM-DDThh:mm:ss.sTZD (eg 1997-07-16T19:20:30.45+01:00)
> 
> where:
> 
>      YYYY = four-digit year
>      MM   = two-digit month (01=January, etc.)
>      DD   = two-digit day of month (01 through 31)
>      hh   = two digits of hour (00 through 23) (am/pm NOT allowed)
>      mm   = two digits of minute (00 through 59)
>      ss   = two digits of second (00 through 59)
>      s    = one or more digits representing a decimal fraction of a second
>      TZD  = time zone designator (Z or +hh:mm or -hh:mm)
> 
> This profile does not specify how many digits may be used to represent the decimal fraction of a second. An adopting standard that permits fractions of a second must specify both the minimum number of digits (a number greater than or equal to one) and the maximum number of digits (the maximum may be stated to be "unlimited").

My suggestion: add an optional `options.fractionDigits` that is a number, which defaults to 0. The reasons:

1. We have `options.fractionDigits` in another format function: https://date-fns.org/v2.28.0/docs/formatRFC3339, so I think options similarity would be great
2. It would enable porting to another format better, e.g. the Oracle database can specify 0-6 decimal millisecond digits behind the seconds: https://docs.oracle.com/en/database/oracle/oracle-database/21/adjsn/iso-8601-date-time-and-duration-support.html.
3. Defaulting it to 0 will make the previous versions' behavior consistent when people are updating `date-fns` (because this is a behavior change and perhaps can be considered "breaking" in a way)

What do you think @tan75 @leshakoss?

mandaputtra:
> When you have the time, could you also update the tests for the IST as well?

I'm still confused on how to update the IST test, could you point me out where?

imballinst:
@mandaputtra I think it's here:

https://github.com/date-fns/date-fns/blob/928172c0262a378871bc9afb84e65558fa51c822/test/formatISO/india.js#L1-L13

But, I believe the test should pass if we provide non-breaking options (default the fraction to `0`).

mandaputtra:
Added the fractional digits function 💯 tell me if there are anything else to change

imballinst (approved):
Looks good to me! I have only one comment with regards to the docs.

AJStacy:
+1 to merging this PR.

Glandos:
Some background in other languages regarding subsecond precision

defy93:
Want to bump this, its a big issue for our timestamps in a code base i work on.

mandaputtra:
Just rebased the PR. Please kindly check again

Nealsoni00:
Assuming milliseconds are still not supported? Would amazing to have this support for any application that relies on time based sorting client side.

bartelemi:
Hello, it would be amazing to get this PR merged. If it's become stale for any reason, I'm happy to create a fresh one and address any outstanding comments.

mandaputtra:
Hi, I know this is a long standing PR lol. I'm okay if you create a fresh one. Go ahead 🚀

kossnocorp:
Hey @mandaputtra! Sorry for taking so long to get to it. It is my fault. Unless you want to do it yourself, I'm happy to continue and get it to release.

mandaputtra:
So I just need to resolve this conflict right? I'm happy to do that

AJStacy:
Any status update on getting this merged?

Issue #2926, opened 2022-01-22
Value precision when calculating years


Mondal10:
Currently, I cannot find any direct way to get the precise value of a year when we convert either months to year value or when we get the difference between two dates, it either returns me a whole number that is Math.round or Math.floor or Math.ceil,  what if I want decimal or fractional value?

For eg:
monthsToYears(9) return 0 instead I need something like 0.75
monthsToYears(18) return 1 instead I need something like 1.5

It will be great if anyone can help here.

One way would be to update the `monthsToYears` function by passing an optional boolean parameter that will decide on whether to pass direct division result or Math.floor(result). Just wanted to discuss this approach and what can affect it. If it sounds good then I can raise a PR for the same.

PR #3006, opened 2022-03-31
feat: add decimals options to monthsToYears()

Alanscut:
linked #2926 
add options let users can get number include decimals point，also can set digits after the decimal point.

```
 * const result = monthsToYears(40, {decimals: true})
 * //=> 3.33
 * result = monthsToYears(40, {decimals: true, digits: 4})
 * //=> 3.3333
```

pr is add a new options, so this change will not affect any existing code


Alanscut:
@tan75 @leshakoss 
please check

Issue #3007, opened 2022-03-31
locale en-ZA has wrong short date format


jeni-1999:
Locale en-ZA have wrong short date format. It should be yyyy/mm/dd.

PR #3014, opened 2022-04-14
Fix en-ZA Date and Time default formats

alexandresaura:
Issue: https://github.com/date-fns/date-fns/issues/3007

Site used as reference: https://metacpan.org/dist/DateTime-Locale/view/lib/DateTime/Locale/en_ZA.pod

Screenshots of Intl.DateTimeFormat as another proof:
image
image


alexandresaura:
I didn't know about recreate the snapshot.md, so I pushed another commit with the fix.

tan75 (approved):
Thank you!

PR #3045, opened 2022-05-16
Fix weekday matcher in `tr` locale

### LINKED ISSUE #3028: date-fns turkish problem
` 
console.log(toString(parse('Pazartesi', 'eeee', new Date(), { locale: tr })))`

so when i try something like this, it gives Invalid Date error, but if i try something like this:

` 
console.log(toString(parse('Salı', 'eeee', new Date(), { locale: tr })))`

It works correctly, What could be the problem ? Pazartesi means Monday in English, Salı means Tuesday.

fturmel:
Closes #3028


tan75 (approved):
Thank you! 🙏

PR #3057, opened 2022-05-19
Fix zh-tw translations

tso1158687:
The zh-tw translation about quarter is totally wrong. 
And the vocabulary about anno is old style way, so I adjust it to modern style


tan75:
hi @date-fns/i18n-zh-tw 
can you please review this PR?

PR #3069, opened 2022-05-31
Implement getDefaultOptions, setDefaultOptions

leshakoss:


src/endOfWeek/index.ts
```diff
@@ -32,21 +33,17 @@ import type { LocaleOptions, WeekStartOptions } from '../types'
...
-    options.weekStartsOn == null
-      ? defaultWeekStartsOn
-      : toInteger(options.weekStartsOn)
+  const weekStartsOn = toInteger(
+    options?.weekStartsOn ??
+      options?.locale?.options?.weekStartsOn ??
+      _defaultOptions.weekStartsOn ??
+      _defaultOptions.locale?.options?.weekStartsOn ??
+      0
+  )
```
kossnocorp, line 46: I'm comparing this PR to https://github.com/date-fns/date-fns/pull/3087 and this one, with the wider scope, feels more future-prone.

If we decide to move locales to separate packages and make `Intl` the default approach, the change will still make sense. Furthermore, we might think of option autodetection and apply user-specific defaults for `weekStartsOn`, which would be impossible if we stick to locale-only.

It is indeed more code, but we can approach it slightly differently and reduce it to something like:

```suggestion
  const weekStartsOn = getOption(options, 'weekStartsOn')
```

That will also help to abstract the default behavior that we repeat from function to function.

src/_lib/defaultOptions/index.ts
```diff
@@ -0,0 +1,39 @@
...
+    format?: FormatOptions['format'] | FormatDurationOptions['format']
+  } & FormatRFC3339Options &
+  AdditionalDigitsOptions &
+  AdditionalTokensOptions &
+  FormatDistanceOptions &
+  FormatDistanceStrictOptions &
+  RoundToNearestMinutesOptions &
+  AreIntervalsOverlappingOptions
+
+export let _defaultOptions: AllOptions = {}
```
kossnocorp, line 35: This approach doesn't feel safe, let's use a getter function instead. Module exports are supposed to be static if I'm not mistaken and modifying it might break them. Also `getOptions` looks better than `_defaultOptions`.

src/_lib/defaultOptions/index.ts
```diff
@@ -0,0 +1,39 @@
...
+  RepresentationOptions &
+  Omit<FormatDurationOptions, 'format'> & {
+    format?: FormatOptions['format'] | FormatDurationOptions['format']
+  } & FormatRFC3339Options &
+  AdditionalDigitsOptions &
+  AdditionalTokensOptions &
+  FormatDistanceOptions &
+  FormatDistanceStrictOptions &
+  RoundToNearestMinutesOptions &
+  AreIntervalsOverlappingOptions
```
kossnocorp, line 33: Maybe we don't need to make **all** options global, but `locale`, `weekStartsOn`, etc.

kossnocorp (changes requested):

src/_lib/defaultOptions/index.ts
leshakoss: Reduced the scope to the big three: locale, weekStartsOn, firstWeekContainsDate

src/formatDuration/index.ts
```diff
@@ -83,18 +77,19 @@ interface Options {
...
 
-  const format = options?.format || defaultFormat
-  const locale = options?.locale || defaultLocale
-  const zero = options?.zero || false
-  const delimiter = options?.delimiter || ' '
+  const defaultOptions = getDefaultOptions()
+  const locale = options?.locale ?? defaultOptions.locale ?? defaultLocale
+  const format = options?.format ?? defaultFormat
+  const zero = options?.zero ?? false
+  const delimiter = options?.delimiter ?? ' '
```
kossnocorp, line 92: I'm afraid that this can be a breaking change, but I don't mind it haha

src/formatISO/test.ts
```diff
@@ -6,7 +6,7 @@ import formatISO from '.'
 import addLeadingZeros from '../_lib/addLeadingZeros'
 
 // This makes sure we create the consistent offsets across timezones, no matter where these tests are ran.
-function generateOffset(originalDate: Date) {
+export function generateOffset(originalDate: Date) {
```
kossnocorp, line 9: Maybe move it to the `src/_lib/test/index.ts`?

kossnocorp (approved):
Look goods except for exsessive usage of `??` that might alter behavior. I don't mind shipping it as is because it's better for the future.

olee:
When might we expect a new version to be published with those changes, if I dare to ask? :-)

leshakoss:
This closed  #816

leshakoss:
@olee published in v2.29: https://github.com/date-fns/date-fns/releases/tag/v2.29.1

PR #3087, opened 2022-06-10
Implement getDefaultLocale, setDefaultLocale

leshakoss:
This PR attempts to solve the same problem as #3069 but in narrower scope


leshakoss:
Deprecated by #3069

PR #3088, opened 2022-06-11
Remove global types from typings.d.ts

### LINKED ISSUE #1472: [TS]: Locale and Interval types are global
In our application, we have a global `type Locale = 'en' | 'de'` that let us use it without importing it. I know this is not really clean...

However, `date-fns` typings definition also have a global `type Locale` (and also `Interval`): https://github.com/date-fns/date-fns/blob/master/typings.d.ts#L35-L62

I think this is not good as we should not be able to use the `Locale` from date-fns without importing it explicitly.

Suggestion:

Instead of this definition,
```ts
type Interval = {
  start: Date | number
  ...
}
type IntervalAliased = Interval

type Locale = {
  formatDistance: Function
  ...
}
type LocaleAliased = Locale

// Exported Type Aliases

declare module 'date-fns' {
  export type Interval = IntervalAliased

  export type Locale = LocaleAliased
}

// Regular Functions

declare module 'date-fns' {
  function addBusinessDays(date: Date | number, amount: number): Date
  namespace addBusinessDays {}
   ...

}
```

use 

```ts
declare module 'date-fns' {
  export type Interval = {
    start: Date | number
    ...
  } 

  export type Locale = {
    formatDistance: Function
    ...
  }

  function addBusinessDays(date: Date | number, amount: number): Date
  namespace addBusinessDays {}

  ...
}
```

What do you think?

smmoosavi:
move global types inside `declare module` in typings.d.ts

please review with hiding white space [flag](https://github.com/date-fns/date-fns/pull/3088/files?diff=split&w=1)

fixes #1472

### DIFF
--- a/scripts/build/_lib/typings/typeScript.js
+++ b/scripts/build/_lib/typings/typeScript.js
@@ -52,26 +52,9 @@ function getTypeScriptTypeAlias(type) {
   const { title, properties, content } = type
 
   return formatBlock`
-    type ${title} = ${
+    export type ${title} = ${
     properties ? getParams(properties) : content.type.names.join(' | ')
   }
-    type ${title}Aliased = ${title}
-  `
-}
-
-function getExportedTypeScriptTypeAlias(type) {
-  const { title } = type
-
-  return formatBlock`
-    export type ${title} = ${title}Aliased
-  `
-}
-
-function getExportedTypeScriptTypeAliases(aliases) {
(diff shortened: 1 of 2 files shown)


smmoosavi:
can anybody review this?

justinbhopper:
This appears unnecessary now and appears to already be fixed in latest master.

PR #3089, opened 2022-06-13
UTCDate integration

kossnocorp:
- This integrates the `UTCDate` class (https://github.com/date-fns/utc) that replaces internal UTC functions making the build size smaller.
- It makes all functions that accept `Date` as an argument generic. This allows using any class that extends `Date`.
- It also updates TypeScript config and forces using `index` in imports


src/closestTo/index.ts
```diff
@@ -23,9 +23,9 @@ import requiredArgs from '../_lib/requiredArgs/index'
  * ])
  * //=> Tue Jan 01 2030 00:00:00
  */
-export default function closestTo(
-  dirtyDateToCompare: Date | number,
-  dirtyDatesArray: Array<Date | number>
+export default function closestTo<DateType extends Date>(
+  dirtyDateToCompare: DateType | number,
+  dirtyDatesArray: Array<DateType | number>
 ): Date | undefined {
```
leshakoss, line 29: Missed the return type

src/max/index.ts
```diff
@@ -23,7 +23,9 @@ import requiredArgs from '../_lib/requiredArgs/index'
  * ])
  * //=> Sun Jul 02 1995 00:00:00
  */
-export default function max(dirtyDatesArray: Array<Date | number>): Date {
+export default function max<DateType extends Date>(
+  dirtyDatesArray: Array<DateType | number>
+): DateType | Date {
```
leshakoss, line 28: Can this be just `DateType`? Also in `min` and `clamp`

leshakoss (dismissed):
What I found so far

src/max/index.ts
kossnocorp: Nope, if there's no elements, we'll return `new Date(NaN)`

gkostov:
Hey guys, is this work going to be released soon? (hopefully as part of version 2)

As I understand it, it mostly improves the internal functionality of the library but luckily it happens to also open up the library for a solution that we're trying to come up with in relation to working with arbitrary timezones. So I'm looking for a way to get this down with an official release.

Cheers

PR #3090, opened 2022-06-13
Added optional flag removeZeros to remove values with 0 from formatISODuration

### LINKED ISSUE #3060: Add option to hide zeros in ISO formatted durations
As of now formatISODuration represents every allowed field in ISO duration:

https://github.com/date-fns/date-fns/blob/fd6bb1a0bab143f2da068c05a9c562b9bee1357d/src/formatISODuration/index.ts#L36-L45

However it may be a bit verbose to have the whole string filled with zeros when the ISO-8601 allows them to be removed (`P0Y0M1DT0H0M0S` is equivalent to  `P1D`).

Balastrong:
This PR fixes #3060.

Added an optional flag that gets rid of all the values with 0 to have a shorter (but still valid) ISO duration string.

The case where only `minutes` are present is also handled, rendering them with `TxM` instead of the usual `xM` to avoid ambiguity with `months`.

### DIFF
--- a/src/formatISODuration/index.ts
+++ b/src/formatISODuration/index.ts
@@ -10,6 +10,7 @@ import requiredArgs from '../_lib/requiredArgs/index'
  * Format a duration object according to the ISO 8601 duration standard (https://www.digi.com/resources/documentation/digidocs/90001437-13/reference/r_iso_8601_duration_format.htm)
  *
  * @param {Duration} duration - the duration to format
+ * @param {boolean} removeZeros - if true, values with 0 are removed from the formatted string
  *
  * @returns {String} The ISO 8601 duration string
  * @throws {TypeError} Requires 1 argument
@@ -26,8 +27,21 @@ import requiredArgs from '../_lib/requiredArgs/index'
  *   seconds: 0
  * })
  * //=> 'P39Y2M20DT0H0M0S'
+ *
+ * @example
+ * // Format the given duration as ISO 8601 string with trailing zeros removed
+ * const result = formatISODuration({
+ *   years: 39,
+ *   months: 0,
+ *   days: 0,
+ *   hours: 7
+ * });
+ * //=> 'P39Y7H'
  */
-export default function formatISODuration(duration: Duration): string {
+export default function formatISODuration(
+  duration: Duration,
+  removeZeros: boolean = false
+): string {
   requiredArgs(1, arguments)
 
   if (typeof duration !== 'object')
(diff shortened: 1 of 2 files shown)


src/formatISODuration/index.ts
```diff
@@ -42,5 +56,36 @@ export default function formatISODuration(duration: Duration): string {
...
+      ]
+
+      return (
+        'P' +
+        chunks
+          .filter(({ value }) => value > 0)
+          .map(({ suffix, value }) => `${value}${suffix}`)
+          .join('')
+      )
+    }
```
sevdog, line 87: I do not belive this fully reflects the ISO-8601 standard. For what I have read so far the `T` separator is mandatory to recognize the "time part" from the "days part".

This code will parse a duration of 10 hours to "P10H" while other formatters will use "PT10H". I think that many parsers would not recognize that as a valid ISO duration.

My suggestio in to build the formatted string from two pieces like:

```javascript
`P${dayPart}T${timePart}`
```

Balastrong: Thank you very much for the comment and the reference! You're indeed right, some fix is needed here.

Would `P10DT` be a valid date? If I got that right I should probably get rid of the T if no time is present.

sevdog: For what I have seen so far when the `T` is present it should be followed by the time part. Thus `P10DT` may be considered invalid by some parsers depending on the used regex.

The RFC states that the `T` must be followed by at least one time part.

TL; DR: yes, the `T` can be omitted if there are no hours/minutes/seconds.

Balastrong: Great, thank you! I added a commit to fix that :)

src/formatISODuration/index.ts
```diff
@@ -26,8 +27,21 @@ import requiredArgs from '../_lib/requiredArgs/index'
...
+ *
+ * @example
+ * // Format the given duration as ISO 8601 string with trailing zeros removed
+ * const result = formatISODuration({
+ *   years: 39,
+ *   months: 0,
+ *   days: 0,
+ *   hours: 7
+ * });
+ * //=> 'P39Y7H'
```
sevdog, line 39: Just found this leftover here:

```suggestion
 * //=> 'P39YT7H'
```

Balastrong: Good catch!

sevdog (approved):
For me it is fine.

esoca:
Hello !

Any change to get this merged ?

gregholvoet (approved):

PR #3091, opened 2022-06-13
Added support for ceil and floor in roundToNearestMinutes

### LINKED ISSUE #3063: Add support for "ceil" or "floor" based rounding
Hey folks, first off, thanks for the great library!

I'm looking to round numbers down to the nearest 15 minute interval. `roundToNearestMinutes` does rounding, but unfortunately doesn't have an option for rounding up or down.

What are your thoughts on adding the ability to do this?

```js
// Rounding (current functionality)
roundToNearestMinutes(/* 12:03 */, { nearestTo: 15 }) // 12:00
roundToNearestMinutes(/* 12:14 */, { nearestTo: 15 }) // 12:15

// Rounding down (floor)
roundToNearestMinutes(/* 12:03 */, { nearestTo: 15, mode: 'floor' }) // 12:00
roundToNearestMinutes(/* 12:14 */, { nearestTo: 15, mode: 'floor' }) // 12:00

// Rounding up (ceil)
roundToNearestMinutes(/* 12:03 */, { nearestTo: 15, mode: 'ceil' }) // 12:15
roundToNearestMinutes(/* 12:14 */, { nearestTo: 15, mode: 'ceil' }) // 12:15
```

Balastrong:
This PR fixed #3063 

Added an optional field to options on `roundToNearestMinutes` to toggle between `Math.floor` and `Math.ceil`, with default on `floor`.


tan75:
@Balastrong 
`normalizeDuration` is this a new function? where is this used?

src/types.ts
```diff
@@ -116,6 +116,7 @@ export interface FormatDurationOptions {
 
 export interface RoundToNearestMinutesOptions {
   nearestTo?: number
+  mode?: 'floor' | 'ceil'
```
tan75, line 119: Instead of adding `mode` here please reuse the code from here:
https://github.com/date-fns/date-fns/blob/master/src/_lib/roundingMethods/index.ts

tan75 (dismissed):
A good idea 🙏
Please change this function to use the existing rounding options.

Balastrong:
> @Balastrong `normalizeDuration` is this a new function? where is this used?

Ah I was doing some research for #3092 but I didn't mean to commit them on this branch, I'll remove that commit.

src/types.ts
Balastrong: Ah, I didn't see that, thank you!

PR updated :)

leshakoss (approved):
Thanks!

PR #3115, opened 2022-07-14
Added Belarusian Classic (Tarask) language support

nopears:
please accept it, we really need classic spelling of our native language


tan75 (approved):
Thank you! 🙏

leshakoss:
As far as I understand the standard: https://en.wikipedia.org/wiki/IETF_language_tag the correct language tag is `be-tarask`. I'll merge it as it is and rename it in master. Thanks for the PR!

PR #3123, opened 2022-07-22
Removed duplicate declaration of const daysInWeek

swetangdev:
**const daysInWeek: number;**  _declared twice_ throwing error of redeclare variable.


fturmel:
#3125 should take care of this

PR #3124, opened 2022-07-22
Amended multiple redeclaration of daysInWeek

swetangdev:
**const daysInWeek: number;**  _declared twice_ at multiple places,  throwing error of redeclare variable.


fturmel:
Thanks for giving it a shot! #3125 should properly fix the root cause.

Issue #3150, opened 2022-08-10
parseISODuration function


matinzd:
Is there any plan to add functionality to parse ISO duration? 
Reverse of https://date-fns.org/v2.29.1/docs/formatISODuration.

Examples:
```js
parseISODuration('P3Y6M4DT12H30M5S') /* {
  years: 3,
  months: 6,
  days: 4,
  hours: 12,
  minutes: 30,
  seconds: 5
}*/

parseISODuration('P1M') /* {
  years: 0,
  months: 1,
  days: 0,
  hours: 0,
  minutes: 0,
  seconds: 0
}*/
```

https://en.wikipedia.org/wiki/ISO_8601

davetapley:
See also: https://github.com/date-fns/date-fns/issues/418

matinzd:
@davetapley PR is waiting for review.

matinzd:
> @matinzd any idea when your PR could go live? It's a really handy method to have inside date-fns

As I communicated with @kossnocorp, this PR will be included in v3.

MKHenson:
There may be an issue with this. I tried it using a negative value and it threw an error. From what I've read online, the standard should support negative values. I tried the following value: P0Y0M0DT-5H0M0S and it threw an error

cj-christoph-gysin:
> > @matinzd any idea when your PR could go live? It's a really handy method to have inside date-fns
> 
> As I communicated with @kossnocorp, this [PR](https://github.com/date-fns/date-fns/pull/3151) will be included in v3.

It looks like v3 was released, but the PR was not included.

matinzd:
@cj-christoph-gysin Yes it was supposed to be included. Pinging @kossnocorp

kossnocorp:
Sorry, I was overwhelmed with the release to include any more new features. This will go with one of the next versions. If you need a function, I recommend to copy-paste it to your source code and look for new versions.

acocheo:
Any news?

PR #3151, opened 2022-08-10
feat: add parseISODuration

matinzd:
Related: https://github.com/date-fns/date-fns/issues/3150


matinzd:
@leshakoss @kossnocorp

matinzd:
@leshakoss @kossnocorp

alfaproject:
Is there any hope to get this in?

matinzd:
> Is there any hope to get this in?

I messaged @kossnocorp, it seems that they are waiting for v3.0.0 release so this can be included there if there won't be any bugs.

alfaproject:
Perhaps we can re-target for v2 instead of main branch?

kossnocorp:
If you need to use the function, please copy-paste the code to your project locally. It's not worth it for us to rebuild the function for v2.

alfaproject:
Alright, fair enough

alex-collibra:
First of all, thanks a lot for this workaround.
It seems like parsing of negative durations doesn't work.
```
parseISODuration('P-1D');

// results in Invalid format
```
Is it on purpose? Can this be added to this feature as well? It is supported in libs like luxon or dayjs, so technically it may be possible. Here is an example of luxon and dayjs if you want live example:
https://codesandbox.io/s/dazzling-liskov-4dqzg5?file=/src/App.tsx

matinzd:
> First of all, thanks a lot for this workaround. It seems like parsing of negative durations doesn't work.
> 
> ```
> parseISODuration('P-1D');
> 
> // results in Invalid format
> ```
> 
> Is it on purpose? Can this be added to this feature as well? It is supported in libs like luxon or dayjs, so technically it may be possible. Here is an example of luxon and dayjs if you want live example: https://codesandbox.io/s/dazzling-liskov-4dqzg5?file=/src/App.tsx

It seems negative offsets are available in the [specification](https://en.wikipedia.org/wiki/ISO_8601). I will try to update the PR accordingly.

matinzd:
> First of all, thanks a lot for this workaround. It seems like parsing of negative durations doesn't work.
> 
> ```
> parseISODuration('P-1D');
> 
> // results in Invalid format
> ```
> 
> Is it on purpose? Can this be added to this feature as well? It is supported in libs like luxon or dayjs, so technically it may be possible. Here is an example of luxon and dayjs if you want live example: https://codesandbox.io/s/dazzling-liskov-4dqzg5?file=/src/App.tsx

According to the [specification](https://en.wikipedia.org/wiki/ISO_8601#:~:text=Durations%20define%20the,number%20of%20seconds.) all these values are positive and above zero. In my understanding I would assume we should not have a negative day or negative year?

alex-collibra:
> > First of all, thanks a lot for this workaround. It seems like parsing of negative durations doesn't work.
> > ```
> > parseISODuration('P-1D');
> > 
> > // results in Invalid format
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Is it on purpose? Can this be added to this feature as well? It is supported in libs like luxon or dayjs, so technically it may be possible. Here is an example of luxon and dayjs if you want live example: https://codesandbox.io/s/dazzling-liskov-4dqzg5?file=/src/App.tsx
> 
> According to the [specification](https://en.wikipedia.org/wiki/ISO_8601#:~:text=Durations%20define%20the,number%20of%20seconds.) all these values are positive and above zero. In my understanding I would assume we should not have a negative day or negative year?

Agree, but it was the same discussion in other libraries (sorry to link to other libs from here, but this is the one of many discussions: https://github.com/moment/moment/issues/2408#issue-85650097)

cj-christoph-gysin:
Any chance to get the conflicts resolved so we can still get this into v3?

src/parseISODuration/index.ts
```diff
@@ -0,0 +1,70 @@
...
+ * //=>
+ * {
+ *   years: 3,
+ *   months: 6,
+ *   weeks: 0,
+ *   days: 4,
+ *   hours: 12,
+ *   minutes: 30,
+ *   seconds: 5,
+ * }
```
kossnocorp, line 36: Please format the whole object as comments otherwise the code highlight on the website will break

src/parseISODuration/index.ts
```diff
@@ -0,0 +1,70 @@
...
+  return Object.entries({
+    years,
+    months,
+    weeks,
+    days,
+    hours,
+    minutes,
+    seconds,
+  }).reduce((prev, [key, value]) => {
+    return { ...prev, [key]: Number(value) }
```
kossnocorp, line 68: It will be faster to mutate the object:

```suggestion
    prev[key] = +value;
    return prev;
```

kossnocorp (changes requested):
Thank you a lot for your work and waiting! I finally have time to work on new stuff and will be happy to ship it with the next version

kossnocorp:
Please rebase and reformat code with the default Prettier config, thanks!

src/parseISODuration/index.ts
matinzd: Fixed [320c1f9](https://github.com/date-fns/date-fns/pull/3151/commits/320c1f9862a1003b751eb04fc88d993c83b53d09).

src/parseISODuration/index.ts
matinzd: Fixed [320c1f9](https://github.com/date-fns/date-fns/pull/3151/commits/320c1f9862a1003b751eb04fc88d993c83b53d09).

src/parseISODuration/index.ts
```diff
@@ -0,0 +1,70 @@
...
+ * Parse the given string in ISO 8601 format and return Duration object.
+ *
+ * Function accepts duration format of ISO 8601.
+ * ISO 8601: http://en.wikipedia.org/wiki/ISO_8601
+ * Specifically: https://en.wikipedia.org/wiki/ISO_8601#Durations
+ *
+ * If the argument is not a valid ISO Duration it will throw Invalid format error.
+ *
+ * @param argument - ISO Duration string
+ * @returns duration object
```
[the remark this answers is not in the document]
matinzd: Fixed [320c1f9](https://github.com/date-fns/date-fns/pull/3151/commits/320c1f9862a1003b751eb04fc88d993c83b53d09).

src/parseISODuration/index.ts
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * @throws {RangeError} Duration in ISO format cannot be `Invalid format`
```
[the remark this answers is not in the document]
matinzd: Fixed [320c1f9](https://github.com/date-fns/date-fns/pull/3151/commits/320c1f9862a1003b751eb04fc88d993c83b53d09).

matinzd:
> Please rebase and reformat code with the default Prettier config, thanks!

Rebased now and all suggestions are applied. However I don't know how to test generated docs to make sure if the example works properly. Let me know if that still has an issue.

matinzd:
I guess we may need to add more test cases as I have heard a lot about negative values. WDYT? @kossnocorp

kossnocorp:
@matinzd don't worry about the docs. Adding more tests never hurt, if you feel there's something to cover please go ahead.

andrioid:
What is missing so this can be merged? Anything I can do to help?

matinzd:
@andrioid 
There are some test cases that the current implementation is not covering. I will try to update the PR accordingly.

matinzd:
@yjukaku Yes, you are right! The test case I found was not related to supporting negative values. Also negative values are not in the standard as far as I checked. 
I really want to be mindful when merging PR because I don't want people to use broken function in their codebase especially that date-fns is widely used in the ecosystem.

src/parseISODuration/index.ts
```diff
@@ -0,0 +1,73 @@
...
+ * //=>
+ * const result = {
+ *   years: 3,
+ *   months: 6,
+ *   weeks: 0,
+ *   days: 4,
+ *   hours: 12,
+ *   minutes: 30,
+ *   seconds: 5,
+ * }
```
ckcherry23, line 36: @matinzd I think what we need here is for the entire result object to be formatted as a comment. You can refer to `src/intervalToDuration/index.ts` for an example.

```suggestion
 * //=> { years: 3, months: 6, weeks: 0, days: 4, hours: 12, minutes: 30, seconds: 5 }
```

ckcherry23 (commented):
Hi @matinzd, can you let us know what test cases you think can be added before we merge this in? I can help to contribute as well.

ckcherry23 (commented):
Hi @matinzd, maybe we could also add more invalid test cases for this, for example:
* `P` - only P
* `PT` - only P and T
* `P1H` - missing T
* `PT1D` - unneeded T
* `P1W1YT1M` - incorrect date order before T
* `P1WT1M1H` - incorrect time order after T

We could also have:
* `P0D` - gives zero instead of undefined
* `PT0S`- gives zero instead of undefined
* `PT0.0021S` - 0.0021S (instead of 2.1 milliseconds, which could be useful for `ExtendedDuration` if supported later)

PR #3163, opened 2022-08-24
Nl locale friday

ThomasTrone:


vadimpopa:
Hi @ThomasTrone.

1. You need to change in matchDayPatterns too in `src/locale/nl/_lib/match/index.ts`, for `format` -> `parse` compatibility. Otherwise parse will return invalid date.
2. I think you need to rebase this from `v2` branch, as it can be fixed as a patch in v2.29.3. But that would have to be confirmed by @leshakoss

ThomasTrone:
Thank you, I've added those changes as well. I checked once more to be sure and it's not present anywhere else (except the snapshot.md files, which I assume are automatically generated).

vadimpopa:
Looks good, just you forgot to update nl-BE one too ?

ThomasTrone:
That is strange, I'm pretty sure I did. Must have accidentally undone it somewhere. Thanks.

ThomasTrone:
I see that v2 has now been merged with no comment or review on this PR. Am I expected to rebase this again?

Issue #3170, opened 2022-09-06
[uk] Incorrect noun grammar form for xDays, xWeeks, xMonths regular distance in Ukrainian


alinkedd:
In Ukrainian the correct noun form for regular x distance of **few** numbers (2, 3, 4 and complex with 2, 3, 4) is **Nominative Plural** in most cases, not Singular Genitive:
- xMonths regular - `4 місяці` (not `4 місяця`)
- xWeeks regular - `3 тижні` (not `3 тижня`)
- xDays regular - `2 дні` (not `2 дня`)

All other cases are correct only because:
- there is no difference in writing between Nominative Plural and Singular Genitive forms of seconds, minutes and hours in Ukrainian (xSeconds regular, xMinutes regular, xHours regular), though Nominative Plural should be used instead of Singular Genitive too
- someone used Nominative Plural form for xYears regular in singularGenitive field

**_References:_**
- [online corrector (uk)](https://onlinecorrector.com.ua/uk/%D0%BF%D0%B5%D1%80%D0%B5%D0%B2%D1%96%D1%80%D1%82%D0%B5-%D1%83%D0%B7%D0%B3%D0%BE%D0%B4%D0%B6%D0%B5%D0%BD%D0%BD%D1%8F-%D1%96%D0%B7-%D1%87%D0%B8%D1%81%D0%BB%D1%96%D0%B2%D0%BD%D0%B8%D0%BA%D0%BE%D0%BC/)
- [stack exchange answer (en)](https://ukrainian.stackexchange.com/questions/6682/case-usage-with-the-numbers-2-3-and-4)
- [name of the movie, btw (uk)](https://uk.wikipedia.org/wiki/%D0%A7%D0%BE%D1%82%D0%B8%D1%80%D0%B8_%D0%BC%D1%96%D1%81%D1%8F%D1%86%D1%96,_%D1%82%D1%80%D0%B8_%D1%82%D0%B8%D0%B6%D0%BD%D1%96_%D1%96_%D0%B4%D0%B2%D0%B0_%D0%B4%D0%BD%D1%96)
- same settings at [moment.js](https://github.com/moment/moment/blob/2.29.4/src/locale/uk.js#L21-L23) and [day.js](https://github.com/iamkun/dayjs/blob/v1.11.5/src/locale/uk.js#L18-L20) with correct form

**_Additional info:_**
- Library version with an issue: 2.29.2
[...]

PR #3175, opened 2022-09-12
Fix grammar in formatDistance of Ukrainian locale

leshakoss:


leshakoss:
Trying to address #3170 (I don't speak the language)  @alinkedd @date-fns/i18n-uk could you please verify grammar?

alexgul1:
> @leshakoss looks good for me 👍

the same

leshakoss:
@korzhyk @alexgul1 thanks a lot!

alinkedd:
@leshakoss thank you so much for the PR, I would not hope someone will release a fix so fast.

And sorry that I'm a bit late for this suggestion - I found it a bit confusing to have plural nominative form in singular genitive field for x distances though it fixes everything 😅. imho, rewriting declension function (which is almost identical to ru and be locales and caused the problem) is more approriate way here, so we don't change library approach. I know that that would make function more complicated, but could not leave it without a few words in comment at least.

Thanks again for your work!

PR #3199, opened 2022-10-03
Fix parseISO for sub-ms decimal seconds (#3194)

### LINKED ISSUE #3194: parseISO unexcpeted behaviour for ISO-String with nanoseconds
Hi,

i was surprised when my UI was showing a different date than what i expected based on the date stored in my database.
In my opinion parseISO does not handle ISO-strings correctly if the minute fractions are nanos.

I expect parseISO to parse
`2022-10-24T23:59:59.999999999+02:00`
and return it as
`Mon Oct 24 2022 23:59:59 GMT+0200(Central European Summer Time)`
but it actually returns
`Tue Oct 25 2022 00:00:00 GMT+0200(Central European Summer Time)`

Given the following sequence this looks like a rounding problem.

```
input to parseISO -> date -> date.time
2022-10-24T23:59:59.999999999+02:00 -> Tue Oct 25 2022 00:00:00 GMT+0200(Central European Summer Time) -> 1666648800000
2022-10-24T23:59:59.99999999+02:00 -> Tue Oct 25 2022 00:00:00 GMT+0200(Central European Summer Time) -> 1666648800000
2022-10-24T23:59:59.9999999+02:00 -> Tue Oct 25 2022 00:00:00 GMT+0200(Central European Summer Time) -> 1666648800000
2022-10-24T23:59:59.999999+02:00 -> Mon Oct 24 2022 23:59:59 GMT+0200(Central European Summer Time) -> 1666648799999
2022-10-24T23:59:59.99999+02:00 -> Mon Oct 24 2022 23:59:59 GMT+0200(Central European Summer Time) -> 1666648799999
2022-10-24T23:59:59.9999+02:00 -> Mon Oct 24 2022 23:59:59 GMT+0200(Central European Summer Time) -> 1666648799999
2022-10-24T23:59:59.999+02:00 -> Mon Oct 24 2022 23:59:59 GMT+0200(Central European Summer Time) -> 1666648799999
2022-10-24T23:59:59.99+02:00 -> Mon Oct 24 2022 23:59:59 GMT+0200(Central European Summer Time) -> 1666648799990
2022-10-24T23:59:59.9+02:00 -> Mon Oct 24 2022 23:59:59 GMT+0200(Central European Summer Time) -> 1666648799900
2022-10-24T23:59:59+02:00 -> Mon Oct 24 2022 23:59:59 GMT+0200(Central European Summer Time) -> 1666648799000
```

I already looked at the parseISO-code where we calculate the milliseconds for the date-part and the time-part of the date separately. In this case the time-part is the interesting one.
https://github.com/date-fns/date-fns/blob/b8b7e521247f65d464712f5499de14e97bbef3bf/src/parseISO/index.ts#L61
The captured hours, minutes, seconds are turned into milliseconds
https://github.com/date-fns/date-fns/blob/b8b7e521247f65d464712f5499de14e97bbef3bf/src/parseISO/index.ts#L213-L228
and at the very end of parseISO we simply call the Date constructor and pass the millis.
https://github.com/date-fns/date-fns/blob/b8b7e521247f65d464712f5499de14e97bbef3bf/src/parseISO/index.ts#L94

Since we multiply the second capture in parseTime by 1000, we are still carrying the nanos with the final number we use to construct the date.
```
(as it happens in parseTime / seconds * 1000)
59.9999999 * 1000 = 59999.9999
59.999999 * 1000 = 59999.999
59.99999 * 1000 = 59999.99
59.9999 * 1000 = 59999.9
59.999 * 1000 = 59999
59.99 * 1000 = 59990
```

The solution would be to simply ommit the decimal places, which results in a "clean" non-fractional millis-number.

I'm not sure whether i just expect the wrong thing here or if there is a general consensus about such cases.
One could argue that `2022-10-24T23:59:59.999999999+02:00` is only a single nanosecond away from `2022-10-25T00:00:00+02:00` thus the rounding that happens somewhere in Date seems more appropriate timewise.

fredericbonnet:
This fixes #3194 by simply rounding down the fractional seconds to the nearest millisecond. Consider the two cases below:

| ISO string                    | `timestamp` = date part | `time` = milliseconds part | `timestamp`+`time` |
| ----------------------------- | ----------------------: | -------------------------- | ------------------ |
| `2014-02-11T11:30:30.999999`  |           1392076800000 | 41430999.999               | 1392118230999.999  |
| `2014-02-11T11:30:30.9999999` |           1392076800000 | 41430999.9999              | 1392118231000      |

In the second case the sum of `timestamp` and `time` is incorrectly rounded up because of the limited precision of JavaScript numbers. Rounding down the `time` part explicitly ensures consistent handling of such cases.

### DIFF
--- a/src/parseISO/index.ts
+++ b/src/parseISO/index.ts
@@ -58,7 +58,8 @@ export default function parseISO(
   let offset
 
   if (dateStrings.time) {
-    time = parseTime(dateStrings.time)
+    // Round down time to nearest ms to handle sub-ms fractions properly
+    time = Math.floor(parseTime(dateStrings.time))
     if (isNaN(time)) {
       return new Date(NaN)
     }
--- a/src/parseISO/test.ts
+++ b/src/parseISO/test.ts
@@ -180,6 +180,14 @@ describe('parseISO', () => {
         )
       })
 
+      it('rounds float seconds down to nearest millisecond', () => {
+        const result = parseISO('2014-02-11T11:30:30.9999999')
+        assert.deepStrictEqual(
+          result,
+          new Date(2014, 1 /* Feb */, 11, 11, 30, 30, 999)
+        )
+      })
+
       it('parses , as decimal mark', () => {
         const result = parseISO('2014-02-11T11,5')
         assert.deepStrictEqual(result, new Date(2014, 1 /* Feb */, 11, 11, 30))


fredericbonnet:
PS: Happy Hacktoberfest! 🎃

PR #3202, opened 2022-10-04
minor update in param name of getDay()

Ayushi96:
Happy Hacktoberfest 2022 <3

### DIFF
--- a/src/getDay/index.ts
+++ b/src/getDay/index.ts
@@ -8,7 +8,7 @@ import toDate from '../toDate/index'
  * @description
  * Get the day of the week of the given date.
  *
- * @param date - the given date
+ * @param dirtyDate - the given date
  * @returns the day of week, 0 represents Sunday
  *
  * @example


Ayushi96:
@leshakoss could you please review this PR

src/getDay/index.ts
```diff
@@ -8,7 +8,7 @@ import toDate from '../toDate/index'
  * @description
  * Get the day of the week of the given date.
  *
- * @param date - the given date
+ * @param dirtyDate - the given date
```
naveenkollu-code, line 11: @Ayushi96 the function is taking a date as param and getting the day of the week. There is no need of changing the `date` param to `dirtyDate` param. It is an unnecessary code enchantment. I think you can close this PR.

naveenkollu-code (changes requested):

src/getDay/index.ts
Ayushi96: Hey @naveenkollu-code, in that case would it make more sense to update the param name, cause I see an inconsistency in the doc and the code

Issue #3206, opened 2022-10-07
JSDoc annotation mismatch similar to #3202


romanbalayan:
Happy hacktoberfest 2022!

Saw other functions with similar JSDoc mismatch where JSDoc specifies `date` but param name is `dirtyDate`

```
 * @param date - the given date
 * @param dirtyDate - the given date
 ```

Can I take on updating these on the other instances? Also would you prefer 1 PR per function, or just 1 aggregated PR for all the changes?

Thanks!

PR #3207, opened 2022-10-07
Updated functions to match parameters with JSDoc #3206

romanbalayan:
Saw #3202, and found that there are lots of mismatch on JSDoc vs actual function param names. 

Collated all into 1 PR, and modified param name to match with previously published JSDoc.

Also doing this for Hacktoberfest 2022 :) 

This will fix issue #3206


romanbalayan:
Hi @tan75,

Saw that you tagged some of the PRs for hacktoberfest 2022, can this be included as well?

Thanks!

PR #3235, opened 2022-10-19
fix: Correct Lithuanian translation of seconds in some instances.

### LINKED ISSUE #3204: formatDuration Lithuanian translation for seconds is wrong
formatDuration helper has wrong translation for seconds:
for seconds 1,21,31,41,51 the translation is good (sekundė)
for seconds 10-20,30,40,50 the translation is good (sekundžių)

the rest are wrong. Returns "sekundė", when should return "sekundės"

nzdjb:
Lithuanian uses a different form of 'seconds' for numbers ending in 1. This change handles that form.

Fixes #3204

### DIFF
--- a/src/locale/lt/_lib/formatDistance/index.ts
+++ b/src/locale/lt/_lib/formatDistance/index.ts
@@ -21,7 +21,7 @@ type FormatDistanceTokenValue =
     }
 
 const translations = {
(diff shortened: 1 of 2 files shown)


revertiz (approved):
snapshot is now correct, thank you for fixing this!

Issue #3195, opened 2022-09-30
round to nearest half hour


byelik:
Hi, thanks for such great library.
Is it possible to add this function to round to nearest half hour? 

private roundToNearestHalfHour(time): Date {
		const interval = 30 * 60 * 1000;
		return new Date(Math.ceil(time.getTime() / interval) * interval)
}

thgaskell:
There's [`roundToNearestMinutes`](https://date-fns.org/v2.29.3/docs/roundToNearestMinutes) which should be able to do what you're looking for:

```js
roundToNearestMinutes(new Date(), { nearestTo: 30, roundingMethod: 'ceil' })
```

byelik:
Unfortunately it doesn’t work as expected

пт, 7 жовт. 2022 р. о 19:16 Tony Gaskell ***@***.***> пише:

polakowski:
@byelik Can you explain how do you expect it to work, and why it doesn't meet your expectations? It will be easier to help

byelik:
> @byelik Can you explain how do you expect it to work, and why it doesn't meet your expectations? It will be easier to help

Hi @polakowski, sorry that didn't attach any sample.
Here I attached a link to the StackBlitz editor.
I hope it will be clear and you will understand my thoughts about it

[https://stackblitz.com/edit/angular-date-fns-x7zglw?file=src/app/app.component.ts](url)

MelvinVermeer:
I have encountered issues with the rounding method as well. I think the rounding method tries to do too many things. And I think the default rounding of 'trunc' is a bit odd.

In this #3267 if have suggested my fix, without changing the default rounding method.

byelik:
> I have encountered issues with the rounding method as well. I think the rounding method tries to do too many things. And I think the default rounding of 'trunc' is a bit odd.
> 
> In this #3267 if have suggested my fix, without changing the default rounding method. I used comments on my own changes to justify them.


k, so what do you suggest? close this issue with the link to your comment.

MelvinVermeer:
I'd suggest that the package maintainer will have a look at this issue , described in the pull request. And considers merging either now or in a next major update. Since the change is breaking

byelik:
> I'd suggest that the package maintainer will have a look at this issue , described in the pull request. And considers merging either now or in a next major update. Since the change is breaking

k, thanks for your time

PR #3132, opened 2022-07-26
`roundToNearestMinutes` fixes and improvements

fturmel:
Closes #3129

The recent addition of the `roundingMethod` option (#3091) was not implemented correctly, and some unit tests were not checking for the right behavior. This PR fixes the problems, adds several unit tests, improves the JS docs and adds two additional examples.

I've also internally renamed the `getRoundingMethod` file and made the function as a default export for consistency to match other functions in the library.

PR #3267, opened 2022-11-02
Fix nearest minute rounding

MelvinVermeer:
Fixes: #3195 & #3268

The issue that I was facing was that when rewrote this test:

```ts
it('rounds according to the passed mode - floor - when nearestTo is provided', () => {
  const result = roundToNearestMinutes(
    new Date(2014, 6 /* Jul */, 10, 12, 10, 30, 5),
    { nearestTo: 10, roundingMethod: 'floor' }
  )
  assert.deepStrictEqual(result, new Date(2014, 6 /* Jul */, 10, 12, 10, 0))
})
```

to this

```diff
it('rounds according to the passed mode - floor - when nearestTo is provided', () => {
  const result = roundToNearestMinutes(
-    new Date(2014, 6 /* Jul */, 10, 12, 10, 30, 5),
+    new Date(2014, 6 /* Jul */, 10, 12, 19, 30, 5),

    { nearestTo: 10, roundingMethod: 'floor' }
  )
  assert.deepStrictEqual(result, new Date(2014, 6 /* Jul */, 10, 12, 10, 0))
})
```

My expectation was that `12:19:30:500` would be rounded down (floor) to the nearest interval on 10 minutes, i.e. `12:10:00:000`

This updated test failed.

I have updated the tests and implementation accordingly.


src/roundToNearestMinutes/test.ts
```diff
@@ -70,7 +71,7 @@ describe('roundToNearestMinutes', () => {
       new Date(2014, 6 /* Jul */, 10, 12, 10, 30, 5),
       { roundingMethod: 'ceil' }
     )
-    assert.deepStrictEqual(result, new Date(2014, 6 /* Jul */, 10, 12, 12, 0))
+    assert.deepStrictEqual(result, new Date(2014, 6 /* Jul */, 10, 12, 11, 0))
```
MelvinVermeer, line 74: The original expectation rounds the time up from `12:10:30:5`, to `12:12:00:000` - a false assertion i guess

src/roundToNearestMinutes/test.ts
```diff
@@ -54,7 +55,7 @@ describe('roundToNearestMinutes', () => {
       new Date(2014, 6 /* Jul */, 10, 12, 10, 30, 5),
       { roundingMethod: 'floor' }
     )
-    assert.deepStrictEqual(result, new Date(2014, 6 /* Jul */, 10, 12, 11, 0))
+    assert.deepStrictEqual(result, new Date(2014, 6 /* Jul */, 10, 12, 10, 0))
```
MelvinVermeer, line 58: Incorrect assertion if i pass in a rounding method 'floor' i expect to round down from `12:10:30` to `12:10:00` and not `12:11:00`

src/roundToNearestMinutes/test.ts
```diff
@@ -21,14 +21,15 @@ describe('roundToNearestMinutes', () => {
   it('rounds to the closest x minutes if nearestTo is provided', () => {
     const result = roundToNearestMinutes(
       new Date(2014, 6 /* Jul */, 10, 12, 10, 30),
-      { nearestTo: 4 }
+      { nearestTo: 4, roundingMethod: 'round' }
```
MelvinVermeer, line 24: the assertion in the test is based on 'normal' rounding behaviour, while `trunc` is date-fns' default.

src/roundToNearestMinutes/test.ts
```diff
@@ -21,14 +21,15 @@ describe('roundToNearestMinutes', () => {
...
     )
     assert.deepStrictEqual(result, new Date(2014, 6 /* Jul */, 10, 12, 12, 0))
   })
 
   it('rounds up >=30 seconds for nearestTo=1', () => {
     const result = roundToNearestMinutes(
-      new Date(2014, 6 /* Jul */, 10, 12, 13, 30)
+      new Date(2014, 6 /* Jul */, 10, 12, 13, 30),
+      { roundingMethod: 'round' }
```
MelvinVermeer, line 32: the assertion in the test is based on 'normal' rounding behaviour, while trunc is date-fns' default.

src/roundToNearestMinutes/test.ts
MelvinVermeer: other option is to update the assertion

src/roundToNearestMinutes/test.ts
MelvinVermeer: other option is to update the assertion

TimNuman (approved):

MelvinVermeer:
I understand that this change is considered breaking, and might belong in a major update only.

Bysmyyr:
> I understand that this change is considered breaking, and might belong in a major update only.

Could it be added as a new roundingMethod for now so it does not change anything and those who want to use it could use it. Something like, "fixedFloor"

vladshcherbin:
Such basic feature is working incorrectly when the fix's ready 🤦

mp035:
There seems to be no mention of it here, but roundToNearestMinute has  a bug which will be fixed when [pr 3467](https://github.com/date-fns/date-fns/pull/3467) is merged.  My cursory reading of this issue indicates that the bug may be clouding understanding of the underlying issue.

johnhunter:
This is a change in behaviour but I wonder how many are using it since it doesn't work as expected when the round method is "floor".

Would be nice to see this released.

src/roundToNearestMinutes/index.ts
```diff
@@ -46,14 +46,13 @@ export function roundToNearestMinutes<DateType extends Date>(
   if (nearestTo < 1 || nearestTo > 30) return constructFrom(date, NaN);
 
   const _date = toDate(date);
-  const seconds = _date.getSeconds(); // relevant if nearestTo is 1, which is the default case
-  const minutes = _date.getMinutes() + seconds / 60;
+  const seconds = date.getSeconds(); // relevant if nearestTo is 1, which is the default case
```
kossnocorp, line 49:
```suggestion
  const seconds = _date.getSeconds(); // relevant if nearestTo is 1, which is the default case
```

src/roundToNearestMinutes/index.ts
```diff
@@ -47,13 +47,12 @@ export function roundToNearestMinutes<DateType extends Date>(
 
   const _date = toDate(date);
   const seconds = _date.getSeconds(); // relevant if nearestTo is 1, which is the default case
-  const minutes = _date.getMinutes() + seconds / 60;
+  const minutes = date.getMinutes() + seconds / 60;
```
kossnocorp, line 50:
```suggestion
  const minutes = _date.getMinutes() + seconds / 60;
```

src/roundToNearestMinutes/index.ts
```diff
@@ -47,13 +47,12 @@ export function roundToNearestMinutes<DateType extends Date>(
...
   const roundingMethod = getRoundingMethod(options?.roundingMethod);
+
   const roundedMinutes = roundingMethod(minutes / nearestTo) * nearestTo;
-  const remainderMinutes = minutes % nearestTo;
-  const addedMinutes = Math.round(remainderMinutes / nearestTo) * nearestTo;
+  const result = constructFrom(date, date);
```
kossnocorp, line 54:
```suggestion
  const result = constructFrom(date, _date);
```

fturmel:
@kossnocorp please consult #3132 while working on this. It should be useful even though it was based on v2. It has fixes for bad unit tests, fractional milliseconds bug, docs improvements, etc.

kossnocorp:
@fturmel, thank you for the heads up. I'm on it! I want to compare the fix implementations, as I'm not very familiar with the function.

kossnocorp:
Thank you so much for your work. I'll make sure to mention you in the release notes. I used this PR as it covers more cases, adds more edge case tests and was also proposed a bit earlier: https://github.com/date-fns/date-fns/pull/3132

kossnocorp:
Shipped with https://github.com/date-fns/date-fns/releases/tag/v3.2.0

PR #3304, opened 2022-12-23
add custom rounding function to 'roundingMethod' parameter

micolator:


micolator:
Has anyone had the time to look at this? It would be nice to have :)

PR #3314, opened 2023-01-03
Fix serious errors in locale/is

maranomynet:
See commit messages for more info.

I added a start of a test.ts file for the icelandic locale, focusing on the parsing patterns, which were badly broken.

In addition to fixing typos in the matching and parsing patterns, I also made the matchers way (!) more forgiving/lenient in terms of the "width"  of the mathced pattern, to better adhere to the [Robustness Principle](https://en.wikipedia.org/wiki/Robustness_principle).

Also: Added proper pluralizations and declensions in formatDistance strings.


src/locale/is/_lib/formatDistance/index.ts
```diff
@@ -1,88 +1,89 @@
...
+    singular: '{{count}} sekúnda',
+    plural: '{{count}} sekúndur',
   },
 
   halfAMinute: 'hálf mínúta',
 
   lessThanXMinutes: {
-    one: 'minna en 1 mínúta',
-    other: 'minna en {{count}} mínútur',
+    singular: 'minna en {{count}} mínúta',
```
derekblank, line 25: Should we update these `{{count}}` templates to use ein/einn/eina/eitt instead? Using the numeral `1` here was originally a grammatical shortcut due to the complexity of Icelandic numbers. I think it would be better to not pass `{{count}}` to the `singular` param, which matches with how other locales behave.

src/locale/is/_lib/match/index.ts
```diff
@@ -3,89 +3,97 @@ import type { Match } from '../../../types'
...
+        /j(?:l|úl[.í]?)/.source,
+        /á(?:gú(?:\.|st)?)?/.source,
+        /s(?:ept?(?:\.|ember)?)?/.source,
+        /o(?:kt(?:\.|óber)?)?/.source,
+        /n(?:óv(?:\.|ember)?)?/.source,
+        /d(?:es(?:\.|ember)?)?/.source,
+      ].join('|') +
+      ')',
+    'i'
+  ),
```
derekblank, line 42: 🙇

derekblank (changes requested):
@maranomynet Just catching up here - as the original author of the v1 Icelandic locale, thank you for making these updates for v2 and beyond. I've just left a note about not passing `{{count}}` to the `formatDistance` singular params. IMO this would break the convention with other locales where the grammatical declension is hardcoded into the string. Otherwise, this looks good.

src/locale/is/_lib/match/index.ts
maranomynet: Thank you.

src/locale/is/_lib/formatDistance/index.ts
maranomynet: So my vote would be for not changing this any further.

maranomynet:
@derekblank 
I just noticed there's an `options?.addSuffix`  section at the bottom of `formatDistance/index.ts, that I hadn't noticed.
This PR does not make it any worse than before, but it's a little broken.

I wouldn't mind fixing it also, but that will take a bit more work, as it involves adding declensions.

I can create a separate PR for that fix, later, or if you don't mind waiting a bit, I'll add it to this one.

maranomynet:
I rabbit-holed into fixing the `addSuffix` option, and here we are.

As you see the grammar is hairy enough already, so I'd prefer not adding "one"-form exceptions all over the place.

What we have now is a grammatically coherent and correct localization for Icelandic — something I've been dying to see released for 8 months now.

src/locale/is/_lib/formatDistance/index.ts
derekblank: > Keep in mind that in Icelandic the singular applies to all numbers that end in "1": 21, 31, 41, 101, 12341, etc. (Except ones that end in "11").
> BTW, the one/other vs. singular/plural is a implementation detail local to this file, and by using a proper pluralizer at all we're already diverging from the convention quite a bit.

Good points. This is why `1` was used in the first place -- I agree it is complex enough as it is. Let's retain your changes to keep the nominative. I wonder if there are other locales that may benefit from this one/other vs. singular/plural pattern in the future.

derekblank (approved):

src/locale/is/_lib/formatDistance/index.ts
maranomynet: This page shows how (often subtly) varied pluralization grammar is: http://docs.translatehouse.org/projects/localization-guide/en/latest/l10n/pluralforms.html?id=l10n/pluralforms

So, yes the "one/other" approach only makes sense for about 50% of languages

I think it might be useful to add a link in the documentation to a couple of different examples of mature localization files where pluralization and grammar is tackled.
Also, adding more detailed "best practices" checklist for localizations would be helpful. This might include strategies for relaxed parsing/matching, and grammar.

Also, it might be a good idea to add links back to the documentation from the localization files — especially from those files that have obviously not recieved any TLC from a local.

Quick default loacalizations (like the Icelandic one was) tend to give a false sense of correctness. It's even possible that explicitly not supporting a locale is no worse than supporting it badly/incorrectly.
No support spurs action from interested locals (usually).

</rant-mode>

maranomynet:
I'm nost sure why the CI fails to detect `describe` and `it` globals for my test file only.

harshith-venkatesh-freshworks:
Hi team,
We have also observed similar issues for icelandic language.
This PR would definitely help us in resolving issues faced by our users.
https://github.com/date-fns/date-fns/issues/3933 have added a detailed points for the same.

axelbjornsson:
Can this be accepted? Some icelandic months aren't working when using v4

Issue #423, opened 2017-02-01
Explore performance bottlenecks


josemussa:
Hi guys, first of all, thanks for this amazing library. I think there's a performance issue on the `format` function, I wrote this performance comparison between momentJS and date-fns, date-fns is superior in almost all the test cases but not in the `format` test case.

Full performance test: https://github.com/josemussa/date-fns-momentjs-perf-test

##### Test
```js
import format from 'date-fns/format';
import { FORMAT } from '../constants'; //'YYYY-MM-DD'

export const dateFnsCase = () =>
    format(new Date(), FORMAT);
```
```js
import moment from 'moment';
import { FORMAT } from '../constants'; //'YYYY-MM-DD'

export const momentCase = () =>
    moment().format(FORMAT);
```
##### Result
```
date-fns x 221,146 ops/sec ±2.96% (81 runs sampled)
moment   x 300,595 ops/sec ±2.42% (79 runs sampled)
```

leshakoss:
@josemussa thank you for using `date-fns`! You actually can run benchmarks right from `date-fns` repository with `npm run benchmark`.

I ran your test on `node v7.0.0` and MBA 11″:
```
Running: Display date using the format YYYY-MM-DD.

  2 tests completed.

  moment   x 144,757 ops/sec ±1.70% (78 runs sampled)
  date-fns x 145,134 ops/sec ±1.78% (77 runs sampled)

  Fastest are: date-fns, moment
```

On my configuration the libraries work at around the same speed. Even with your result, gap is not that big. Almost no one formats 200 thousands of dates anyway.

josemussa:
Thanks, I just saw the benchmark tool included in the repo, good to know. Seems that difference in `format` performance is not that much with moment, which is perfectly ok, I just was intrigued because all the other functions are clearly superior, so that made me raise some question. 

Keep up the great work ❤️

kossnocorp:
@josemussa we collect performance stats as artifacts on our private CI for some time already and going to publish it on our website eventually.

_bigger — faster; cyan bars — Moment.js performance:_

![](http://i.ncrp.co/0c2y0G2T4411/Image%202017-02-04%20at%204.36.33%20PM.png)

Once we're done functionality-wise, we'll approach performance tuning using the collected data. 🙇 

Thank you for `date-fns-momentjs-perf-test`! If you think that you help us with the builtin benchmark, you're welcome.

I'm going to rename this issue and keep it this conversation as a reference for the future development.

[...]

PR #3325, opened 2023-01-18
Multiply performance of isValid by 100

Glandos:
- Don't try to convert to date
- use getTime() for Date objects is twice as fast than Number(Date)

Benchmark code:
```ts
import Benchmark from 'benchmark';
const isValidNew = require('./src/isValid').default
const isValidOld = require('./src/isValidOld').default

const testsFunctions = {
  "old isValid": {
    test: (date) => isValidOld(date)
  },
  "new isValid": {
    test: (date) => isValidNew(date)
  },
};

const dates = [
  new Date(''),
  new Date(),
  Date.now(),
  new Date(NaN),
]

dates.forEach(date => {
  const suite = new Benchmark.Suite;

  console.log('Testing with', date)
  Object.values(testsFunctions).forEach(test => console.log(test.test(date)))
  Object.entries(testsFunctions).forEach(([name, test]) => suite.add(name, () => test.test(date)))
  suite
    .on('cycle', (event) => console.log(String(event.target)))
    .on('complete', () => console.log(''))
    .run();
});

```
`isValidOld` is just a copy of `isValid` before my modifications.

Output:
```
❯ yarn node -r ts-node/register benchmark-date.js
yarn node v1.22.19
Testing with Invalid Date
false
false
old isValid x 6,919,462 ops/sec ±0.65% (96 runs sampled)
new isValid x 717,325,104 ops/sec ±1.54% (84 runs sampled)

Testing with 2023-01-18T10:27:40.650Z
true
true
old isValid x 6,026,135 ops/sec ±3.77% (85 runs sampled)
new isValid x 737,867,098 ops/sec ±0.54% (95 runs sampled)

Testing with 1674037660650
true
true
old isValid x 6,524,821 ops/sec ±1.61% (90 runs sampled)
new isValid x 725,461,354 ops/sec ±1.38% (88 runs sampled)

Testing with Invalid Date
false
false
old isValid x 6,603,779 ops/sec ±1.84% (93 runs sampled)
new isValid x 503,228,757 ops/sec ±29.41% (61 runs sampled)

Done in 45.06s.

```

I have a laptop, 16GB RAM, Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz

Partially address #423


src/isValid/index.ts
```diff
@@ -32,9 +31,10 @@ import toDate from '../toDate/index'
  * //=> false
  */
 export default function isValid(dirtyDate: unknown): boolean {
```
eminvergil, line 33: Changing `dirtyDate` to `date` would be more clear

Glandos: It should be better now

PR #3327, opened 2023-01-19
typo in the example if formatISODuration

wildcard:
instead of "P39Y2M20DT0H0M0S" should be "P39Y2M20DT7H5M0S"

It's a misleading example


wildcard:
Anyone can review? It's a misleading example. Just doing my part as a good boy scout ⚜️

fturmel:
Good catch!

(BTW - the "main tests" suite is failing, but it's obviously unrelated. I'll open a separate issue)

PR #3328, opened 2023-01-20
formatISODuration example fix

hexonthebeach:
The example in the docs for formatISODuration is showing the wrong output.

This change is only fixing the output-text in the example.


salmantec:
Hey @hexonthebeach, I think @wildcard already raised [PR-#3327](https://github.com/date-fns/date-fns/pull/3327) for this same problem. Please check and close your PR, if it's duplicate one.

hexonthebeach:
@salmantec you're right, we created them so close to eachother 😄

PR #3269, opened 2022-11-04
[CS] Long time formats for Czech do not include number of hours

TheKvikk:
Resolves #2921

Standardizes Czech number formats

- Added "1" to all keys containing translation for one of something – one hour, one minute etc.

PR #3376, opened 2023-03-23
Add "1" to time periods

kvetis:
Standalone period names are only usable in a sentence, but not in standalone text. Also their English locale equivalent also states the number 1 in front of the period name.

Adding 1 to the period name does not break current usage, but adds usability.

Consider following Czech sentence, which retains same meaning, possibly adding more readability: Zbývá {{distance}} – Zbývá hodina 40 minut. -> Zbývá 1 hodina 40 minut.

Consider following GUI, which currently is not usable and gains usability: ⏱️ {{distance}} – ⏱️ hodina 40 minut -> ⏱️ 1 hodina 40 minut

That is the reason to include the extra 1 in front of the period name.

### DIFF
--- a/src/locale/cs/_lib/formatDistance/index.ts
+++ b/src/locale/cs/_lib/formatDistance/index.ts
@@ -39,7 +39,7 @@ const formatDistanceLocale: FormatDistanceLocale<FormatDistanceTokenValue> = {
 
   xSeconds: {
     one: {
-      regular: 'sekunda',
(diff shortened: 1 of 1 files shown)


TheKvikk:
Already waiting for #3269

kvetis:
Closing in favour of [#3269](https://github.com/date-fns/date-fns/pull/3269).

Issue #3281, opened 2022-11-21
fr_CA abbreviated month incorrect for July


mnigh:
Should be **juill.** instead of **juil.**
https://github.com/date-fns/date-fns/blob/fadbd4eb7920bf932c25f734f3949027b2fe4887/src/locale/fr/_lib/localize/index.ts#L31

References:
- https://www.noslangues-ourlanguages.gc.ca/fr/cles-de-la-redaction/mois.html
- https://st.unicode.org/cldr-apps/v#/fr_CA/Gregorian/62c3099362a1abe7

PR #3381, opened 2023-03-25
fix: French Canadian (fr-CA) date and time formats

vinnymac:
### Summary
As you can see in the [International Components for Unicode](https://icu4c-demos.unicode.org/icu-bin/locexp?d_=en&_=fr_CA), the formats `date-fns` uses for the French Canadian locale date and time formats are incorrect.

I propose we update the French Canadian date and time styling to use the most up to date formats from the ICU, just as `Intl` does today.

I've also regenerated the snapshot.md for French Canadian, let me know if there is anything else that is needed to make this change.

Thanks

### Reproduction
[Here is a small reproduction](https://codesandbox.io/s/date-fns-fr-ca-demo-t1rhpj?file=/src/App.js) that demonstrates the problem visually comparing `fr-CA` to `date-fns` and `Intl` timestyles.

### ICU Data Snapshot
Screenshot 2023-03-25 at 12 52 13 PM


vinnymac:
@mnigh This PR covers date and time formatting. A separate PR should be opened to fix the issue you describe in #3281 though 👍🏼

PR #3401, opened 2023-04-27
Use external Babel helpers

Andarist:


.babelrc.js
```diff
@@ -7,8 +7,22 @@ const presets = [
 ]
 const plugins = ['@babel/plugin-proposal-class-properties']
 
+const transformRuntimeOptions = {
+  version: require('./package.json').dependencies['@babel/runtime'].replace(
+    /^\^/,
+    ''
+  ),
+  regenerator: false,
```
nicolo-ribaudo, line 15: With recent `@babel/runtime` versions there is no special handling of regeneratorRuntime anymore, it should be safe to remove this.

Andarist: @nicolo-ribaudo I've set this up based on the docs [here](https://babeljs.io/docs/babel-plugin-transform-runtime#regenerator). What does `regenerator: true/false` means nowadays and what's the default? Are those docs up to date? 🤔

nicolo-ribaudo: Uh yeah... the docs need to be updated. Whether generator functions are transformed using regenerator or not doesn't depend on this option, but by whether or not `@babel/plugin-transform-regenerator` is enabled.

Until a couple of releases ago (about 1 year, https://github.com/babel/babel/pull/14538), the default behavior was that generator functions transformed by regenerator relied on a global `regeneratorRuntime` variable. When using `@babel/plugin-transform-runtime` with `regenerator: true` (which is the default), all those global references were replaced with an import to `@babel/runtime/regenerator`.

However now `_regeneratorRuntime` is just a normal Babel helper, with no special handling.

Andarist, line 15:
```suggestion
```

kossnocorp:
I've opened a new PR with the commits, this time targeting v2 branch: https://github.com/date-fns/date-fns/pull/3402

Thank you a lot!

PR #3421, opened 2023-05-10
Add Kurdish language

Revan99:
I added Kurdish CKB (Central Kurdish Branch)


Revan99:
@kossnocorp can you check this one out?

kossnocorp:
@Revan99 sure, I'll review it when I have time

fturmel:
@Revan99 Do you need help converting this PR to be compatible with date-fns v3? There has been a few changes, mainly export/import and new Prettier rules. LMK. Once this is done, you can run `npm run locale-snapshots` to generate a proper `snapshot.md` file for the locale, and review it to validate the parser/formatter behavior.

I don't have first hand experience with this language, but did a bit of research and I think the right identifier for this locale would be `ckb-IQ` or `ckb` (`IQ` is the default region). `ckb` is a subset of `kur` and the identifier is supposed to be `language-REGION`, so `ku-CKB` is probably not right. The JSDocs in the index file could be updated to better identify the locale language and region as well.

https://en.wikipedia.org/wiki/Kurdish_languages
https://en.wikipedia.org/wiki/Sorani
https://icu4c-demos.unicode.org/icu-bin/locexp?_=ckb_IQ

Revan99:
thank you @fturmel, I'll try to make it compatible with the new version and fix the other issues as well, and if I needed anything I'll let you know.
thanks again.

kossnocorp (approved):
Thank you so much!

PR #3442, opened 2023-06-01
Correct URL to ISO 8601 duration standard in formatISODuration description

runjak:
It seems that the URL to digi.com listed on https://date-fns.org/v2.30.0/docs/formatISODuration is no longer accurate.
From what I could gather https://www.digi.com/resources/documentation/digidocs/90001488-13/reference/r_iso_8601_duration_format.htm should be the current working URL describing the ISO 8601 duration format.


runjak:
Ah, I think this is indeed a duplicate of #3414. I'll close it then :)

PR #3467, opened 2023-07-04
Fixes for roundToNearestMinutes()

### LINKED ISSUE #3429: `roundToNearestMinutes` producing unexpected results
I'm seeing confusing results with the `roundToNearestMinutes()`  using `"date-fns": "^2.30.0"`
Given a time such as 6:28, and rounding to the nearest 10-minute mark, I'd expect it to round to 6:30, but instead, it rounds to 6:40.

### Rounding
```js
roundToNearestMinutes(new Date('2023-02-01T06:28:00.000Z'),{nearestTo:10,roundingMethod:'round'});
```
Produces: `2023-02-01T06:40:00.000Z`
Expected: `2023-02-01T06:30:00.000Z`

### Flooring
```js
roundToNearestMinutes(new Date('2023-02-01T06:28:00.000Z'),{nearestTo:10,roundingMethod:'floor'});
```
Produces: `2023-02-01T06:30:00.000Z`
Expected: `2023-02-01T06:20:00.000Z`

### Ceiling
```js
roundToNearestMinutes(new Date('2023-02-01T06:28:00.000Z'),{nearestTo:10,roundingMethod:'ceil'});
```
Produces: `2023-02-01T06:40:00.000Z`
Expected: `2023-02-01T06:30:00.000Z`

jcarstairs-scottlogic:
#3429 demonstrated that `roundToNearestMinutes()` did not produce expected behaviour. This PR makes a few changes so that `roundToNearestMinutes()` fixes that misbehaviour...

* Previously, there was an inexplicable `addedMinutes` variable which added minutes on to the result. There was no obvious reason for this, and it could only ever cause the result to go wrong: indeed, this seemed to be the chief culprit in the case of #3429. I have removed this variable
* Fixing this behaviour caused the tests to fail. This was because the tests had been engineered to expect the wrong results. I fixed this, and while I was there, removed duplicate tests and replaced test inputs with boundary cases
* Fixing the tests to expect what they ought to expect and putting in boundary cases caused further test failures. This in turn was partly because there was no account made of the milliseconds field of the input. Now, milliseconds are accounted for, which allowed some more of the tests to pass.
* Finally, the remaining issue was that the rounding method was expected to default to `round`, but in fact it was defaulting to `trunc`. I have fixed this as well.

Now, all tests pass, and I have manually checked that this should fix #3429

### DIFF
--- a/src/roundToNearestMinutes/index.ts
+++ b/src/roundToNearestMinutes/index.ts
@@ -49,7 +49,7 @@ export default function roundToNearestMinutes<DateType extends Date>(
   const milliseconds = date.getMilliseconds()
   const seconds = date.getSeconds() + milliseconds / 1000
   const minutes = date.getMinutes() + seconds / 60
-  const roundingMethod = getRoundingMethod(options?.roundingMethod)
+  const roundingMethod = getRoundingMethod(options?.roundingMethod || 'round')
   const roundedMinutes = roundingMethod(minutes / nearestTo) * nearestTo
 
   const result = constructFrom(date, date)
--- a/src/roundToNearestMinutes/index.ts
+++ b/src/roundToNearestMinutes/index.ts
@@ -51,10 +51,8 @@ export default function roundToNearestMinutes<DateType extends Date>(
   const minutes = date.getMinutes() + seconds / 60
   const roundingMethod = getRoundingMethod(options?.roundingMethod)
   const roundedMinutes = roundingMethod(minutes / nearestTo) * nearestTo
-  const remainderMinutes = minutes % nearestTo
-  const addedMinutes = Math.round(remainderMinutes / nearestTo) * nearestTo
 
   const result = constructFrom(date, date)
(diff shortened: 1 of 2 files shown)


mp035:
I just came across this bug, and was going to submit my own PR, but upon reveiw, @jcarstairs-scottlogic , your PR is perfect and goes even deeper to completely solve the issue.   Great work. **I really hope the maintainers pull it in soon :wink:** 

I note however that the date-fns documentation clearly states that the default rounding method is 'trunc'.  You may need to add the documentation change to your PR.  Although that is a breaking change, it looks like the function hasn't performed correct truncation anyway, so anything you break would have been relying on broken code. :anguished:

kossnocorp:
Thank you so much for your work. I'll make sure to mention you in the release notes. I used this PR as it adds more edge case tests and also was proposed a bit earlier: https://github.com/date-fns/date-fns/pull/3132

kossnocorp:
Shipped with https://github.com/date-fns/date-fns/releases/tag/v3.2.0

PR #3472, opened 2023-07-12
Fix single-letter form of wednesday in Spanish

mindouro:
The single-letter form of `Wednesday` in Spanish is 'X' and not 'M'. 

Quoting [Wikipedia](https://en.wikipedia.org/wiki/Date_and_time_notation_in_Spain): "The Spanish language also has an established convention for days of the week using one letter. These are: L – lunes (Monday); M – martes (Tuesday); X – miércoles (Wednesday); J – jueves (Thursday); V – viernes (Friday), S – sábado (Saturday); and D – domingo (Sunday). Each day of the week is written using its first letter except Wednesday, which is represented by the letter X in order to avoid confusion between martes (Tuesday) and miércoles (Wednesday), which both begin with an m."

### DIFF
--- a/src/locale/es/_lib/localize/index.ts
+++ b/src/locale/es/_lib/localize/index.ts
@@ -47,7 +47,7 @@ const monthValues = {
 } as const
 
 const dayValues = {
-  narrow: ['d', 'l', 'm', 'm', 'j', 'v', 's'],
+  narrow: ['d', 'l', 'm', 'x', 'j', 'v', 's'],
   short: ['do', 'lu', 'ma', 'mi', 'ju', 'vi', 'sá'],
   abbreviated: ['dom', 'lun', 'mar', 'mié', 'jue', 'vie', 'sáb'],
   wide: [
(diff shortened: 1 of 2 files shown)


harogaston:
Hi @mindouro thank you for your contribution. I see your point and I agree with your evidence. Nonetheless I have one concern. That use ('x' instead of 'm') makes sense in environments where all days of the week are listed. I am concerned that it makes less sense in isolation - i.e. `x 13 Octubre' looks a bit off.

I would like to have the opinion of other members on this @date-fns/i18n-es

harshsingh32 (commented):
lgtm

harogaston:
@mindouro You will have to update your PR, there are more files that need changes on order for this to work. Please have a look at the rest of the files i.e. https://github.com/date-fns/date-fns/blob/4e5621915afdd1b4cf99009bcc371499eba4857a/src/locale/es/_lib/match/index.ts#L67-L76

PR #3473, opened 2023-07-14
Fix weekday order in spanish

mindouro:
This fixes the order of the weekdays in spanish since it should start with lunes (monday) and end with domingo (sunday).

Please check:   https://langster.org/en/blog/seasons-days-and-months-in-spanish


harogaston:
I am sorry but langster.org does not count as official evidence. [Wikipedia](https://en.wikipedia.org/wiki/Week) reports a very divided scenario according to which I see not enough motivation for the change since we will be making it more misaligned with other countries at the same time.

PR #3536, opened 2023-10-04
I have modified the contributing.md file

Harshil233:
Under the testing heading in the contributing.md file, I have removed the jest testing, as it was of no use.


kossnocorp:
Thank you for the effort, but it's outdated. You also pulled random change into the PR, what is that?

PR #3538, opened 2023-10-06
Adding Usage Trend for date-fns and Moment.js

cwtuan:
I would like to propose a change that adds NPM usage trend links for date-fns and Moment.js to our repository. This enhancement will make it easier for users and contributors to track the popularity and adoption of our library compared to legacy Moment.js.

By providing these links, we aim to offer our community a quick and easy way to compare the popularity and trends of date-fns and Moment.js. This information can be valuable for users who are evaluating which library to choose for their projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,10 @@ yarn add date-fns
 [See date-fns.org](https://date-fns.org/) for more details, API,
 and other docs.
 
+## Usage Trend of Date & Time Manipulation Libraries
+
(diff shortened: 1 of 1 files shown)


kossnocorp:
I'm not really a fan of marketing at the expense of other creators. I respect the Moment team a lot, but also, tbh honest, now it has become irrelevant I don't see value in it at all.

PR #3552, opened 2023-10-23
Improve type matching in buildLocalizeFn

seito2:
Fixed `@ts-ignore` in buildLocalizeFn to remove it and match in Typescript


kossnocorp:
Sorry for all the force push noise. I was trying to merge the changes to see what actually was done. I'm not 100% sure that I did it right.

I can't accept the PR unless I have removed or messed up the changes. The problem is that slapping `as` is not better than `@ts-expect-error`, which at least points out that there's a problem. There's probably a solution, but it's not too important for me to invest time into fixing it properly.

PR #3553, opened 2023-10-24
chore: remove moment.js from the deps list

azizoid:
I noticed the moment.js was in the dependency list, however it is never used in the code (why would it)


kossnocorp:
It was spying! I removed it in another commit. Thank you for the effort though!

PR #3559, opened 2023-10-26
Add tests

tan75:


src/constructFrom/test.ts
```diff
@@ -0,0 +1,50 @@
...
+    class CustomDate extends Date {}
+    const referenceDate = new CustomDate(1635158400000) // October 25, 2023
+    const value = new CustomDate('2023-10-26T12:00:00')
+
+    const result = constructFrom(referenceDate, value)
+
+    assert.ok(result instanceof CustomDate)
+    assert.deepStrictEqual(result, value)
+    assert.strictEqual(result.constructor, CustomDate)
+  })
```
kossnocorp, line 49: This case is equal to the previous one, it doesn't matter for the function how you create `CustomDate`.

For the symmetry, you better add test case where both `referenceDate` and `value` are numbers.

kossnocorp (changes requested):
Looking good! I would only change the forth test case (see my comment on the diff)

kossnocorp:
To fix the tests, please rebase to main

kossnocorp:
Can you please rebase again? 🙏

kossnocorp (approved):
Let's go!

PR #3567, opened 2023-10-31
Generic name added for format function

luizcarlosom:
Some users expressed dissatisfaction with the function's name, 'format,' as it was not considered a sufficiently generic term. Subsequently, users had to create aliases to assign a more fitting name to the function, prompting a collective decision to enhance its nomenclature.


kossnocorp:
Thank you for the effort, but this is not a way to fix the problem. I will consider adding an alias, but the renaming function is not it.

PR #3598, opened 2023-12-14
Fix TypeScript 5.4 build break

RyanCavanaugh:
This type changed names for clarity a while back and the old name has finally been removed

See https://github.com/microsoft/TypeScript/pull/52996


kossnocorp:
Thank you!

kossnocorp:
Shipped the fix with `date-fns@3.0.2`

PR #3607, opened 2023-12-18
fix: "X cannot be used as a value"

### LINKED ISSUE #3588: 3.0.0-alpha.2, format cannot be used as value
I installed `date-fns` and it installed the `alpha` version by default, I haven't seen this reported before so here it is

```typescript
import { format } from 'date-fns';

const createdAt = project.createdAt ? format(project.createdAt.toDate(), 'PP') : 'N/A';
```

Which gives me the following error `'format' cannot be used as a value because it was exported using 'export type'.`
The format is exported as `export declare function format<DateType extends Date>(date: DateType | number | string, formatStr: string, options?: FormatOptions): string;`

` typescript 5.3.2` with `date-fns 3.0.0-alpha.2`

FedericoBiccheddu:
Fix #3588 as @wojtekmaj suggested.

### DIFF
--- a/scripts/build/mts.ts
+++ b/scripts/build/mts.ts
@@ -27,7 +27,7 @@ async function createMTSFiles(dir: string): Promise<void> {
       } else if (file.isFile() && file.name.endsWith(".d.ts")) {
         const newFilePath = fullPath.replace(".d.ts", ".d.mts");
         promises.push(
-          writeFile(newFilePath, `export type * from './${file.name}'`),
+          writeFile(newFilePath, `export * from './${file.name}'`),
         );
       }
     }


nstuyvesant:
Please see https://github.com/date-fns/date-fns/pull/3617. Having all the .d.mts files just export everything from the .d.ts files isn't very DRY (just for the sake of the .mts extension). Changing the exports in the package.json and not generating the .d.mts files reduces the complexity.

FedericoBiccheddu:
> Please see #3617. Having all the .d.mts files just export everything from the .d.ts files isn't very DRY (just for the sake of the .mts extension). Changing the exports in the package.json and not generating the .d.mts files reduces the complexity.
I was about to write that I'm happy to close this PR if yours does not introduce other errors and solve issue this PR is trying to solve :raised_hands:

FedericoBiccheddu:
Close in favor of #3618.

kossnocorp:
@FedericoBiccheddu thank you! Sorry, I didn't see it earlier!

Issue #3615, opened 2023-12-19
v3 Type error: 'parseISO' cannot be used as a value because it was exported using 'export type'.


nemchik:
In v2 I had
```ts
import { format, parseISO } from "date-fns";
```

In v3 I have tried the above and also
```ts
import { format } from "date-fns/format";
import { parseISO } from "date-fns/parseISO";
```

Both give me this in v3
```
Type error: 'parseISO' cannot be used as a value because it was exported using 'export type'.
Type error: 'format' cannot be used as a value because it was exported using 'export type'.
```

naphtul:
Also true for `parse`.
Would love to get the new usage.

b-maslennikov:
Issue #3588

nstuyvesant:
Try going to node_modules/date-fns/package.json and doing a global replace of "*.mts" with "*.ts" and the problem with go away.

There was never a need to create the .d.mts files for ESM as the .d.ts files can be used without issues.

kossnocorp:
Can someone please produce a minimal reproduction repo? It will help to fix the problem.

We have a set of different environments [where it's all good](https://github.com/date-fns/date-fns/actions/runs/7247215749/job/19740705115): https://github.com/date-fns/date-fns/tree/main/examples

Sorry that it slipped!

kossnocorp:
I shipped with fix with `date-fns@3.0.1`, please try it

PR #3617, opened 2023-12-19
fix(3616,3615,3588): remove generation of problematic and unnecesary .d.mts files (use .d.ts files instead)

### LINKED ISSUE #3616: 3.0.0 - Errors when running typedoc on code that uses date-fns
Working on migrating `@carbon/charts` to `date-fns@3.0.0`.  When we produce documentation using typedoc, getting dozens of errors like:
```
src/components/essentials/tooltip.ts:184:11 - error TS1362: 'format' cannot be used as a value because it was exported using 'export type'.

184    return format(value, 'MMM d, yyyy')
              ~~~~~~

  ../../node_modules/date-fns/index.d.mts:1:1
    1 export type * from "./index.d.ts";
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    'format' was exported here.
```

If I do a global replace in node_modules/date-fns/package.json of ".mts" with ".ts", the problem is resolved (locally).

Not sure the .d.mts files are really necessary especially since they just re-export everything in the .d.ts files anyway.

nstuyvesant:
Closes #3616, #3615, #3588. All .d.mts files were doing an `export type * from` the corresponding .d.ts file. The problem is that not everything is a type but you can't export non-types from a .d.ts.

The simplest and cleanest solution is not to produce .d.mts files and update the exports in the package.json so modules use the .d.ts files. While it's nice to have a .d.mts file name to indicate it's ESM, it's not required especially when it creates other problems.


kossnocorp:
It's not a solution as it breaks other environments, see: https://github.com/date-fns/date-fns/actions/runs/7267702106/job/19804657316?pr=3617

More info: https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseCJS.md

kossnocorp:
Can someone please produce a minimal reproduction repo? It will help to fix the problem.

We have a set of different environments [where it's all good](https://github.com/date-fns/date-fns/actions/runs/7247215749/job/19740705115): https://github.com/date-fns/date-fns/tree/main/examples

Sorry that it slipped!

kossnocorp:
I shipped with the fix with `date-fns@3.0.1`, please try it

PR #3618, opened 2023-12-19
Fix errors caused by d.mts files exporting only types

### LINKED ISSUE #3616: 3.0.0 - Errors when running typedoc on code that uses date-fns

kossnocorp:
Closes #3616, #3615, #3588.

~~I still don't know what environmental settings cause this, but both smoke and "Are the types wrong?" are passing, and people reported that it helped, so I assume it should work.~~

~~I shipped it as `date-fns@3.0.1-rc.1`, so if you had this problem, I'd appreciate it if you could try it and tell me if it helped.~~

~~Additionally, I'd love to have a minimal reproduction example so this doesn't happen in the future. Please add it to https://github.com/date-fns/date-fns/tree/main/examples and send a PR mentioning me.~~

Sorry that the problem slipped. Making ESM work in all possible environments is a real nightmare.


kossnocorp:
I shipped with fix with `date-fns@3.0.1`, please try it

Issue #3620, opened 2023-12-20
Cannot import in 3.0.1: A declaration file cannot be imported without 'import type'


perrin4869:
I can't find other reports here yet, but I ran into https://github.com/date-fns/date-fns/issues/3588 like everyone else, but unfortunately the fix introduced a new error in my case:

`../../node_modules/date-fns/index.d.mts(1,15): error TS2846: A declaration file cannot be imported without 'import type'. Did you mean to import an implementation file './index.js' instead?`

I am guessing this might be the reason that `import type` was there originally in `3.0.0`. Not quite sure what the fix is though, it's quite surprising...
I can try to put together a reproduction later

kossnocorp:
Hey! I expected that, but I couldn't reproduce it. I'd appreciate it if you could produce a minimal repro example.

kossnocorp:
@perrin4869 please try `date-fns@3.0.2-rc.1`. I still need repro to add to smoke tests, but I hope it finally fixes the problem.

perrin4869:
beautiful, that seems to have done the trick!!! thanks!!
I'll put together a repro soon 😃

nstuyvesant:
<img width="1108" alt="image" src="https://github.com/date-fns/date-fns/assets/5770938/2c6e0340-c170-4327-8f28-28d783a73089">

I ran into this as well. Really seems like the package.json should have exported the types from the .d.ts files as the .d.mts only re-export them.

kossnocorp:
@nstuyvesant they are needed: https://github.com/date-fns/date-fns/pull/3617#issuecomment-1863593883

kossnocorp:
@nstuyvesant I think you're right, and it's possible to point to the same `d.ts`. I tried many combinations to make "Are types wrong" happy, and this one seems to have slipped away.

Regarding reducing, it's possible as it will give another error.

kossnocorp:
Shipped the fix with `date-fns@3.0.2`

kossnocorp:
Unfortunately, I messed up attw results, and pointing to the same d.ts indeed gives the error: https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseCJS.md

So I rolled back to the original solution by copying the content and released it as `date-fns@3.0.3`

PR #3621, opened 2023-12-20
Another crack at fixing the d.mts issue

kossnocorp:
I thought I fixed the problem in https://github.com/date-fns/date-fns/pull/3618 but it created another issue: https://github.com/date-fns/date-fns/issues/3620

This is another attempt to fix it. Please try if you have this problem: `date-fns@3.0.2-rc.1`

### DIFF
--- a/scripts/build/mts.ts
+++ b/scripts/build/mts.ts
@@ -7,7 +7,7 @@
  * It's a part of the build process.
  */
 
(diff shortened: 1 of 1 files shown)


kossnocorp:
Solving this problem by pointing ESM types to CJS types is possible.

PR #3629, opened 2023-12-21
Update wikipedia location [snack]

tyrw:
The wikipedia link in the current documentation for "Week numbering" has a hash that's not valid (#Week_numbering).

This PR replaces the invalid link hash with a valid one:

Old
`https://en.wikipedia.org/wiki/Week#Week_numbering`
New
`https://en.wikipedia.org/wiki/Week#The_ISO_week_date_system`


kossnocorp:
Appreciate it!

Issue #3630, opened 2023-12-22
getDay return Day type


mikicho:
This code has a type error:
```js
eachWeekOfInterval(
  { start, end },
  { weekStartsOn: getDay(start) }, <-- Type error
)
```
> Type 'number' is not assignable to type 'Day | undefined'

imwh0im:
i did tried resolve this issue . but it was not simple.
retuned `number` type from `getDate` function in `new Date();`.
so tried Assertion `Day` type. but it is make problem an `NaN` validate.

JamieCWW:
i'd like to give this a go

PR #3633, opened 2023-12-24
fix(get-day): Type Narrowing

imwh0im:
- #3630 
- Typescript `new Date().getDay()` return type is `number`. type Assertion for this


imwh0im:
i did tried resolve #3630 . but it was not simple. 
retuned `number` type from `getDate` function in `new Date();`.
so tried Assertion `Day` type. but it is make problem an `NaN` validate.

Issue #3641, opened 2024-01-02
Using FP functions together with Lodash's flow function is no longer properly typed


Zjaaspoer:
If you would take the following code:

```
import { addDays, addHours } from 'date-fns/fp'
import flow from 'lodash/fp/flow'

const nowPlusADayAndAnHour = flow(addDays(1), addHours(1))(new Date())
console.log(nowPlusADayAndAnHour)
```

This used to work with version >= date-fns 2.30.0 (lodash 4.17.21 & @types/lodash: 4.14.202).

But with the new typings of >= 3.0.0 the TypeScript error `Expected 0 arguments, but got 1.` is thrown for the code `new Date()`

I haven't found anything breaking in the [Changelog](https://date-fns.org/v3.0.6/docs/FP-Guide) that would influence this piece of code, and I believe it still follows the [FP Guide](https://date-fns.org/v3.0.6/docs/FP-Guide).

I would like to understand if my code is wrong or if the types broke. 😅

`const nowPlusADayAndAnHour = addDays(1)(addHours(1)(new Date()))` is still properly typed, by the way

grossbart:
The same problem happens with [fp-ts](https://gcanti.github.io/fp-ts/)’s `pipe`.

```typescript
import { isEqual } from 'date-fns/fp'
import { pipe } from 'fp-ts/pipe'

const d1 = new Date()
const d2 = new Date()

// Type of r1 is boolean
const r1 = isEqual(d2)(d1)

// Type of r2 is "any"
const r2 = pipe(d1, isEqual(d2))
```

The problem seems to be that the `FPFn<Fn, Arity>` types are opaque to these pipeable functions, which in fp-ts’s case are defined like this:

```typescript
export function pipe<A>(a: A): A
export function pipe<A, B>(a: A, ab: (a: A) => B): B
export function pipe<A, B, C>(a: A, ab: (a: A) => B, bc: (b: B) => C): C
// etc.
```

For our use-case, this makes it impossible to use `date-fns` functionally with TypeScript.

Would you be open to a PR to get this fixed (if we can find a solution) or was this a conscious decision / trade-off that was already discussed?

kossnocorp:
> Would you be open to a PR to get this fixed (if we can find a solution) or was this a conscious decision / trade-off that was already discussed?

Of course, I didn't dig into the problem and don't get it rn. If anyone can produce a repro repo that I can pull and see myself, I'm sure we'll get it fixed.

[...]

PR #3647, opened 2024-01-04
Type-safe implementation of convertToFP

grossbart:
As seen in #3641, the functions provided in v3 of `date-fns/fp` can't be used with `compose`/`flow`/`pipe` constructs from TS libraries for functional programming or even with the native Array `map`.

```typescript
import { isEqual } from 'date-fns/fp'
import { pipe } from 'fp-ts/function'

const d1 = new Date()
const d2 = new Date()

// Type of r1 is boolean
const r1 = isEqual(d2)(d1)

// Type of r2 is "any"
const r2 = pipe(d1, isEqual(d2))
```

This PR is two things:
- A demonstration of the problem through tests for the `convertToFP` function.
- A starting point for a discussion around currying.

It provides an alternative implementation named `convertToFP2` that resolves the type-level issues. However, this implementation does no longer allow for more advanced partial application scenarios, because they can't be properly implemented in TypeScript (to my current knowledge – maybe this has become possible in recent TS versions, but neither [fp-ts](https://gcanti.github.io/fp-ts/) nor [effect-ts](https://github.com/Effect-TS/effect) do this, so I assume it still isn't possible).

An example of what would no longer be possible is this:

```typescript
it("allows to group arguments", () => {
  const fpFn = convertToFP(fn, 3);
  assert(fpFn(3, 2)(1) === "1 2 3");
  assert(fpFn(3)(2, 1) === "1 2 3");
});
```

And this is an example that I would like to question: this seems really unsound and unnecessary to me, is this really something that needs to be supported?

```typescript
it("ignores calls without curried arguments", () => {
  const fpFn = convertToFP(fn, 3);
  assert(fpFn()()(3, 2)()()(1) === "1 2 3");
});
```

In my opinion the type-safety gained is more important than being able to write `fpFn(3, 2)(1)` instead of the simpler (and more idiomatic/pipeline-compatible in TS functional programming) `fpFn(3)(2)(1)`. My question for resolving this issue is whether this kind of currying could be removed as a way forward with this PR or whether there are other ideas.

### DIFF
--- a/src/fp/_lib/convertToFP/index.ts
+++ b/src/fp/_lib/convertToFP/index.ts
@@ -23,3 +23,34 @@ export function convertToFP<Fn extends FPFnInput, Arity extends FPArity>(
       : (...args: unknown[]) => convertToFP(fn, arity, curriedArgs.concat(args))
   ) as FPFn<Fn, Arity>;
 }
+
+/**
+ * An experimental alternative implementation of {@link convertToFP} with
+ * more type-safety.
+ */
+export function convertToFP2<A1, B>(fn: (a1: A1) => B): (a1: A1) => B;
+export function convertToFP2<A1, A2, B>(
+  fn: (a1: A1, a2: A2) => B,
+): (a2: A2) => (a1: A1) => B;
+export function convertToFP2<A1, A2, A3, B>(
+  fn: (a1: A1, a2: A2, a3: A3) => B,
+): (a3: A3) => (a2: A2) => (a1: A1) => B;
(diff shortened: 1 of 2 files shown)


kossnocorp:
Thank you for your work! I fixed this problem in https://github.com/date-fns/date-fns/commit/25f1bd6b7b3dabd3e025dbc53c6976200d9e0ea6

It fixes the problem in types without changing its behavior. You can see that I just needed to swap the order of overloads in the interface to make it work 🙈

grossbart:
Great to see that the solution has been so simple! 👏

PR #3653, opened 2024-01-07
Add alias formatDate for format function

ckcherry23:
Part of #776 

This PR adds an alias `formatDate()` for the `format()` function based on the implementation approach suggested in the latest comment in PR #793.


kossnocorp (approved):
🎉 Looking good!

kossnocorp:
Shipped with v3.3.0: https://github.com/date-fns/date-fns/releases/tag/v3.3.0

PR #3659, opened 2024-01-09
fix French date month ordinal

antoinerousseau:
Fixes https://github.com/date-fns/date-fns/issues/1391.


fturmel:
Thoughts on [my comment in this similar PR](https://github.com/date-fns/date-fns/pull/2796#issuecomment-981778071)?

It might be less common, but I can see a valid use case (invoicing?) that would use day ordinals without months (ex: 5e du mois / 5e jour du mois). The date-fns v1 implementation was only applying this behavior if day ordinals were followed by `MMM` or `MMMM`, which seems like the safer approach but would require a revamp of the `format` function as far as I can tell.

fturmel:
Also, there used to be unit tests for this edge case in v1. It might be a good idea to bring this back because the locale snapshot tests sometimes don't surface these more subtle issues.

https://github.com/date-fns/date-fns/blob/67a0329d663f82393261f38c4c47ad65c038bd35/src/locale/fr/build_format_locale/test.js#L299-L328

fturmel:
Not my call to make, but the concern I want to raise (especially to the non-speaking core team) is the trade-off between breaking valid edge cases that are probably already out there in favor of fixing the more common use case.

I might poke around the `format` function a bit to see if I can find a somewhat sane approach to going back to this v1 behavior.

antoinerousseau:
@Dallas62 indeed, that's why @fturmel opened the even better #3662 :)

PR #3669, opened 2024-01-15
Update CONTRIBUTING.md

fturmel:
Unlike Yarn, `npm` by itself just lists the available commands. Contributors need to run `npm i` or `npm install` to properly install the dev dependencies.


kossnocorp (approved):
Great catch!

Issue #3672, opened 2024-01-18
Rename isExist → isExisting


kossnocorp:
`isExist` is not grammatically correct, so let's rename it to `isExisting`. See comment: https://github.com/date-fns/date-fns/issues/776#issuecomment-1897687931

PR #3673, opened 2024-01-19
Add alias isExisting for isExists

Edgaraszs:
#3672 

adds alias `isExisting()` for `isExists()`


ckcherry23:
Hi @Edgaraszs! I believe #3672 is about renaming the function itself rather than adding a new alias.
Since `isExists()` is grammatically wrong, we can just change the name of the function to `isExisting()`.

PR #3687, opened 2024-01-30
Add alias `parseDate` for `parse` function

domlimm:
Part of #776.

This PR adds an alias `parseDate` for the `parse` function.


domlimm:
@kossnocorp Hey Sasha! Sorry to nudge you. May I ask if there's anything else I need to do for the issue?

PR #3697, opened 2024-02-02
Pref addDays

wang1223440313:


kossnocorp:
While I can guess why the change has been made, I won't even try explaining why I can't accept it as I need to know your arguments.

Feel free to justify it, so I can get back to you more thoroughly.

PR #3699, opened 2024-02-03
fix addMonths when amount is passed as a String

jluiz20:
When using this library in a javascript project, we can pass an amount as a string and when we do that, it generates an incorrect value as it is just concatenating the string 

image

here is where it happens

image

This conversion makes sure that this works and doesn't affect number values.

image

I tried to add a test case, but as it is using TypeScript, it does not allow the string to be passed as a parameter.

PS: I am not a javascript developer and if this solution is bad, feel free to provide a better solution or just tell me what could make it better

### DIFF
--- a/src/addMonths/index.ts
+++ b/src/addMonths/index.ts
@@ -46,7 +46,7 @@ export function addMonths<DateType extends Date>(
   // month and using a date of 0 to back up one day to the end of the desired
   // month.
   const endOfDesiredMonth = constructFrom(date, _date.getTime());
-  endOfDesiredMonth.setMonth(_date.getMonth() + amount + 1, 0);
+  endOfDesiredMonth.setMonth(_date.getMonth() + +amount + 1, 0);
   const daysInMonth = endOfDesiredMonth.getDate();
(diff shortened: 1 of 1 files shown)


kossnocorp:
Hey! I understand your frustration and will probably disappoint you even more, but I can't accept this PR. I'm sorry for that.

The right solution would be to use `parseInt` or prefix with `+` on your side rather than changing date-fns. You should have control over it, and if you're not confident in your data, it's better to validate or parse it. A more systematic approach to the problem would be to use a type-system that will yell at you when you try to pass a string instead of a number, but I guess it's too much.

I'm closing the PR because introducing an undocumented behavior in one of the many functions that accept number arguments 
will make API less consistent and predictable, which will cause more bugs in the long run.

Even if I were to agree to add this behavior to every function, add tests, and document it (which is a ton of work considering that there are more than 200 functions), the new functions will eventually break it, so it's even more work as it will be yet another item on the checklist that has to be verified manually.

So, a ton of work is needed for a wrong approach to the problem: validate/parse input in the app or use TypeScript.

PR #3703, opened 2024-02-07
formatDuration: Include undefined values if formatting with zero option set to true

tommaso-sebastianelli:
If I provide format values in `FormatDurationOptions` and set `zero` to `true`, I'm expecting also undefined  `Duration` values to be considered in the output.

Example

#### Input

```
formatDuration(
  {
    seconds: 10,
  },
  {
    format: ["hours", "minutes", "seconds"],
    zero: true,
  }
```

#### Actual output:

`10 seconds`

#### Expected output:
Since I've explicitly set hours, minutes, and seconds I'm expecting to see them as 0 in the output.

`0 hours 0 minutes 10 seconds
`

Note: It seems this was working like expected in 2.x but it is not working anymore in 3.x. 
If this is the wanted behaviour instead, I suggest to update the relative [doc](https://date-fns.org/v3.3.1/docs/formatDuration) in order to avoid misleading interpretations of `zero` param.

### DIFF
--- a/src/formatDuration/index.ts
+++ b/src/formatDuration/index.ts
@@ -105,7 +105,7 @@ export function formatDuration(
       const token = `x${unit.replace(/(^.)/, (m) =>
         m.toUpperCase(),
       )}` as FormatDistanceToken;
-      const value = duration[unit];
+      const value = duration[unit] ?? (zero ? 0 : undefined);
       if (value !== undefined && (zero || duration[unit])) {
         return acc.concat(locale.formatDistance(token, value));
       }
--- a/src/formatDuration/test.ts
+++ b/src/formatDuration/test.ts
@@ -77,4 +77,38 @@ describe("formatDuration", () => {
         "9 months, 2 days",
     );
   });
+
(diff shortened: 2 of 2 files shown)


kossnocorp:
Hey! I hate to say it, but merging the PR will break the function for anyone using `zero` right now. Additionally, the `Duration` [doesn't allow `undefined` to be set as property](https://github.com/date-fns/date-fns/blob/main/src/types.ts#L61-L76) (I realize that not all use TypeScript and set `exactOptionalPropertyTypes`).

Said that, breaking the function behavior was not a good idea to begin with, but let's work with what we have.

Let's hide this behavior behind an option. I feel like `strict` sounds the best as what you expect is to `formatDuration` strictly follow the format and print all units.

The docs can also be more explicit; I would appreciate it if you could update them as well.

What do you think?

kossnocorp (changes requested):
Let's hide it behind an option, see my comment: https://github.com/date-fns/date-fns/pull/3703#issuecomment-1989688508

MAKWritingHouse (approved):

PR #3720, opened 2024-02-28
Fix incorrect formatDistance/weeks in belarusian locale 

world1dan:
Fixes incorrect `aboutXWeeks` and `xWeeks` strings in belarusian (be) and belarusian classic (be-tarask) locales. 
Correct variant: https://www.unicode.org/cldr/cldr-aux/charts/28/summary/be.html#1436 (same in be-tarask)


kossnocorp (approved):
Дзякуй вялікі!

PR #3721, opened 2024-02-28
Add babel plugin transform optional chaining for cjs build

jankojjs:
This pull request introduces a new Babel plugin, @babel/plugin-transform-optional-chaining, to our project's Babel configuration. The optional chaining feature (?.) provides a concise and robust way to access properties on potentially nullable objects, improving code readability and reducing the risk of runtime errors caused by null or undefined values.

Changes Made:

-   Added @babel/plugin-transform-optional-chaining to the list of plugins in the Babel configuration.
-   Ensured consistency across different environments (CommonJS and ECMAScript Modules) by including the plugin in both configurations.
-   Did not modify any existing code or behavior; this change is additive and does not affect the functionality of the project.

Purpose:

-   Enhance code maintainability and readability by adopting modern JavaScript syntax features.
-   Mitigate potential runtime errors related to null or undefined values.
-   Align our codebase with best practices and industry standards for JavaScript development.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -13,6 +13,7 @@ module.exports = {
       plugins: [
         ['@babel/plugin-transform-modules-commonjs', { strict: true }],
         ['babel-plugin-add-import-extension', { extension: 'js' }],
+        ['@babel/plugin-transform-optional-chaining']
       ],
     },
 
@@ -21,7 +22,10 @@ module.exports = {
         ['@babel/preset-env', { targets: { node: 'current' }, modules: false }],
       ],
 
-      plugins: [['babel-plugin-add-import-extension', { extension: 'mjs' }]],
+      plugins: [
+        ['babel-plugin-add-import-extension', { extension: 'mjs' }],
+        ['@babel/plugin-transform-optional-chaining']
+      ],
     },
   },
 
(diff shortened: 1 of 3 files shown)


bodryi:
These changes will help a lot for older browsers support. Thank you!

bodryi:
@kossnocorp Thank you for reply. 

Our team doesn't support IE. We have to deal with old Firefox, Chrome and Chromium browsers. To be exact, our app has to run in Qt WebView, which is equivalent to Chromium v65. Now we have couple of options: downgrade to date-fns v2 or start using babel to process date-fns in node_modules. Second option isn't good at all. Don't know, how much bytes will it cost, to approve this PR. It's up to you, as a member of date-fns team.

Thanks for great lib!

kossnocorp:
> Second option isn't good at all. Don't know, how much bytes will it cost, to approve this PR. It's up to you, as a member of date-fns team.

It's actually a very good option. From the little I know about what you do, I would recommend processing ALL dependencies to ensure they are compatible with your requirements. It's not unique to date-fns to use modern JS features available in all browsers. Of course, you know better, so take it with a grain of salt.

> Don't know, how much bytes will it cost, to approve this PR. 

It's actually pretty significant, considering optional chaining is probably not the only issue you will encounter.

> Thanks for great lib!

You're welcome!

PR #3724, opened 2024-03-04
feat: se locale

audunru:
Locale for the Northern Sámi language.

This locale was/is offered by moment, so having this in date-fns as well should make the transition smoother.


kossnocorp:
Shipped with [`date-fns@3.5.0`](https://github.com/date-fns/date-fns/releases/tag/v3.5.0)

PR #3729, opened 2024-03-12
Add exact digits patterns (#916)

### LINKED ISSUE #916: Correct way to validate time
Hi guys, thanks for the impressive work on this library. I'm looking to replace `moment` entirely across 4 large projects my team and I are working on. For validating the format of a time string, we do this with `moment` in some places:

```
moment('04:30:00', 'HH:mm:ss', true).isValid() // true
moment('04:30:004', 'HH:mm:ss', true).isValid() // false
moment('04:3:00', 'HH:mm:ss', true).isValid() // false
```

Using `date-fns`, I gave this a try:

```
isValid(parse(
  '04:30:00',
  'HH:mm:ss',
  new Date()
)) // true

isValid(parse(
  '04:30:005',
  'HH:mm:ss',
  new Date()
)) // false

isValid(parse(
  '04:3:00',
  'HH:mm:ss',
  new Date()
)) // true   <-- I'd expect this to be false
```

Maybe I'm not using the right method to achieve this, but I haven't found anything else in the docs and #326 is what inspired me to try `parse`. What's the correct way to accomplish what I'm looking for here? Thank you!

yesl-kim:
description
- Issue fixes #916
- Added a new pattern: `exactNDigits`
- Modified `HH`, `mm`, `ss` parser to use exact digits patterns.

comment

- I feagured out that the original flexible digit patterns (twoDigits, threeDigits, fourDigits) are pretty universally used, including `y` and `Y`, as well as the thing kind of `mm`.
- It seems that a new pattern needs to be added to make the `y`, `Y` format also check the digit strictly as they said in the issue.
- But first of all, I wanted to ask your opinion, so I modified only the pattern of `mm`, `ss`, and `hh` format that people need the most in that issue.
- Do you agree to be revised in this direction?

Plus
- In moment.js, strictly can be chosen with a third argument.
```
moment('04:3:00', 'HH:mm:ss', true) // strict
moment('04:3:00', 'HH:mm:ss') // not strict

```
- I think it's better for backward compatibility
- what do think of it?

### DIFF
--- a/src/parse/_lib/constants.ts
+++ b/src/parse/_lib/constants.ts
@@ -16,14 +16,25 @@ export const numericPatterns = {
   fourDigits: /^\d{1,4}/, // 0 to 9999
 
   exactTwoDigits: /^\d{2}/, // 00 to 99
-  exactThreeDigits: /^\d{3}/, // 000 to 999
-  exactFourDigits: /^\d{4}/, // 0000 to 9999
+  exactThreeDigits: /^\d{3}/, // 00 to 99
+  exactFourDigits: /^\d{4}/, // 00 to 99
+
+  minSingleDigits: /^(?!0\d)\d+/, // more than 0
+  minTwoDigits: /(^(?!0\d)\d{3,})|(\d{2})/, // more than 00
+  minThreeDigits: /(^(?!0\d)\d{4,})|(\d{3})/, // more than 000
+  minFourDigits: /(^(?!0\d)\d{5,})|(\d{4})/, // more than 0000
 
   anyDigitsSigned: /^-?\d+/,
   singleDigitSigned: /^-?\d/, // 0 to 9, -0 to -9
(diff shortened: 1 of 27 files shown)


kossnocorp:
I like where it's going! Strict parsing is one of the most requested features, and I want to see it addressed. The way to go is to hide it behind the `strict` option and make it strict in the next major version by default.

Feel free to complete the PR by adding the option and expanding the approach to the rest of the numeric tokens!

MAKDevelopmentHouse (approved):

yesl-kim:
@kossnocorp Thank you for comment!

changes

- Added new patterns and utils for strict mode
    - `minNDigits`: specifies only padding (eg. `yyy` -> 001, 012, 123, 1234, ...)
    - `exactNDigits`: specifies the maximum length as well as padding
- Modified it to parse strictly as options

comment

- Normally used minNDigits pattern since the validate method is checking the range of values.
- Tested all tokens I modified but wrote the part of them randomly since they're mostly the same.

PR #3733, opened 2024-03-14
Rebuild lock file

fturmel:
Had to rebuild the lock file to be able to run the unit tests again on macOS. Deleting the "node_modules" directory and re-installing the dependencies was not sufficient.

It appears to be a known bug, see https://github.com/npm/cli/issues/4828

(While we're at it, `pnpm` would be a nice improvement for the project. It's been very fast and reliable for me over the last few years.)

CleanShot 2024-03-14 at 10 47 52@2x

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -62,22 +62,22 @@
       }
     },
     "node_modules/@ampproject/remapping": {
-      "version": "2.2.1",
-      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.2.1.tgz",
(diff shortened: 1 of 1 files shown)


kossnocorp:
Thank you!

> (While we're at it, pnpm would be a nice improvement for the project. It's been very fast and reliable for me over the last few years.)

I wouldn't say I like this idea, as it's an additional entry point for the project. I want to keep it simple, as npm is good enough, and such problems happen rarely.

We also use Bun to run timezone tests, so if anything, I would utilize it if not the previous point. I'm open to convincing, though.

kossnocorp (approved):
🫡

PR #3731, opened 2024-03-14
Make date extensions work with now functions (fixes #3730)

kossnocorp:
Fixed functions that use current date internally and made them work with date extensions like `UTCDate`.

PR #3734, opened 2024-03-14
Fix `constructNow` import paths

fturmel:
Introduced by #3731. None of the internal imports go through the root index that re-exports all functions.

Although maybe this is not an issue anymore? The browser tests on GitHub actions are not timing out like they used to (see #3677) since the Vitest 1.3 upgrade.

If this can still cause problems, maybe the root index (and fp index) could be removed from the repo and only generated when publishing to the npm registry?

### DIFF
--- a/src/constructNow/index.ts
+++ b/src/constructNow/index.ts
@@ -1,4 +1,4 @@
-import { constructFrom } from "../index.js";
+import { constructFrom } from "../constructFrom/index.js";
 
 /**
  * @name constructNow
--- a/src/formatDistanceToNow/index.ts
+++ b/src/formatDistanceToNow/index.ts
@@ -1,6 +1,6 @@
+import { constructNow } from "../constructNow/index.js";
 import type { FormatDistanceOptions } from "../formatDistance/index.js";
 import { formatDistance } from "../formatDistance/index.js";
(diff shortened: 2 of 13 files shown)


kossnocorp:
> Introduced by https://github.com/date-fns/date-fns/pull/3731. None of the internal imports go through the root index that re-exports all functions.

Thank you so much. I used auto imports and missed it 🤦‍♂️ I have to find or write an ESLint plugin to prevent it.

> Although maybe this is not an issue anymore? The browser tests on GitHub actions are not timing out like they used to (see https://github.com/date-fns/date-fns/pull/3677) since the Vitest 1.3 upgrade.

I had to fix browser tests, and it seems Vitest was choking because of the number of tests. See: https://github.com/vitest-dev/vitest/issues/5382

> If this can still cause problems, maybe the root index (and fp index) could be removed from the repo and only generated when publishing to the npm registry?

That's another way to approach it. I like it even more than linting. I'll look into that!

---

PS: I tried to contact you via email. Not sure if you saw my messages. Can you please open a [GitHub Sponsors](https://github.com/sponsors) account? I love your work so much and would like to support it.

kossnocorp (approved):
🙏 thank you

PR #3739, opened 2024-03-19
Fix Hungarian without suffix distance format

1ed:
Hi,

Currently the before and after format of a date distance is the same in the Hungarian translation of the `distanceFormatter`.
Which makes no sense to me, it was changed in https://github.com/date-fns/date-fns/pull/1765 by accident, I think.

This PR fixes the issue. Thanks!


1ed:
Hi @kossnocorp, could you or someone else have a look on this, please? Does it need more work or could it be merged? Thanks!

PR #3749, opened 2024-03-29
fix: updates for Basque (eu)

eillarra:
As a reference you can take a look at moment.js, which was updated years ago by me: https://github.com/moment/moment/blob/develop/locale/eu.js

For example, "en {result}" was still in Spanish...


fturmel (approved):
I don't speak the language, but the changes look good to me!

fturmel:
@eillarra can you confirm that the updated snapshot looks good to you? d97aa761f5a67e296c1e1ab5b150d4fb9d88e40f

eillarra:
@fturmel snapshot looks good. There is only one thing that it is not perfect, but I don't think it is possible with the current implementation:

> `6 urte gutxi gorabehera barru` should be `6 urte barru gutxi gorabehera`

But uffix is added after result is calculated, and the 'about' (gutxi gorabehera) is added beforehand to the result, so I guess it needs to stay like this. Thanks!

eillarra:
Keep this on hold for a moment... I only checked `formatDistance`, but I will check the other files too so the revision is complete. There might be a couple of small extra issues that can be solved.

fturmel:
Great! Here's the unicode CLDR reference if needed: https://www.unicode.org/cldr/charts/44/summary/eu.html

If you change anything else, you can update the snapshot markdown file yourself by running `npm run locale-snapshots`.

I'm sure it's possible to reshape the formatDistance function to support the "about" form you're describing. How does this look?

```ts
if (options?.addSuffix) {
  if (options.comparison && options.comparison > 0) {
    if (token.startsWith("about")) {
      return result.replace("gutxi", "barru gutxi");
    }
    return result + " barru";
  } else {
    return "duela " + result;
  }
}
```

eillarra:
@fturmel Thank you for your patience. I finally found a moment to go through all files yesterday. I added several small changes to conform to the official Basque rules.

Please take a look to the match file, as I worked directly on the file and I hope I didn't break any regex with the added punctuation...

src/locale/eu/_lib/formatLong/index.ts
```diff
@@ -2,10 +2,10 @@ import type { FormatLong } from "../../../types.js";
 import { buildFormatLongFn } from "../../../_lib/buildFormatLongFn/index.js";
 
 const dateFormats = {
-  full: "EEEE, y'ko' MMMM'ren' d'a' y'ren'",
-  long: "y'ko' MMMM'ren' d'a'",
+  full: "EEEE, y'eko' MMMM'ren' d'a'",
+  long: "y'eko' MMMM'ren' d'a'",
   medium: "y MMM d",
-  short: "yy/MM/dd",
+  short: "y-MM-dd",
```
eillarra, line 8: CLDR shows both `y/MM/dd` and `y-MM-dd`:

- `y/MM/dd`: https://www.unicode.org/cldr/charts/44/summary/eu.html#793cd78298fbed73
- `y-MM-dd`: https://www.unicode.org/cldr/charts/44/summary/eu.html#439e81cebcf40890

and **both are valid** in Basque: https://www.euskaltzaindia.eus/dok/arauak/Araua_0037.pdf
but the dashed version is more common.

fturmel: https://www.unicode.org/cldr/charts/44/summary/eu.html#57dac0d1b36c1261 is probably the CLDR field to look at for this value (gregorian date format short `yy/M/d`)

src/locale/eu/_lib/match/index.ts
```diff
@@ -46,39 +46,53 @@ const parseMonthPatterns = {
...
-    /^api/i,
-    /^mai/i,
-    /^eka/i,
-    /^uzt/i,
-    /^abu/i,
-    /^ira/i,
-    /^urr/i,
-    /^aza/i,
-    /^abe/i,
+  abbreviated: [
```
fturmel, line 49: we still need an `any` key in this structure, see the type errors in the tests

src/locale/eu/_lib/formatLong/index.ts
eillarra: I still think `y-MM-dd` will be a better choice, but I can change it to `y/MM/dd` if you prefer

fturmel: Fair. It's hard for a non-speaker to take a stance when things deviate from the CLDR. Maybe @date-fns/I18n-eu can chime in (and @eillarra should probably be added to this team).

fturmel (approved):
LGTM, thanks! Up to @kossnocorp to approve and merge, provided there are no additional comments from the `i18n-eu` team.

src/locale/eu/_lib/formatLong/index.ts
eillarra: Sure! Just keep in mind that even the CLDR is not consistent here... some examples use `y/MM/dd` and some others `y-MM-dd`:

- `y/MM/dd`: https://www.unicode.org/cldr/charts/44/summary/eu.html#793cd78298fbed73
- `yy/M/d`: https://www.unicode.org/cldr/charts/44/summary/eu.html#57dac0d1b36c1261
- `y-MM-dd`: https://www.unicode.org/cldr/charts/44/summary/eu.html#679dff5b9b612aa0
- `y-MM-dd`: https://www.unicode.org/cldr/charts/44/summary/eu.html#720b48090ba506b2

eillarra: This is maybe anecdotal, but the Academy itself, when showing its latest publications, uses the `y-MM` format for the publication date... :)

image

eillarra:
@kossnocorp do you need something else from my side here?

franatdorlet:
(Version 3.6) The fomat of a date in eu locale is yy/MM/dd and must be yyyy/MM/dd.

Euskaltzaindia* say in the following link: "Era laburtuan ere eman daiteke —zenbaki bidez, alegia—, dagozkien lotura-bokalak zainduz betiere: 2008/03/07; 2008-III-07; 2008-03-07tik; 2008/III/05eko, etab".

https://www.euskaltzaindia.eus/index.php?option=com_ebe&view=bilaketa&Itemid=1161&task=bilaketa&id=388

(*) Euskaltzaindia is an institution whose mission is to investigate and formulate the grammatical laws of the Basque language.

eillarra:
> (Version 3.6) The fomat of a date in eu locale is yy/MM/dd and must be yyyy/MM/dd.
> 
> Euskaltzaindia* say in the following link: "Era laburtuan ere eman daiteke —zenbaki bidez, alegia—, dagozkien lotura-bokalak zainduz betiere: 2008/03/07; 2008-III-07; 2008-03-07tik; 2008/III/05eko, etab".
> 
> https://www.euskaltzaindia.eus/index.php?option=com_ebe&view=bilaketa&Itemid=1161&task=bilaketa&id=388
> 
> (*) Euskaltzaindia is an institution whose mission is to investigate and formulate the grammatical laws of the Basque language.

Is this comment regarding the existing version or the changes proposed here?

This PR proposes to change the _short format date_ to `y-MM-dd` which results in _2008-03-07_:
https://github.com/date-fns/date-fns/pull/3749/files#diff-3d0076a9e6dac0229f1115385552761d544069f3078898f54db2c70c89752490

Please take a look to the whole conversation for a discussion about `y-MM-dd` vs `y/MM/d` and why I chose to use it.

franatdorlet:
@eillarra
Yes, this comment is about the existing version and I agree with the PR.

alenhorvat:
What are the plans to merge the changes into main?
Thank you!

PR #3766, opened 2024-04-13
Spelling correction of months and days in Spanish

sonmyd:


kossnocorp:
Hey! I'm pretty sure these are never capitalized like in English. Is it a regional standard, or do you need it for your app? In the latter case, I would advise capitalizing the string yourself.

sonmyd:
@kossnocorp ok thanks, I will do

kossnocorp:
Thank you for getting back to me! I'm closing the PR

PR #3768, opened 2024-04-15
Add `parseISODuration` function

ckcherry23:
This is a continuation of PR #3151 by @matinzd.

Mainly I have:
- Fixed the docs formatting
- Added more edge cases in the tests
- Made it invalid to use durations with just P or nothing after T as per the [Wikipedia page](https://en.wikipedia.org/wiki/ISO_8601#Durations)
- However, we still accept decimal values for any unit (eg. P0.5YT1M), even though the standard accepts decimal values only for the last unit specified (eg. P0.5Y)


src/parseISODuration/index.ts
```diff
@@ -0,0 +1,64 @@
...
+ * @returns Parsed duration object
+ *
+ * @throws It throws an "Invalid format" error if the string doesn't match the ISO Duration format
+ *
+ * @example
+ * // Convert string 'P3Y6M4DT12H30M5S' to duration:
+ * const result = parseISODuration('P3Y6M4DT12H30M5S')
+ * //=> { years: 3, months: 6, weeks: 0, days: 4, hours: 12, minutes: 30, seconds: 5 }
+ */
+
```
kossnocorp, line 29:
```suggestion
```

src/parseISODuration/index.ts
```diff
@@ -0,0 +1,64 @@
...
+  const [
+    ,
+    years = 0,
+    months = 0,
+    weeks = 0,
+    days = 0,
+    hours = 0,
+    minutes = 0,
+    seconds = 0,
+  ] = match;
```
kossnocorp, line 48: I don't like the idea of assigning undefined values to `0`. I know that `formatISODuration` assigns zero values, but these two functions together lose information. I recently played with Temporal's duration, and there you can `format(parse('P1Y2M')) === 'P1Y2M'`, which I think is a very good approach. While we can't change `formatISODuration` in v3, we can move in that direction.

WDYT?

kossnocorp (changes requested):

src/parseISODuration/index.ts
ckcherry23: I agree! I've updated the implementation to preserve undefined values instead of assigning them to 0.

keabard:
Hello there !
Any ETA to integrate this into `main` ?
Even if it's a good lib, it sucks to have to add `https://www.npmjs.com/package/iso8601-duration` just for parsing an ISO duration :D

slavafomin:
A humble question, can we release it please?

Having a date-time library w/o an ISO duration's support is so weird.

jimmy213:
@slavafomin no updates for a year now, so maybe the project reached its EOL unfortunately.

Issue #3428, opened 2023-05-16
Problem with isMatch function


arijit-itobuz:
There is an issue with the [isMatch function](https://date-fns.org/v2.30.0/docs/isMatch)

  ## The problem

isMatch not producing expected output, wrong format is also getting passed.

  ### 💻 Code demonstrating the problem

import { isMatch } from 'date-fns';

console.log(isMatch('2023-04-28', 'yyyy-MM-dd'));  // true
console.log(isMatch('2023-04-2', 'yyyy-MM-dd'));    // true
console.log(isMatch('2023-4-28', 'yyyy-MM-dd'));    // true
console.log(isMatch('223-4-28', 'yyyy-MM-dd'));      // true

  ### 🙁 Actual behavior

console.log(isMatch('2023-04-28', 'yyyy-MM-dd'));  // true
console.log(isMatch('2023-04-2', 'yyyy-MM-dd'));    // true
console.log(isMatch('2023-4-28', 'yyyy-MM-dd'));    // true
console.log(isMatch('223-4-28', 'yyyy-MM-dd'));      // true

  ### 🙂 Expected behavior

console.log(isMatch('2023-04-28', 'yyyy-MM-dd'));  // true
console.log(isMatch('2023-04-2', 'yyyy-MM-dd'));    // false
console.log(isMatch('2023-4-28', 'yyyy-MM-dd'));    // false
console.log(isMatch('223-4-28', 'yyyy-MM-dd'));      // false

  ## Debug information

  - date-fns version: 2.30.0
  - Browser/Node.js version: v18.10.0
  - Your timezone: Asia/Kolkata
  - Your current time: 12:46:39

ulrikdom:
This still exists in 3.x. So isMatch('13:3', 'HH:mm') equals true. Since isMatch uses parse, this would also be parsed to '13:03'. This seems like pretty major bug to me?

PR #3779, opened 2024-04-24
Feature: isMatchStrictMode

MatheusFelizardo:
## The problem
This PR proposes a fix for the issue #3428 

## Proposed solution
The proposed solution creates a new isMatch function called isMatchStrictMode where the date is tested but is compared the length of the date string with the format string

## Design after change
image


arijit-itobuz:
@MatheusFelizardo please add Reviewers

Issue #3594, opened 2023-12-11
dateFnsAddYears gives different results depending on system timezone. 


IanRobertsonPhocas:
When system timezone has a negative UTC offset (eg EST) , 
`dateFnsAddYears(new Date('2020-02-29T00:00:00.000Z'), 1)` returns `2021-03-01T00:00:00.000Z `
When the offset is positive eg(NZT), 
`dateFnsAddYears(new Date('2020-02-29T00:00:00.000Z'), 1)` returns `2021-02-28T00:00:00.000Z `

This function should be timezone agnostic.

PR #3783, opened 2024-04-25
feature: Update addYears function to make it timezone agnostic conversion

AllanKM:
Adjust the function AddYears to be timezone Agnostic.  addYears gives different results depending on system timezone

Resolves:
https://github.com/date-fns/date-fns/pull/3783

Issue:
https://github.com/date-fns/date-fns/issues/3594

### DIFF
--- a/src/addYears/index.ts
+++ b/src/addYears/index.ts
@@ -1,5 +1,3 @@
-import { addMonths } from "../addMonths/index.js";
-
 /**
  * @name addYears
  * @category Year Helpers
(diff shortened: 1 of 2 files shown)


kossnocorp (changes requested):
Hey, thank you for the PR.

But before moving on, can you please expand on what it does? The PR links to itself, and while I can guess, I would rather see a clear explanation of the problem.

Even if it's a breaking change, there must be a solid reason to update existing tests. Furthermore, I would ask you not to use UTC in the tests because it prevents the timezone tests from catching bugs. They should pass regardless of the timezone, and if they don't, I would like to know why so I can advise you on how to approach it properly.

PR #3787, opened 2024-04-26
Update lock file with Bun 1.1.x to include optional win32 binaries

fturmel:
The lock file is using Bun 1.0.30, which is before they started shipping Windows binaries. This is causing `npm install` errors for new contributors on Windows.

I believe simply updating the `bun` and `@types/bun` deps should fix the problem.

see #3785 for reference

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2904,9 +2904,9 @@
       }
     },
     "node_modules/@oven/bun-darwin-aarch64": {
-      "version": "1.0.30",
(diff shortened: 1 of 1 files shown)


fturmel:
🤷 It wouldn't hurt, but also doesn't really matter because of the [caret range](https://github.com/npm/node-semver?tab=readme-ov-file#caret-ranges-123-025-004). The goal is to unpin 1.0.x from the lock file, which is the actual cause of the problem.

PR #3796, opened 2024-05-06
intlFormatDistance: refactor options

Renegade334:
This small change refactors the options interface in intlFormatDistance to inherit from Intl.RelativeTimeFormatOptions, similar to how the equivalent options interface works in intlFormat.

This has the advantage of keeping the RelativeTimeFormat options up-to-date with any TypeScript library changes, as well as simplifying the declarations somewhat.

Unit tests are unaffected.

### DIFF
--- a/src/intlFormatDistance/index.ts
+++ b/src/intlFormatDistance/index.ts
@@ -26,7 +26,7 @@ export interface IntlFormatDistanceOptions
   /** Force the distance unit */
   unit?: IntlFormatDistanceUnit;
   /** The locales to use (see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#locales_argument) */
-  locale: DateFns.Utils.MaybeArray<
+  locale?: DateFns.Utils.MaybeArray<
     Intl.ResolvedDateTimeFormatOptions["locale"]
   >;
 }
--- a/src/intlFormatDistance/index.ts
+++ b/src/intlFormatDistance/index.ts
@@ -16,16 +16,19 @@ import { differenceInHours } from "../differenceInHours/index.js";
 import { differenceInMinutes } from "../differenceInMinutes/index.js";
 import { differenceInSeconds } from "../differenceInSeconds/index.js";
 import { toDate } from "../toDate/index.js";
-import type { IntlFormatLocale } from "../intlFormat/index.js";
+import type { DateFns } from "../types.js";
 
 /**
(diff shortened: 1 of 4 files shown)


src/intlFormatDistance/index.ts
```diff
@@ -16,21 +16,16 @@ import { differenceInHours } from "../differenceInHours/index.js";
...
   unit?: IntlFormatDistanceUnit;
   /** The locale(s) to use (see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#locales_argument) */
-  locale?: Intl.UnicodeBCP47LocaleIdentifier | Intl.UnicodeBCP47LocaleIdentifier[];
-  /** The locale matching algorithm to use. Other value: 'lookup'. See MDN for details [Locale identification and negotiation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#locale_identification_and_negotiation) */
-  localeMatcher?: Intl.RelativeTimeFormatLocaleMatcher;
-  /** The output message format. The values are 'auto' (e.g. `yesterday`), 'always'(e.g. `1 day ago`) */
-  numeric?: Intl.RelativeTimeFormatNumeric;
-  /** The length of the result. The values are: 'long' (e.g. `1 month`), 'short' (e.g. 'in 1 mo.'), 'narrow' (e.g. 'in 1 mo.'). The narrow one could be similar to the short one for some locales. */
-  style?: Intl.RelativeTimeFormatStyle;
+  locale?: IntlFormatLocale;
```
Renegade334, line 28: Previously, `locale` also accepted an array instead of a single string. While this is acceptable to the RelativeTimeFormat API, it isn't the convention of the date-fns library, so I've removed this option.

This behaviour wasn't originally present, and was added in what seems like an unrelated commit relating to docblocks (3fb4948). There were no unit tests covering this, and it doesn't seem like this behaviour was ever deliberately endorsed, although it is technically a breaking change.

kossnocorp (approved):
Thank you! I also added some more type changes.

Renegade334:
Thanks!

As an aside, looking at the overload hell that is intlFormat, I wonder whether the canonical signature could become that of intlFormatDistance (ie. a single options object including both `locale` and format options), with the various current signatures deprecated? Would also probably solve some of the "TS hates me when I try to call intlFormat" issues that are still open.

I won't open an issue myself, but I'll leave the decision to you!

PR #3804, opened 2024-05-16
Simplify checking result for negative zero

Connormiha:
I did a little simplification for checking for `-0` in returned result.


fturmel:
I think it's OK, but let's double check that none of these functions are expected to return `NaN`.

CleanShot 2024-05-16 at 12 37 02@2x

fturmel:
Yeah... that's a no-go / breaking change unfortunately.

Connormiha:
@fturmel  Yes, I knew about this effect when I made the edit. Should a function really return `NaN` if the function parameters are incorrect after implemented Typescript?

PR #3808, opened 2024-05-18
fix: I save the initial full time and set the final result with #1757 

### LINKED ISSUE #1757: addISOWeekYears resets times to 0:00:00
`addISOWeekYears` resets times to midnight local time. I didn't find this behavior in any other date-fns APIs. 

Repro:

```js
const base = new Date(2020, 4, 8, 3);
console.log (base)
// Fri May 08 2020 03:00:00 GMT-0700 (Pacific Daylight Time)
const addOneIsoYear = addISOWeekYears(base, 1)
console.log (addOneIsoYear)
// Fri May 14 2021 00:00:00 GMT-0700 (Pacific Daylight Time)
// note that the time is lost.
```

Tiago-Silva:
Solving issue #1757 which was returning the time to zero.
Now I save the start time and then set the final result with the start time.

close: #1757

### DIFF
--- a/src/addISOWeekYears/index.ts
+++ b/src/addISOWeekYears/index.ts
@@ -23,7 +23,6 @@ import { toDate } from "../toDate";
  * // Add 5 ISO week-numbering years to 2 July 2010:
  * const result = addISOWeekYears(new Date(2010, 6, 2), 5)
  * //=> Fri Jn 26 2015 00:00:00
- * I save the initial full time and set the final result with that initial time
  */
 export function addISOWeekYears<DateType extends Date>(
   date: DateType | number | string,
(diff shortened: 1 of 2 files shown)


src/addISOWeekYears/index.ts
```diff
@@ -22,10 +23,20 @@ import { setISOWeekYear } from "../setISOWeekYear/index.js";
  * // Add 5 ISO week-numbering years to 2 July 2010:
  * const result = addISOWeekYears(new Date(2010, 6, 2), 5)
  * //=> Fri Jn 26 2015 00:00:00
+ * I save the initial full time and set the final result with that initial time
```
JhonySpark, line 26: Maybe this comment should removed.

JhonySpark (changes requested):
I still in doubt if its best solution, i'll let a request changes for comment and will wait for another review.

JhonySpark (approved):

PR #3818, opened 2024-05-27
feat: add decades to a given date

banjoadeola17:


cupofjoakim:
What's the reasoning for wanting this added? Why not just call the `addYears` function with that number directly?

src/addDecades/index.ts
```diff
@@ -0,0 +1,28 @@
+import { addYears } from "../addYears/index.js";
+
+/**
+ * @name addYears
+ * @category Year Helpers
```
kossnocorp, line 5:
```suggestion
 * @name addDecades
 * @category Decade Helpers
```

src/addDecades/index.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @param date - The date to be changed
+ * @param amount - The amount of decades to be added.
+ *
+ * @returns The new date with the decades added
+ *
+ * @example
+ * // Add 5 years to 1 September 2014:
+ * const result = addYears(new Date(2014, 8, 1), 2)
+ * //=> Sun Sep 01 2019 00:00:00
```
kossnocorp, line 21: Please update examples

kossnocorp (changes requested):
Please rebase to `main` branch, with v4 I've added options to most of the functions, changed the approach to generics and added relevant time zone tests.

kossnocorp:
@banjoadeola17 thank you for the PR. I would appreciate a short explanation what was your motivation.

@cupofjoakim that's excellent question. I'm open to new functionality, but it's crucial to set at least some boundaries. In this case, we already have decade functions, so it only makes sense to complete the collection. It kind feels sad to have add but not sub function, but I don't want to ask too much.

Issue #3805, opened 2024-05-16
FormatDistance doesn't work the same way for all locales


rezof:
### The issue: 

`formatDistance` returns words instead of numbers in certain locales & `formatDuration` has the issue as well since it uses the locale function `formatDuration`.

for Example:

sv
```js
const duration =  {years: 1, days: 25, months: 3};
formatDuration(duration, { format, zero, delimiter, locale }); // =>  "ett år, tre månader, 25 dagar"
```

en :
```js
const duration =  {years: 1, days: 25, months: 3};
formatDuration(duration, { format, zero, delimiter, locale }); // =>  "1 year, 3 months, 25 days"
```

i believe the correct behavior is the second since the majority of the locales return numbers & not words.

rezof:
Am up to create a PR to fix this issue.

cupofjoakim:
+1 on this issue. We support a multitude of languages and really enjoy having date-fns here, but right now we have to monkey patch `sv` and `nn` due to this behaviour.

Issue #3188, opened 2022-09-21
Swedish locale defaults to `onlyNumeric: false` for `formatDistance` with no way to change it when using `formatDuration`


RCNNT-Industries:
Hey!

`formatDuration` yields results like `fyra timmar` instead of `4 timmar`, where numbers are replaced with words — with no way to change that behaviour except overriding the entire locale and changing the default of the `formatDistance` method. The flag `onlyNumeric` seems to only exist for Swedish (sv) and Nynorsk (nn).

It would be nice if all option flags passed to `formatDuration` were passed on to `formatDistance`, in case other languages also have special flags like that. 

My workaround was this:
```
const locale = {
  ...sv,
  formatDistance: (token, count, options) =>
    formatDistance(token, count, { ...options, onlyNumeric: true }),
};
```

Perhaps this is the way it's supposed to be done?

Best regards,
Erik

Lilja:
I had a look at this with some colleagues to try to find out what the issue could be. It seems like the swedish and nynorsk locale accepts a `onlyNumeric` prop for their `formatDistance` method. According to typescript, `FormatDistanceFn` accepts `options`, which type is `FormatDistanceFnOptions`. It accepts two properties, `addSuffix` and `comparision`.

You can't simple send `onlyNumeric` as a prop on an object down to formatDistance inside `formatDuration` in [`src/formatduration/index.ts`](https://github.com/date-fns/date-fns/blob/fadbd4eb7920bf932c25f734f3949027b2fe4887/src/formatDuration/index.ts#L103).

The typing for that particular line is also messed up, as [it's casting the options to any](https://github.com/date-fns/date-fns/blob/fadbd4eb7920bf932c25f734f3949027b2fe4887/src/locale/sv/_lib/formatDistance/index.ts#L114) so it can check for `onlyNumeric` presence 🤦 

I see two options:

* Add `onlyNumeric?: boolean` to `FormatDistanceFnOptions`. Then all locales can understand this prop and/or make use of it. Then it makes sense to pass options from `formatDuration` to `formatDistance`.
[...]

PR #3824, opened 2024-05-31
Remove non-standard wordMapping in formatDistance for nn and sv locales

cupofjoakim:
Today norwegian nynorsk (nn) and swedish (sv) replaces numbers under 13 with words. As far as I can tell, they are the only locales that do so. The other big norwegian dialect (nb) does not do it. I tried looking at the git history and it seems the word mapping has been here a long time and just hasn't been questioned.

As a swedish man who grew up on the border to norway, I propose that we leave this behind. I don't find the `formatDistance` behaviour problematic when used in isolation, but considering its' use in `formatDuration`, it gets a bit more iffy:

```ts
// given 
formatDuration(duration, {delimiter: ', ', locale})
// now
// `15 dagar, fyra timmar, tio minuter, 15 sekunder`
// after
// `15 dagar, 4 timmar, 10 minuter, 15 sekunder`
```

Earlier the option `onlyNumeric` existed for sv and nn to bring about this behaviour, but it was undocumented and removed in v3.0.0: https://github.com/date-fns/date-fns/releases/tag/v3.0.0

Addresses #3805 and (i believe) #3188

### DIFF
--- a/src/locale/nn/_lib/formatDistance/index.ts
+++ b/src/locale/nn/_lib/formatDistance/index.ts
@@ -86,22 +86,6 @@ const formatDistanceLocale: FormatDistanceLocale<FormatDistanceTokenValue> = {
   },
 };
 
-const wordMapping = [
-  "null",
-  "ein",
-  "to",
-  "tre",
-  "fire",
-  "fem",
-  "seks",
-  "sju",
-  "åtte",
-  "ni",
-  "ti",
-  "elleve",
-  "tolv",
-];
-
 export const formatDistance: FormatDistanceFn = (token, count, options) => {
   let result;
 
@@ -111,10 +95,7 @@ export const formatDistance: FormatDistanceFn = (token, count, options) => {
   } else if (count === 1) {
     result = tokenValue.one;
   } else {
-    result = tokenValue.other.replace(
-      "{{count}}",
-      count < 13 ? wordMapping[count] : String(count),
-    );
+    result = tokenValue.other.replace("{{count}}", String(count));
   }
 
   if (options?.addSuffix) {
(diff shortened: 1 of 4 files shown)


alexandernanberg (approved):
SGTM

fturmel:
> @alexandernanberg @kossnocorp The test failure seems to be due to this line: https://github.com/date-fns/date-fns/blob/main/src/isDate/test.ts#L46
> 
> `it.skip` takes a function, not a string. Changing it to `it.skip(() => "returns true for a date passed from another iframe")` makes the test suite work again.
> 
> Should I commit that here? How do you prefer to work with unrelated changes like this?

That's a weird one... The Vitest types allow this form and it works with Node.

https://github.com/vitest-dev/vitest/blob/70a277b5abc8b5586542adec8de10edb43a3e96f/packages/runner/src/types/tasks.ts#L249-L256

This new behavior (bug?) was introduced in Bun v1.1.11 and emerged because the CI always uses the latest Bun version.

IMO you should create a new PR for your proposed solution since every new PR/commit with make the CI test fail and it should be merged with a high priority. I would tweak it to be `it.skip("returns true for a date passed from another iframe", () => {});` since the string is the test name being skipped and not a return value.

cupofjoakim:
@fturmel I created a PR to fix that issue like you suggested: #3827

cupofjoakim:
Given that the test has been fixed upstream i've closed the other PR.

@kossnocorp What do you think about this change? Is it too big of a change to consider a minor?

mheine:
Any updates on this? Stumbled upon this exact issue not too long ago, and while we've found a workaround, I'd love to just be able to use this instead. Cheers and thanks for your work in any case folks! 🎉

wihlke:
I also just stumbled on this issue. Hoping it gets merged soon!

cupofjoakim:
@kossnocorp Bumping this to get a yay or nay. Happy to close this if it's a bad fit.

cupofjoakim:
@kossnocorp I'll close this by the end of the week unless reviewed, no point in letting it hang around forever.

cupofjoakim:
Ok, closing.

wihlke:
That’s depressing

PR #3827, opened 2024-06-04
Fix issue in bun test runner

cupofjoakim:
As discovered in PR #3824, there's a bug that was introduced in bun 1.1.11 that fails the run if a function body is lacked in tests.

### DIFF
--- a/src/isDate/test.ts
+++ b/src/isDate/test.ts
@@ -43,7 +43,7 @@ describe("isDate", () => {
           document.body.appendChild(iframe);
         }));
     } else {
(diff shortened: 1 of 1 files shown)


fturmel (approved):

JhonySpark (approved):

cupofjoakim:
@kossnocorp fixed this himself in this commit: https://github.com/date-fns/date-fns/commit/2b022df406c61f419bedc5debc8a5fd4a8b8310c

Closing this as the PR is redundant.

PR #3838, opened 2024-07-03
Fix formatRFC3339

### LINKED ISSUE #3825: formatRFC3339 outputs two digit year
`formatRFC3339` outputs two digit year

date-fns version is 3.6.0

```js
import { formatRFC3339 } from 'date-fns'

console.log(formatRFC3339(new Date('0024-01-23')));
```

Outputs
```
=> "24-01-22T20:53:32-03:06"
```

Example:
https://playcode.io/1890881

mandryllo:
Fixes #3825

### DIFF
--- a/src/formatRFC3339/index.ts
+++ b/src/formatRFC3339/index.ts
@@ -53,7 +53,7 @@ export function formatRFC3339<DateType extends Date>(
 
   const day = addLeadingZeros(_date.getDate(), 2);
   const month = addLeadingZeros(_date.getMonth() + 1, 2);
-  const year = _date.getFullYear();
+  const year = addLeadingZeros(_date.getFullYear(), 4);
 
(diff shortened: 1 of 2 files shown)


kbventures:
Great job on fixing the bug!  To ensure this fix is robust and won't break in the future, could you add a test that covers this scenario? For example:

```  
it("formats correctly 2 digit years utilising Constructor with number", () => {
    const date = new Date(2024, 11 /* Dec */, 11, 1, 0, 0, 12);
    date.setFullYear(24)
    expect(formatRFC3339(date)).toBe(`0024-12-11T01:00:00${generateOffset(date)}`);
  });
```

This will help us verify that that the bug is indeed fixed and prevent regressions.

kbventures (approved):
Looks great. Works for me locally .

PR #3853, opened 2024-07-25
formatRFC822 

kbventures:
## Describe your changes
Introduction new functionality formatRFC822

"This standard specifies a syntax for text messages that  are
     sent  among  computer  users, within the framework of "electronic
     mail". 

[RFC822](https://datatracker.ietf.org/doc/html/rfc822)

## Issue ticket number and link

[#ISSUE 3797](https://github.com/date-fns/date-fns/issues/3797)

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Unit tests

### DIFF
--- a/src/formatRFC822/index.ts
+++ b/src/formatRFC822/index.ts
@@ -1,0 +1,94 @@
+import { isValid } from "../isValid/index.js";
+import { toDate } from "../toDate/index.js";
+import { addLeadingZeros } from "../_lib/addLeadingZeros/index.js";
+
+/**
+ * The {@link formatRFC822} function options.
+ */
+export interface FormatRFC822Options {
+  /** Day is optional see Section 5.1 : https://datatracker.ietf.org/doc/html/rfc822#section-5.1)
+ */
+  includeDay?: boolean;
+}
+/**
+ * @name formatRFC822
+ * @category Common Helpers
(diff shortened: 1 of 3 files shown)


thernstig:
It is pointed out in https://github.com/date-fns/date-fns/issues/3797#issuecomment-2225603297 that openssl has a _slight_ difference here. Could there be an option added to the format to follow that exact way of openssl as well, as to me I would consider it a _varation_ of it.

kbventures:
@thernstig  Wouldn't it make more sense to create a new module, with a name distinct from RFC822, to implement this "slight difference"? This would help avoid confusion or potential issues that could arise from not adhering strictly to the RFC standard.

thernstig:
@kbventures I think both paths makes sense, it would boil down to personal preference. Or maybe more previous conventions that date-fns has used. I believe date-fns often has used the option object to allow for such differences?

src/formatRFC822/index.ts
```diff
@@ -0,0 +1,94 @@
...
+  if (!isValid(_date)) {
+    throw new RangeError("Invalid time value");
+  }
+
+  const dayName = days[_date.getUTCDay()];
+  const dayOfMonth = addLeadingZeros(_date.getUTCDate(), 2);
+  const monthName = months[_date.getUTCMonth()];
+  const year = _date.getUTCFullYear();
+
+  console.log("year",year)
```
Zen-cronic, line 73: nit: i think you forgot to remove this? 😅

PR #3927, opened 2024-10-15
fix: updates for Icelandic

magnimarels:
Updated the Icelandic localization to fit the Unicode standard

https://www.unicode.org/cldr/charts/44/summary/is.html


harshith-venkatesh-freshworks:
Hi team,
We have also observed similar issues for icelandic language.
This PR would definitely help us in resolving issues faced.
https://github.com/date-fns/date-fns/issues/3933 have added a detailed points for the same.

I believe this changes would also add more value to locale changes
https://github.com/date-fns/date-fns/pull/3314/files#diff-3d741b7557ba8b44115762ea77edcbf1586d6b84c4d77ec3f999aeda1c0f6ca0

axelbjornsson:
Can this be merged @hanse ?

PR #4005, opened 2025-03-14
doc: update roundToNearestHours docs

### LINKED ISSUE #3976: Problem in roundToNearestHours documentation
There is an issue with the [roundToNearestHours documentation](https://date-fns.org/v4.1.0/docs/roundToNearestHours)

  ### The problem with roundToNearestHours documentation

  

The examples are wrong.

  ### Suggested edits

  

1. The examples are 10 June 2014, not 10 Jul 2014.
2. The third example is rounding to nearest quarter hour, not half
3. The fourth example is rounding up, not down
4. The code in the fourth example `Date(2014, 6, 10, 1, 23, 45)` should be `Date(2014, 6, 10, 12, 23, 45)`
5. The last example is rounding down, not up
6. The expected result in all but the first example does not match the function description.  If I am not mistaken, then they should be 
`12:30:00`, `12:30:00` (an alternative example maybe helpful to show distinction with the second examples), `12:30:00`, `13:00:00`

jhulme-ut:
I'm not too happy with how I've described how nearestTo works. Still feels unclear to me. I'm open to better wording of it

- Fix description of examples of nearestTo because you can't round to fractions of hours. You can only round to multiples of hours
- Fix description of ceil and floor examples as they were describing the wrong examples.
- Fix Time in description of ceil example

fixes #3976

### DIFF
--- a/src/roundToNearestHours/index.ts
+++ b/src/roundToNearestHours/index.ts
@@ -40,22 +40,22 @@ export interface RoundToNearestHoursOptions<DateType extends Date = Date>
  * //=> Thu Jul 10 2014 13:00:00
  *
  * @example
- * // Round 10 July 2014 12:34:56 to nearest half hour:
+ * // Round 10 July 2014 12:34:56 to nearest multiple of 6 hours:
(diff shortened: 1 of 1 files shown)


davidjudilla:
@kossnocorp

Issue #3710, opened 2024-02-13
Confusing documentation around `isValid()`/`toDate()`


pleunv:
The docs for [isValid](https://date-fns.org/v3.3.1/docs/isValid) currently state the following:

> Returns false if argument is Invalid Date and true otherwise. Argument is converted to Date using toDate. See [toDate](https://date-fns.org/v3.3.1/docs/toDate) Invalid Date is a Date, whose time value is NaN.

with a type of

```ts
function isValid(
  date: unknown
): boolean
```

Going to the [toDate](https://date-fns.org/v3.3.1/docs/toDate) docs we see the following:

> Convert the given argument to an instance of Date.
> If the argument is an instance of Date, the function returns its clone.
> If the argument is a number, it is treated as a timestamp.
> If the argument is none of the above, the function returns Invalid Date.
> Note: all Date arguments passed to any date-fns function is processed by toDate.

Stating that the input can be either `Date` or `number`. However, the type is as follows:

```ts
function toDate<[DateType extends Date>(
  argument: string | number | DateType
): DateType
```
Implying that `string` should work as well. Upon testing, `toDate('2024-02-13T09:42:53.057Z')` correctly parses as "Tue Feb 13 2024 10:42:53 GMT+0100" leading me to believe that:

- The docs for `toDate()` _do not_ correctly reflect that it also supports ISO strings as input, while the types _do_.
- Since `isValid()` refers to `toDate()` for its behavior, ISO strings should also be valid input for `isValid()`.

The latter, however, appears to be an incorrect assumption as `isValid('2024-02-13T09:42:53.057Z')` returns `false` rather than `true`, which is also reflected by the implementation:

https://github.com/date-fns/date-fns/blob/6f44a167e71053999c10aa9462d7f0c52fec0faa/src/isValid/index.ts#L37-L43

Based on the above I'm inclined to think the `isValid()` documentation shouldn't point to `toDate()` but instead just state that the input is expected to be either a date object or a number in order to validate. I also think a big source of confusion is the fact that the the `isValid` argument is typed as `unknown`. I get why since you _can_ throw any value at it, but at the same time it hides critical information and might even be considered an anti-pattern in TS unless you're working with numbers, since any date object will be valid by default. And especially since v3 now allows `string` input pretty much everywhere, the current behavior of `isValid` is a bit unexpected:

> All functions that accept date arguments now also accept strings.

 Happy to help out if we find common ground 🙂.

PR #4016, opened 2025-04-23
#3710 fix isValid/toDate to consistently accept string arg

jperelli:
Issue #3710 has a good description for this PR.

## Current situation:

- `toDate` is being used in all functions to parse function argument from `Date`, `number`, `string`
- `isValid` is not currently accepting `string` as a valid date
- this situation is not consistent
- documentation is not clear about this

## Proposed Solution

- Make `isValid` accept string and return `true` or `false` depending on `toDate` parsing returning valid or invalid Date. This dependency from `isValid` to `toDate` makes them consistent
- Make documentation clear about strings in these two functions
- Update tests to check on strings valid or invalid

### DIFF
--- a/src/isValid/index.ts
+++ b/src/isValid/index.ts
@@ -23,15 +23,20 @@ import { toDate } from "../toDate/index.js";
  * //=> true
  *
  * @example
- * // For the value, convertible into a date:
+ * // For the numeric value, convertible into a date:
  * const result = isValid(1393804800000)
  * //=> true
  *
+ * @example
+ * // For the string value, convertible into a date:
+ * const result = isValid('2014-02-11T11:30:30.000Z')
+ * //=> true
+ * 
  * @example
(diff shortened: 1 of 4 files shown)


src/isValid/index.ts
```diff
@@ -23,15 +23,20 @@ import { toDate } from "../toDate/index.js";
...
+ * //=> true
+ * 
+ * @example
  * // For the invalid date:
  * const result = isValid(new Date(''))
  * //=> false
  */
 export function isValid(date: unknown): boolean {
-  return !((!isDate(date) && typeof date !== "number") || isNaN(+toDate(date)));
+  return !((!isDate(date) && typeof date !== "number" && typeof date !== "string") || isNaN(+toDate(date)));
```
jperelli, line 41: This `&& typeof date !== "string"` is the relevant change in this PR

Issue #3769, opened 2024-04-16
when using experimental loader flag, issue with longFormatter being exported twice


tomalaforge:
longFormatter is being exported twice. Inside `format` file and inside the `parse` file. 

Do we really need to export it twice? 

When using the flag --experimental-loader, we have the following issue:

````
file:///xxx/node_modules/date-fns/index.mjs?iitm=true:1714
    let $longFormatters = $1713274025149ebd0a3f5.longFormatters
        ^

SyntaxError: Identifier '$longFormatters' has already been declared
[90m    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:168:18)[39m
[90m    at callTranslator (node:internal/modules/esm/loader:279:14)[39m
[90m    at ModuleLoader.moduleProvider (node:internal/modules/esm/loader:285:30)[39m

Node.js v21.6.1
````

tomalaforge:
PR https://github.com/date-fns/date-fns/pull/3770

fturmel:
Hmm I agree this is a problem because of the re-exports in the [root barrel file](https://github.com/date-fns/date-fns/blob/main/src/index.ts).

I wonder if this is contributing to the Parcel tree-shaking issues in #3744 and #3670 as well.

PR #4023, opened 2025-05-05
Generate named exports instead of wildcard exports in index.ts

bgw:
👋 I work on Turbopack, and it was reported to us that we spend a lot of time (about 50ms on a cold build) reading, parsing, and analyzing `date-fns`. It looks like this is likely because of the use of wildcard re-exports in `index.ts`/`index.js`.

Named re-exports can allow JS bundlers to skip reading unused files entirely (using `sideEffects: false`).

Wildcard re-exports require the bundler to read all files to discover which ones contain the desired functions. This doesn't change production bundle sizes (tree shaking still removes unused functions), but it can impact the build speed of projects depending on `date-fns`.

This isn't Turbopack-specific: this would effect *all* bundlers, as there's no other way to perform this analysis of these wildcard imports correctly without reading all the files at least once.

IDK how Typescript handles this, but there might be some benefits there too.

This includes a hacky (but non-breaking) mitigation for #3769/#3770.

Credit goes to @Pruxis for noticing this in Turbopack traces.


alfredbirk:
I hope the maintainers will consider this. We also noticed the same build time problem with V3/V4 of date-fns and decided to downgrade to V2 until it is fixed.

PR #4056, opened 2025-08-04
fix(mn): Correct November month name parsing

filename24:
This pull request addresses a bug in the Mongolian (mn) locale where parsing the full month name for
November would fail, resulting in an Invalid Date.

## Problem

The issue was caused by a typo in the `formattingMonthValues.wide` array within the localize file for the
`mn` locale. The entry for November was "арваннэг дүгээр сар", which was inconsistent with the
corresponding regex  `/^(11|арван нэгдүгээр)/i` in the match file that expected "арван нэгдүгээр сар".

This discrepancy caused the parser to fail when encountering the formatted string for November.

## Solution

The fix involves correcting the typo in `src/locale/mn/_lib/localize/index.ts` to align with the parsing
pattern.

  - Modified `src/locale/mn/_lib/localize/index.ts`: Changed "арваннэг дүгээр сар" to "арван нэгдүгээр 
    сар".

This change ensures that the formatted month name for November can be correctly parsed, resolving the
Invalid Date error. The corresponding snapshot.md has been updated to reflect this fix.


filename24:
@kossnocorp conflict resloved. Please check again

PR #4070, opened 2025-09-03
Update getDay return type

vsn4ik:
The problem of types:
```ts
import { getDay, type WeekOptions } from 'date-fns';

const options: WeekOptions = {
  weekStartsOn: getDay(new Date())
  // ↑ Type 'number' is not assignable to type 'Day | undefined'. (ts 2322)
};
```

- `typescript`: v5.8.3
- `date-fns`: v4.4.0

Refs:
- #1663
- https://date-fns.org/v4.4.0/docs/getDay


vsn4ik:
@kossnocorp Please review my request 🙏

Issue #4127, opened 2025-12-17
Problem with parseISO function


ahorovyi:
There is an issue with the [parseISO function](https://date-fns.org/v4.1.0/docs/parseISO)

## The problem
The function is not null-safe, which leads to a runtime error instead of returning "Invalid Date" as stated in the documentation

core.mjs:6814 ERROR TypeError: Cannot read properties of null (reading 'split')
    at splitDateString (parseISO.js:105:28)
    at parseISO (parseISO.js:49:23)

## Debug information
- date-fns version: 4.1.0

huksley:
This is still something happens on the latest release. Is there any plans to get this fixed? It makes adopting latest version difficult.

PR #4136, opened 2026-01-07
added null-safe in parseISO

Lucifer007-max:


Lucifer007-max:
#4127 

Added Null check so won't throw the error

PR #4147, opened 2026-01-18
Fix bug #4141: Format Distance does not include number of hours/mins/seconds when the hour/min/sec is 1 (Polish, Finnish and Slovak)

### LINKED ISSUE #4141: Format Distance does not include nummber of hours/mins/seconds when the hour/min/sec is 1
When switching between languages, if I have a string as **1hour 1min**, it changes to **hour min**.

Attaching a link for reference for Polish language: https://github.com/date-fns/date-fns/blob/main/src/locale/pl/_lib/formatDistance/index.ts

For Czech, it was fixed with version 3.3.0, but for many other languages its still not fixed.

Vidhi-Kotecha:
Summary:
Format Distance does not include nummber of hours/mins/seconds when the hour/min/sec is 1
When switching between languages, if I have a string as 1hour 1min, it changes to hour min (issue exists in Polish, Finnish and Slovak).

Issue
Fixes https://github.com/date-fns/date-fns/issues/4141

### DIFF
--- a/src/locale/fi/_lib/formatDistance/index.ts
+++ b/src/locale/fi/_lib/formatDistance/index.ts
@@ -36,13 +36,13 @@ function futureYears(text: string): string {
 
 const formatDistanceLocale: FormatDistanceLocale<FormatDistanceTokenValue> = {
   lessThanXSeconds: {
-    one: "alle sekunti",
+    one: "alle 1 sekunti",
     other: "alle {{count}} sekuntia",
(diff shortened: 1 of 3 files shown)


Vidhi-Kotecha:
@kossnocorp Can you please help here? I see a similar PR was accepted earlier for Czech language.
Earlier PR: https://github.com/date-fns/date-fns/pull/3269

PR #4158, opened 2026-02-13
Fix incorrect time in set() JSDoc example

### LINKED ISSUE #4129: Problem in set documentation
There is an issue with the [set documentation](https://date-fns.org/v4.1.0/docs/set#)

### The problem with set documentation

Current state of the documentation:

> Set 12 PM to 1 September 2014 01:23:45 to 1 September 2014 12:00:00:
> ```js
> const result = set(new Date(2014, 8, 1, 1, 23, 45), { hours: 12 })
> //=> Mon Sep 01 2014 12:23:45
> ```

The time is changed to 12:23:45 and the output comment says so, but the sentence above says it would change to 12:00:00 instead.

### Suggested edits

"Set 12 PM to 1 September 2014 01:23:45 to 1 September 2014 12:**23:45**:"

veeceey:
Fixes #4129

The JSDoc comment for the `set()` function had an incorrect description of the expected result. The comment stated that setting only the hours to 12 would result in "12:00:00", but the actual result (correctly shown in the code example) is "12:23:45" since the `set()` function only modifies the specified fields and leaves other time components unchanged.

## Changes
- Updated the JSDoc comment to correctly reflect that the result is "12:23:45" not "12:00:00"

## Testing
Verified the existing code example shows the correct output:
```typescript
const result = set(new Date(2014, 8, 1, 1, 23, 45), { hours: 12 })
//=> Mon Sep 01 2014 12:23:45
```

The minutes (23) and seconds (45) are preserved as expected.

### DIFF
--- a/src/set/index.ts
+++ b/src/set/index.ts
@@ -39,7 +39,7 @@ export interface SetOptions<DateType extends Date = Date>
  * //=> Tue Oct 20 2015 00:00:00
  *
  * @example
- * // Set 12 PM to 1 September 2014 01:23:45 to 1 September 2014 12:00:00:
+ * // Set 12 PM to 1 September 2014 01:23:45 to 1 September 2014 12:23:45:
  * const result = set(new Date(2014, 8, 1, 1, 23, 45), { hours: 12 })
  * //=> Mon Sep 01 2014 12:23:45
  */


veeceey:
Friendly ping - any chance someone could take a look at this when they get a chance? Happy to make any changes if needed.

PR #4179, opened 2026-04-19
fix(de-AT): support parsing "Jänner" with de-AT locale

### LINKED ISSUE #4148: Cannot parse "Jänner" with deAT locale
It is not possible to parse the string "6. Jänner" with the locale set to deAT.

Source encoding is UTF-8
Node.js version 22.20.0

Source:
```javascript
import { de, deAT } from "date-fns/locale";
import { parse as dateFnsParse } from "date-fns/parse";

var options = {};

console.log("16. Jänner");
options.locale = deAT;
var result = dateFnsParse("16. Jänner", "dd. MMMM", new Date(), options);
console.log(result.toDateString());
console.log("\n\n");

console.log("16. J\xE4nner");
options.locale = deAT;
var result = dateFnsParse("16. J\xE4nner", "dd. MMMM", new Date(), options);
console.log(result.toDateString());
console.log("\n\n");

console.log("16. Januar");
options.locale = de;
var result = dateFnsParse("16. Januar", "dd. MMMM", new Date(), options);
console.log(result.toDateString());
console.log("\n\n");
```

Produces:
```

> nodejs-test@0.1.0 run
> node index.js

16. Jänner
Invalid Date



16. Jänner
Invalid Date



16. Januar
Fri Jan 16 2026



```

mixelburg:
Fixes #4148

## Problem
The  locale cannot parse dates containing "Jänner" (Austrian German word for January). The string  results in  even though the locale's own  module correctly lists "Jänner" as January.

## Root Cause
The  locale reuses the  module from the  locale. The  locale's wide month matching regex only includes "januar" but not "jänner", so the match step fails before parsing can begin.

## Fix
Added  as an alternative before  in the wide month match pattern in . The existing  pattern  already correctly matches both "Ja" and "Jä", so the parsed index resolves to January (0) as expected.

### DIFF
--- a/src/locale/de/_lib/match/index.ts
+++ b/src/locale/de/_lib/match/index.ts
@@ -28,7 +28,7 @@ const matchMonthPatterns = {
   narrow: /^[jfmasond]/i,
   abbreviated:
     /^(j[aä]n|feb|mär[z]?|apr|mai|jun[i]?|jul[i]?|aug|sep|okt|nov|dez)\.?/i,
-  wide: /^(januar|februar|märz|april|mai|juni|juli|august|september|oktober|november|dezember)/i,
+  wide: /^(jänner|januar|februar|märz|april|mai|juni|juli|august|september|oktober|november|dezember)/i,
 };
 const parseMonthPatterns = {
   narrow: [


kossnocorp (approved):
Thank you!

PR #4180, opened 2026-04-20
fix: constructFrom producing wrong date for custom Date classes with extra constructor params

### LINKED ISSUE #4160: This line makes my custom Date uncompatible with date-fns
Hi! this line:
https://github.com/date-fns/date-fns/blob/dd66398305c2b015fba3c1b3d31ccff42ee8d4cf/src/constructFrom/index.ts#L57
breaks my custom Date class.

This is why:

```typescript
class RangeDate extends Date {
  constructor(
    public readonly range: number,
    ...args: ConstructorParameters<typeof Date>
  ) { super(...args) }
}

const x = new RangeDate(5, new Date(2026, 1, 3, 5))

console.log(x.getDate()) // works flawless.
console.log((new (x.constructor)(x)).getDate()) // always prints date of today.
```

Btw, I've also noticed that there's TODO in the code to get rid of this weird transformation.

mixelburg:
Fixes #4160

## Problem

When a custom `Date` subclass has extra constructor parameters (e.g. `RangeDate extends Date` with a `range` param), `constructFrom` calls `new (date.constructor)(value)` passing only the date value. The extra required parameters become `undefined`, causing the constructor to call `super()` with no arguments — producing the current time instead of the expected date.

```ts
class RangeDate extends Date {
  constructor(range: number, ...args: ConstructorParameters<typeof Date>) {
    super(...args);
    this.range = range;
  }
}

const x = new RangeDate(5, new Date(2026, 1, 3, 5));
const y = constructFrom(x, new Date(2024, 0, 1));
// Before fix: y is the current time (wrong)
// After fix: y is 2024-01-01 (correct)
```

## Fix

The `date.constructor === Date` fast path handles the common case without changes. For custom subclasses, the constructor is tried and the result's time value is validated against the input — if they don't match, it falls back to `new Date(value)`. This ensures correct results for:

- Plain `Date` instances (unchanged fast path)
- Custom Date subclasses that happen to have Date-compatible constructors (still works)
- Custom Date subclasses with extra required constructor params (now correctly falls back)

Custom Date types that need to preserve extra properties (like `TZDate`) already implement `constructFromSymbol`, so they're unaffected.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,15 +1,15 @@
 [submodule "submodules/tz"]
 	path = submodules/tz
-	url = git@github.com:date-fns/tz.git
+	url = https://github.com/date-fns/tz.git
 [submodule "submodules/utc"]
 	path = submodules/utc
-	url = git@github.com:date-fns/utc.git
+	url = https://github.com/date-fns/utc.git
 [submodule "submodules/docs"]
 	path = submodules/docs
-	url = git@github.com:date-fns/docs.git
+	url = https://github.com/date-fns/docs.git
 [submodule "submodules/build"]
 	path = submodules/build
-	url = git@github.com:date-fns/build.git
+	url = https://github.com/date-fns/build.git
 [submodule "submodules/mcp"]
 	path = submodules/mcp
-	url = git@github.com:date-fns/mcp.git
+	url = https://github.com/date-fns/mcp.git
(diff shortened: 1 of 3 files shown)


mixelburg:
Closing due to inactivity. Over a month with no review.

PR #4194, opened 2026-05-21
Fix Chinese locale month parsing

### LINKED ISSUE #3256: Unable to parse date with locale `zh-CN` when value is any date in October
Parse function returns invalid date instead of Date Object if I use Parse function to parse date string in zh-CN locale. The other locales work fine.

We found that the Parse function has problem only if the input is any date in October.

Demo: https://codesandbox.io/s/date-fns-zhcn-locale-ynon8v

puneetdixit200:
Fixes #3256

This updates the Chinese locale month matchers so October is accepted in both numeric and Chinese forms, and tightens numeric parsing so `10`, `11`, and `12` are not parsed as January. The same matcher structure is shared by `zh-CN`, `zh-HK`, and `zh-TW`, so all three are updated together.

Tests:
- `pnpm vitest run src/locale/zh-CN/test.ts`
- `pnpm exec tsx -e <manual zh-CN parse check for 10/11/12>`
- `pnpm exec prettier --check src/locale/zh-CN/_lib/match/index.ts src/locale/zh-HK/_lib/match/index.ts src/locale/zh-TW/_lib/match/index.ts src/locale/zh-CN/test.ts`
- `pnpm exec oxlint --type-aware src/locale/zh-CN/_lib/match/index.ts src/locale/zh-HK/_lib/match/index.ts src/locale/zh-TW/_lib/match/index.ts src/locale/zh-CN/test.ts`
- `git diff --check`


kossnocorp:
Rebased manually and shipped as [`v4.3.0`](https://github.com/date-fns/date-fns/releases/tag/v4.3.0)

PR #4204, opened 2026-05-31
fix(formatRFC3339): pad years below 1000 to four digits

### LINKED ISSUE #4004: formatRFC3339 doesn't follow the standard with year field
**RFC3339**
`
date-fullyear   = 4DIGIT
`

year variable is missing addLeadingZeros and will result less than 4 digits for years below 1000

`
  const year = date_.getFullYear();
`

viraatdas:
Fixes #4004

## Problem
`formatRFC3339` emits fewer than four year digits for years below 1000 (year `999` formats as `999-...`), which violates RFC 3339 (`date-fullyear = 4DIGIT`). Every other field is zero-padded; the year was the one exception.

## Fix
Pad the year with the existing `addLeadingZeros` helper, matching the rest of the function:
`const year = addLeadingZeros(date_.getFullYear(), 4)`. Year `999` now formats as `0999-...`.

Edited `pkgs/core/src/formatRFC3339/index.ts` (the v4 source), not built output.

## Tests
Added a regression test ("pads years before 1000 to four digits") using the existing offset/vitest pattern in the same test file.

### DIFF
--- a/pkgs/core/src/formatRFC3339/index.ts
+++ b/pkgs/core/src/formatRFC3339/index.ts
@@ -52,7 +52,7 @@ export function formatRFC3339(
 
   const day = addLeadingZeros(date_.getDate(), 2);
   const month = addLeadingZeros(date_.getMonth() + 1, 2);
-  const year = date_.getFullYear();
+  const year = addLeadingZeros(date_.getFullYear(), 4);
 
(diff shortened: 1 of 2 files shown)


viraatdas:
Closing as a duplicate of #3838, which already proposes the same one-line fix. Deferring to that PR.

PR #4207, opened 2026-06-01
Fix transpose invalid date handling

subhwastaken:
Recreated from the previously prepared PR, now using the correct GitHub identity (Subharup-31) and fork.

This applies the same transpose invalid date handling fix as the earlier PR from the wrong account.


subhwastaken:
Closing this PR because the description included process/account context that should not have been posted publicly.

PR #4241, opened 2026-07-10
Fix differenceInMonths reporting a full month one DST offset short

### LINKED ISSUE #1758: differenceInMonths reports a full month even if it's one hour short (if period starts at DST)

Otto-Deviant1904:
Fixes #1758.

`differenceInMonths` compared dates with the UTC-based `compareAsc`, which is the same bug already fixed for `differenceInDays` in #1754 (comparing UTC milliseconds can disagree with the actual local wall-clock ordering across a DST transition).

But swapping in a local-time comparator alone wasn't enough here. `differenceInMonths` also has its own month-length-overflow handling (the `setDate(30)` / `setMonth()` dance used to make e.g. "Feb 28" comparable to "Jan 30"), which walks a `Date` object `difference` months back for the final comparison. That navigation can land on a nonexistent local time during a DST "spring forward" gap (e.g. "March 8, 2:00 AM" in `America/Los_Angeles`, which doesn't exist since clocks jump straight to 3:00 AM). JS silently renormalizes that to a different hour instead of throwing, which was masking the exact shortfall this comparison exists to catch.

Fixed by comparing local calendar fields (year/month/day/time) throughout, sourcing the day-of-month from the renavigated date (safe, since a DST gap never changes the date) but the time-of-day from the original, never-renavigated date (since that's the part a gap can corrupt).

Added a DST regression test mirroring the existing `differenceInDays` one, gated the same way (skipped in timezones with no DST) and using the shared `getDstTransitions` helper so it runs against whatever DST-observing timezone the test machine is in, rather than hardcoding one.

Ran the full suite locally across several DST timezones (`America/Los_Angeles`, `America/New_York`, `Europe/London`, `Pacific/Auckland`, `Australia/Lord_Howe` for the 30-minute-offset case) plus non-DST ones (`UTC`, `Asia/Tokyo`) — all pass. No other tests affected.

(Used AI assistance to dig into this; verified the root cause and the fix against the full test suite before opening this.)

### DIFF
--- a/pkgs/core/src/differenceInMonths/index.ts
+++ b/pkgs/core/src/differenceInMonths/index.ts
@@ -1,5 +1,4 @@
 import { normalizeDates } from "../_lib/normalizeDates/index.ts";
-import { compareAsc } from "../compareAsc/index.ts";
 import { differenceInCalendarMonths } from "../differenceInCalendarMonths/index.ts";
 import { isLastDayOfMonth } from "../isLastDayOfMonth/index.ts";
 import type { ContextOptions, DateArg } from "../types.ts";
@@ -37,7 +36,7 @@ export function differenceInMonths(
     earlierDate,
   );
 
-  const sign = compareAsc(workingLaterDate, earlierDate_);
+  const sign = compareLocalAsc(workingLaterDate, earlierDate_);
   const difference = Math.abs(
     differenceInCalendarMonths(workingLaterDate, earlierDate_),
   );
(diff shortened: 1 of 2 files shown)


Otto-Deviant1904:
Friendly bump — still happy to adjust if anything's off here.

Otto-Deviant1904:
Friendly bump - still happy to adjust if anything's off. Branch is up to date with main.

Otto-Deviant1904:
Closing this for now — no review traffic and I do not plan to keep the branch current. Happy to reopen if you would like to pick it up.

PR #2890, opened 2021-12-29
fix(locale): fix zh_TW quarter translation

gaplo:

Issue #2889, opened 2021-12-29
fix(locale): fix zhTW quarter translation 


gaplo:
Fix the transaction of quarter in zh_TW, sync with zh_HK and zh_CN

PR #2771, opened 2021-11-18
fix(locale): fix zhCN quarter translation

07akioni:
In the current translation, quarter (season) is translated to every 15 minutes. I've fixed it.

PR #4282, opened 2026-07-29
Fix zh-CN quarter parsing and make the locale snapshot test actually run

### LINKED ISSUE #4270: zh-CN cannot parse its own formatted quarter: localize produces 季 forms but match expects 刻 forms
Human note: this is a new issue not currently in the backlog, it seems.

## Description

In the zh-CN locale, formatting a quarter with `QQQ`/`QQQQ` produces 季-based strings, but the locale's own parse patterns (`matchQuarterPatterns` in `src/locale/zh-CN/_lib/match/index.ts`) only accept the 第[一二三四]刻 / 第[一二三四]刻钟 forms. As a result, a format → parse round-trip through the same locale and token returns `Invalid Date`. zh-HK and zh-TW round-trip correctly.

## Reproduction

```js
import { format, parse } from "date-fns";
import { zhCN, zhHK, zhTW } from "date-fns/locale";

const date = new Date(2024, 6, 1); // Q3 2024
const ref = new Date(2024, 0, 1);

format(date, "QQQ", { locale: zhCN });           //=> "第三季"
parse("第三季", "QQQ", ref, { locale: zhCN });    //=> Invalid Date  ❌

format(date, "QQQQ", { locale: zhCN });          //=> "第三季度"
parse("第三季度", "QQQQ", ref, { locale: zhCN }); //=> Invalid Date  ❌

// zh-CN's parser only accepts the 刻 forms it never produces:
parse("第三刻", "QQQ", ref, { locale: zhCN });    //=> Mon Jul 01 2024 (unexpected)
parse("第三刻钟", "QQQQ", ref, { locale: zhCN }); //=> Mon Jul 01 2024 (unexpected)

// zh-HK and zh-TW round-trip fine:
parse(format(date, "QQQ", { locale: zhHK }), "QQQ", ref, { locale: zhHK }); //=> Mon Jul 01 2024 ✅
parse(format(date, "QQQ", { locale: zhTW }), "QQQ", ref, { locale: zhTW }); //=> Mon Jul 01 2024 ✅
```

## Likely origin

PR #2771 (merged 2021-11-30) changed the zh-CN quarter translation in the localize file from the 刻 forms to the 季 forms, but did not update the corresponding `matchQuarterPatterns` in the match file, which still expects the pre-#2771 forms. Related: #2889 / #2890 propose the same 刻 → 季 correction for zh-TW (that PR updates both localize and match, so it would not reproduce this problem in zh-TW), but neither mentions the existing zh-CN format/parse mismatch.

## Note on fix direction

I'm not proposing a specific fix — the resolution is a locale/maintainer decision. Options include updating `matchQuarterPatterns` to accept the 季 forms (possibly keeping 刻 for backward compatibility), or changing the formatter, depending on the intended zh-CN translation.

## Environment

- date-fns: reproduced against current `main`, and the published date-fns@4.4.0 on npm contains the same `matchQuarterPatterns`.
- Node.js: v24.12.0

harshit-d3v:
Fixes #4270.

## The reported bug

zh-CN's `localize` produces the 季 quarter forms, but its `matchQuarterPatterns` still expects the 刻 forms that `localize` stopped producing in #2771. A format → parse round-trip through zh-CN returns `Invalid Date`:

```js
format(new Date(2024, 6, 1), "QQQ", { locale: zhCN });        //=> "第三季"
parse("第三季", "QQQ", new Date(2024, 0, 1), { locale: zhCN }); //=> Invalid Date
```

zh-HK has identical `quarterValues`, so this aligns zh-CN's match patterns with zh-HK's:

```diff
-  abbreviated: /^第[一二三四]刻/i,
-  wide: /^第[一二三四]刻钟/i,
+  abbreviated: /^第[一二三四]季/i,
+  wide: /^第[一二三四]季度/i,
```

zh-TW is deliberately left alone — it uses 刻 on *both* sides, so it round-trips correctly today. Its wording is the separate concern in #2889 / #2890.

This does drop the ability to parse the pre-#2771 刻 strings, which zh-CN itself has not emitted since 2021. If you would rather keep them for backward compatibility, `/^第[一二三四](季|刻)/i` and `/^第[一二三四](季度|刻钟)/i` would do it — happy to switch.

## A second issue this surfaced

Regenerating the snapshots also changed zh-HK and zh-TW, which I had not touched. The cause:

`[tasks."test/locales"]` runs the generator with `--test`, but the script compares the mode against `test`:

```ts
const mode = process.argv[2] || "generate";  // "--test"
...
if (mode === "test") {                       // never true
```

The guard never matched, so the task took the **write** branch — it rewrote the snapshots and exited 0 every time rather than verifying them.

Because that check never ran, drift went unnoticed: #4194 fixed Chinese month parsing on **2026-05-21**, but the zh-CN, zh-HK and zh-TW snapshots were last generated **2026-05-19** and still record October as `Invalid Date`. Those rows are corrected here.

The fix strips a leading `--` so both spellings work:

```ts
const mode = (process.argv[2] || "generate").replace(/^--/, "");
```

Happy to split this into its own PR if you would rather review it separately. I kept it together because regenerating the snapshots for #4270 necessarily picks up the drift, so the two are entangled in the diff either way.

## Verification

- `test/locales` with the guard fixed and snapshots at `main` → **fails**, correctly naming `zh-HK` (it exited 0 before)
- after regenerating → **passes**
- `vitest run` → **37 failed / 3148 passed**, byte-identical to a clean `main` checkout on this machine. All 37 are `ReferenceError: Temporal is not defined`, from running Node 22.18.0 against the 22.19.0 in `.nvmrc`. No change attributable to this PR.
- `oxfmt --check` on both edited source files → clean

A note for anyone reproducing on Windows: with `core.autocrlf=true`, the snapshot comparison fails on every locale on line endings alone, which looks like total drift. Set `core.autocrlf=false` before running `test/locales`.

---

Disclosure: I used an AI assistant while investigating and preparing this change, including running the verification above.


harshit-d3v:
One clarification on the snapshot-guard half of this PR, since it makes the case a bit stronger than I put it originally.

`test/locales` is not a task that merely exists — it runs on every push and PR today, via `.github/workflows/code_quality.yaml`:

```yaml
on: [push, pull_request]
...
      - name: Locale snapshots
        run: mise //pkgs/core:test/locales
```

So the "Locale snapshots" step has been green on every PR while doing the opposite of what it claims: the mode argument is `--test`, the script compares it against `test`, the guard is false, and it takes the write branch. On a CI runner it rewrites the snapshots into the workspace, exits 0, and the job passes.

That is what let the drift sit for two months — #4194 landed on 2026-05-21 with the zh-CN, zh-HK and zh-TW snapshots last generated on 2026-05-19, and every CI run since then reported the snapshots as fine.

After the one-line change here, that step does what its name says: I ran it against `main` with the guard fixed and the snapshots untouched, and it fails, naming `zh-HK`. Regenerating makes it pass.

Worth noting the same class of bug can't recur silently now — but if you'd rather have the guard fail loudly on an unrecognised mode instead of falling through to `generate`, I'm happy to add that too.

harshit-d3v:
Flagging an overlap I should have caught before opening this: #4275 fixes the same zh-CN quarter bug and predates this by six days. It's the original and it declares `Fixes #4270`, so that's the one that should land.

Rather than leave a competing PR open, I'd like to rescope this to the half that doesn't overlap — the `localeSnapshots` runner:

```diff
-const mode = process.argv[2] || "generate";
+const mode = (process.argv[2] || "generate").replace(/^--/, "");
```

`pkgs/core/mise.toml` invokes the script with a flag, not a bare word:

```toml
[tasks."test/locales"]
run = "node ./scripts/build/localeSnapshots/index.ts --test"
```

So `mode` is `"--test"`, and the `mode === "test"` check never matches. Execution falls through to the `else`, which is `writeFile(snapshotPath, formattedSnapshot)` — the comparison that throws on a mismatch never runs.

That means the "Locale snapshots" step in `code_quality.yaml` currently rewrites the snapshots in the CI workspace instead of asserting them, so snapshot drift can't fail CI for any locale. It's also why the zh-CN mismatch in #4270 never surfaced as a failing snapshot in the first place.

That part is independent of the quarter fix and seems worth landing on its own. Happy to close this outright instead if you'd prefer.

PR #4286, opened 2026-08-08
Reject non-number amounts in addDays, addMonths, and addBusinessDays

### LINKED ISSUE #4284: `addDays` produces an unexpected date when `amount` is a string
## Description

While testing addDays, I noticed that passing a string as the amount parameter results in a seemingly valid but unexpected date instead of either rejecting the input or producing a more obvious result.

## Reproduction

```js
import { addDays } from 'date-fns';

const result1 = addDays(new Date(2026, 8, 10), 10);
/* 2026-09-19T18:30:00.000Z */

const result2 = addDays(new Date(2026, 8, 10), '10');
/* 2629-06-05T18:30:00.000Z */

const result3 = addDays(new Date(2026, 8, 10), 1000);
/* 2629-06-05T18:30:00.000Z */

console.log({ result1, result2, result3 });
```

## Expected Behavior

Since `amount` is documented as a `number`, I expected passing a string to either:

- be rejected (e.g. return `Invalid Date` or otherwise fail predictably), or
- behave consistently with the numeric value `10` if numeric strings are intentionally supported.

Instead, passing `"10"` results in a completely different valid date due to JavaScript string concatenation.

## Possible cause

It appears the implementation performs arithmetic similar to:

```js
date.setDate(date.getDate() + amount);
```

When amount is a string:

```js
10 + "10" // => "1010"
```

which effectively becomes:

```js
date.setDate(1010);
```

This results in a valid but unintended date.

## Question

Is this the intended behavior for unsupported input, or would it make sense for `addDays` to fail more predictably when `amount` is not a number?

> I suspect the same behavior may also exist in other `add*` utilities that accept an `amount` parameter, as they may follow a similar implementation pattern.

> If this is considered unintended behavior, I'd be happy to work on a PR to address it.

fudianchn:
AI disclosure: this change was prepared with AI coding agents, reviewed and revised line by line by me.

## 🧩 Context

`addDays`, `addMonths` and `addBusinessDays` guard an invalid `amount` with

```ts
if (isNaN(amount)) return constructFrom(..., NaN)
```

but `isNaN` coerces its argument, so a numeric string like `"10"` bypasses the guard (#4284). It then reaches e.g. `_date.setDate(_date.getDate() + amount)`, where `+` concatenates the string and the date jumps to a bogus value:

```ts
addDays(new Date(2026, 8, 10), 10)   // 2026-09-20  ✓
addDays(new Date(2026, 8, 10), "10") // 2029-06-05  ✗ (string concatenation in setDate)
```

The type signature is `amount: number`, so a string is a misuse — but the current behaviour fails silently with a plausible-looking result instead of failing predictably.

## 📦 Changes

Reject non-number amounts up front, so the call returns an Invalid Date (`NaN`) — matching the existing `NaN`-amount behaviour:

```diff
-  if (isNaN(amount)) return constructFrom(options?.in || date, NaN);
+  if (typeof amount !== "number" || isNaN(amount))
+    return constructFrom(options?.in || date, NaN);
```

Applied to:

- `addDays`, `addMonths`, `addBusinessDays`: the only `add*` helpers that can emit a plausible-but-wrong date from a non-number amount. Each keeps its own `constructFrom(...)` clause.
- `tpyAddDays` and `tpyAddBusinessDays`, the Temporal implementations that share the same guard, so both implementations agree.

The other `add*` helpers don't share this hazard: `addWeeks`, `addYears`, `addQuarters`, `addHours`, `addMinutes`, and `addSeconds` multiply `amount` by a factor, so a numeric string coerces to a number and they return the correct result; `addMilliseconds` computes `+toDate(date) + amount`, which for a string becomes `new Date("<timestamp>10")` and parses to an Invalid Date.

Number amounts — including `0` (no-op) and `NaN` — are unchanged.

## 🧪 Testing

Added a regression test for the `"10"` input to each of `addDays`, `addMonths`, and `addBusinessDays`. The `addDays` and `addBusinessDays` suites are shared with their Temporal implementations via `test.tp.ts`, so those tests cover both the regular and Temporal code paths.

```ts
addDays(date, "10") // Invalid Date  (was: 2029-…)
addDays(date, 10)   // +10 days      (unchanged)
addDays(date, NaN)  // Invalid Date  (unchanged)
addDays(date, 0)    // no-op         (unchanged)
```

Verified locally:

- `vitest run --project main src/addDays/test.ts src/addMonths/test.ts src/addBusinessDays/test.ts` → 38 passed, 15 skipped.
- `vitest run --project temporarily src/addDays/test.tp.ts src/addBusinessDays/test.tp.ts` → 26 passed, 8 skipped.
- `tsgo --build` → passes.
- `oxlint --type-aware` on the changed files → passes.

## 🛠 Alternative considered

Instead of rejecting, the string could be coerced (`amount = Number(amount)`) so `"10"` behaves like `10`. I went with rejecting because `amount` is typed `number`, the existing guard already returns Invalid Date for invalid numeric input, and a predictable failure is safer than an implicit coercion. Happy to switch if coercion is preferred.

Closes #4284

### DIFF
--- a/pkgs/core/src/addBusinessDays/index.tp.ts
+++ b/pkgs/core/src/addBusinessDays/index.tp.ts
@@ -15,7 +15,8 @@ export function tpyAddBusinessDays<
   options?: AddBusinessDaysOptions<ResultDate> | undefined,
 ): ResultDate {
   let [temporal, invalidDate] = toTpInstant(date, options);
-  if (!temporal || isNaN(amount)) return invalidDate;
+  if (!temporal || typeof amount !== "number" || isNaN(amount))
+    return invalidDate;
 
   const startedOnWeekend = tpIsWeekend(temporal);
   const hours = temporal.hour;
--- a/pkgs/core/src/addBusinessDays/index.ts
+++ b/pkgs/core/src/addBusinessDays/index.ts
@@ -51,7 +51,8 @@ export function addBusinessDays<
   const _date = toDate(date, options?.in);
   const startedOnWeekend = isWeekend(_date, options);
 
-  if (isNaN(amount)) return constructFrom(options?.in, NaN);
+  if (typeof amount !== "number" || isNaN(amount))
+    return constructFrom(options?.in, NaN);
 
   const hours = _date.getHours();
   const sign = amount < 0 ? -1 : 1;
--- a/pkgs/core/src/addBusinessDays/test.ts
+++ b/pkgs/core/src/addBusinessDays/test.ts
@@ -69,6 +69,12 @@ describe("addBusinessDays", () => {
     expect(result instanceof Date && isNaN(result.getTime())).toBe(true);
   });
 
+  it("returns `Invalid Date` if the given amount is not a number", () => {
+    // @ts-expect-error - We're testing an invalid amount
+    const result = addBusinessDays(new Date(2014, 8 /* Sep */, 1), "10");
+    expect(result instanceof Date && isNaN(result.getTime())).toBe(true);
+  });
+
   it("starting from a weekend day should land on a weekday when reducing a divisible by 5", () => {
     const subtractResult = addBusinessDays(new Date(2019, 7, 18), -5);
     expect(subtractResult).toEqual(new Date(2019, 7, 12));
--- a/pkgs/core/src/addBusinessDays/test.ts
+++ b/pkgs/core/src/addBusinessDays/test.ts
@@ -69,7 +69,7 @@ describe("addBusinessDays", () => {
     expect(result instanceof Date && isNaN(result.getTime())).toBe(true);
   });
 
-  it("returns `Invalid Date` if the given amount is not a number", () => {
+  it("returns `Invalid Date` if the given amount is a numeric string", () => {
     // @ts-expect-error - We're testing an invalid amount
     const result = addBusinessDays(new Date(2014, 8 /* Sep */, 1), "10");
     expect(result instanceof Date && isNaN(result.getTime())).toBe(true);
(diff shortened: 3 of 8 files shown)


pythonpioneer (changes requested):
I think we should also add tests covering this change.

fudianchn:
Added regression coverage for the `"10"` input to addDays, addMonths, and addBusinessDays. The addDays and addBusinessDays suites are shared with their Temporal implementations via test.tp.ts, which exposed the same coercion there, so tpyAddDays and tpyAddBusinessDays now reject non-number amounts as well.

Verified:
- `vitest run --project main src/addDays/test.ts src/addMonths/test.ts src/addBusinessDays/test.ts` -> 38 passed, 15 skipped.
- `vitest run --project temporarily src/addDays/test.tp.ts src/addBusinessDays/test.tp.ts` -> 26 passed, 8 skipped.
- `tsgo --build` -> passes.
- `oxlint --type-aware` on the changed files -> passes.

pythonpioneer (commented):
Could we update the test description to be more specific? There’s already a test for NaN that covers a different case. Since this test passes "10", I think `"returns `Invalid Date` if the given amount is a numeric string"` would better describe what’s actually being tested.

> BTW, nice work!

fudianchn:
Renamed the descriptions to `returns \`Invalid Date\` if the given amount is a numeric string` in all three suites (cffe43d).

The strict guard is limited to `addDays`, `addMonths`, and `addBusinessDays` because they're the only `add*` helpers that emit a plausible-but-wrong date from a non-number amount. The others either coerce a numeric string via multiplication (`amount * factor` -> number -> correct result) or already yield an Invalid Date (`addMilliseconds`). Extending the guard across the whole `add*` family would change functions that currently work, so I've kept it out of this PR. Happy to open a separate issue to discuss unifying it if that's useful.

pythonpioneer (approved):
Great Work!

pythonpioneer:
@kossnocorp, could you please take a look at this PR when you get a chance? @fudianchn  has resolved the issue, and the changes are ready for review. Thanks!

PR #845, opened 2018-08-19
Add Bengali locale (bn) for v2

touhidrahman:
Fixes #836 
This PR will introduce Bengali (bn) locale in date-fns.

PR #4287, opened 2026-08-10
fix(bn): use full month names for abbreviated month width

DrHaque:
## Summary

The Bengali (`bn`) locale's abbreviated month names (`MMM`/`LLL` format tokens) use truncated forms like `জানু`, `ফেব্রু`, `সেপ্ট`, `অক্টো`, `নভে`, `ডিসে` that don't reflect how Bengali is actually written. This PR changes `abbreviated` to match `wide` (full month names), consistent with how Bengali dates are conventionally written.

## Why

- Bengali doesn't have an established convention for truncating Gregorian month names. A Bangladesh government reference on writing dates in Bengali (https://www.teachers.gov.bd/blog/details/632252) gives the modern rule as day-number + **full** month name (e.g. ২৬ মার্চ, ৮ ডিসেম্বর) — no abbreviated forms appear.
- No Bangla Academy dictionary or publication defines abbreviated forms for these words.
- The current values were introduced unsourced in #845 (2018) — no CLDR/reference citation in that PR.
- This isn't unprecedented for CLDR-based i18n data: CLDR itself sets `abbreviated == wide` for other Indic locales that, like Bengali, transliterate the Gregorian month names rather than using native calendar words — e.g. Odia (`or`) and Nepali (`ne`) both have abbreviated identical to wide for all 12 months. Bengali should be treated consistently.
- Spelling note: the full forms used here (জানুয়ারি, ফেব্রুয়ারি, …) follow Bangla Academy's প্রমিত বানান (standardized spelling) rule that non-tatsama words — তদ্ভব, দেশী, বিদেশী (foreign), and মিশ্র (mixed) — take হ্রস্ব ই-কার (ি) rather than দীর্ঘ ঈ-কার (ী). Since these are বিদেশী (foreign) loanwords from English, the same rule that governs ফারসি, ফরাসি, জাপানি, ইংরেজি, etc. applies here, giving জানুয়ারি rather than the older/informal জানুয়ারী seen in some bilingual dictionaries. This matches the locale's existing (unchanged) `wide` values, so no spelling change is introduced — just confirming the values this PR extends to `abbreviated` are already spelled correctly.

Scope: this only changes `abbreviated` width for **months**. `narrow` width is left as-is (single/double-letter forms serve a different, space-constrained UI purpose, not a claim about everyday usage).

cc @touhidrahman, who authored the original `bn` locale in #845 — I reached out directly and he confirmed the short forms were his own shorthand chosen at the time date-fns's width system looked different, not values drawn from an established convention, and he's supportive of updating them now.

## Changes

- `src/locale/bn/_lib/localize/index.ts`: `monthValues.abbreviated` now matches `monthValues.wide`.
- `src/locale/bn/_lib/match/index.ts`: `matchMonthPatterns.abbreviated` updated to match the full-length forms (needed so `parse()` consumes the whole matched word — previously it only stripped the short prefix, which is fine for the old short values, but round-trip correctness requires the match pattern and the localize array to be in sync).
- `src/locale/bn/snapshot.md`: regenerated via `pnpm run locale-snapshots`.

## Test plan

- [x] Regenerated locale snapshot (`pnpm run locale-snapshots`) — diff limited to `bn` (MMM/LLL/PP/PPpp rows now show full names, and parse round-trips correctly for all 12 months, including January/February which previously used the special-cased short forms জানু/ফেব্রু)
- [x] `pnpm vitest run` — 3166 passed, 0 new failures (46 pre-existing failures are unrelated `Temporal is not defined` errors from experimental `.tp.ts` tests, reproducible on an unmodified checkout)
- [x] `oxfmt` clean


touhidrahman (approved):

DrHaque:
For reviewers weighing whether this is just a style preference: the same underlying problem was independently found and already fixed in two other locale-data sources, which corroborate the reasoning in the PR description.

**glibc** — `bn_BD`/`bn_IN`'s `abmon` (POSIX `%b`) had the identical inconsistency: only জানু/ফেব were ever actually truncated, March–December were already spelled out in full. Reviewed and merged upstream by the glibc i18n maintainer:
https://sourceware.org/git/?p=glibc.git;a=commit;h=ea99fc77eaac0c1abb00c6b17d5dd7e7611406dd

**CLDR** — filed as [CLDR-19739](https://unicode-org.atlassian.net/browse/CLDR-19739), requesting the same `abbreviated == wide` fix for `bn`'s Gregorian month data (pending Survey Tool review; not yet released, but documents the evidence in detail — জানু collides with an unrelated, existing Bengali word meaning "knee," and CLDR's own `abbreviated`/`stand-alone.abbreviated` data is internally inconsistent, already leaving several months unabbreviated).

Both independently trace back to the same root cause this PR's description already identifies for date-fns: an unreviewed source with no real citation behind the short forms. On the CLDR side specifically, that's [CLDR-522](https://unicode-org.atlassian.net/browse/CLDR-522), a 2005 community submission that cited two general-purpose Bangla Academy dictionaries — neither of which actually prescribes month abbreviations — and never completed Survey Tool review before being auto-closed by a 2019 migration bot.

So this isn't three projects independently guessing the same fix is nicer — it's three projects converging on the same conclusion from the same underlying evidence: Bengali has no established convention for abbreviating Gregorian month names, and the specific short forms in circulation aren't real abbreviations Bengali speakers use or recognize.

Issue #3291, opened 2022-12-04
`formatISO()` basic incorrect


20kano:
```js
import { formatISO } from "npm:date-fns"

const d = formatISO(new Date, { format: 'basic' })
console.log(d);
```
```
20221204T131201+09:00
```
expected
```
20221204T131201+0900
```
deno 1.28.3 (release, x86_64-unknown-linux-gnu)
v8 10.9.194.5
typescript 4.8.3

mandaputtra:
Could you make a PR for this issue?

AhmadDecoded:
I am on it @mandaputtra

AhmadDecoded:
@mandaputtra solved it and updated one test too.

PR #4311, opened 2026-08-30
Fix formatISO basic format keeping colon in timezone offset

afonsojanu:
Fixes `formatISO`'s basic format so the timezone offset actually follows ISO 8601 basic notation (no separators), matching how the date and time parts are already handled.

Repro:

```js
formatISO(new Date(), { format: 'basic' })
// before: 20260830T234218+01:00  (colon still there)
// after:  20260830T234218+0100
```

Reported as #3291. There's an older PR for the same issue (#3308) but it's been open for a few years, has merge conflicts now, and its fix only touched the source without noticing the test helper has the identical bug baked into it (more on that below), so I put together a fresh one instead of trying to rebase it.

Cause: the timezone-offset branch in `formatISO` always joined hours and minutes with a literal colon, ignoring the `timeDelimiter` the rest of the function already derives from the `format` option. Extended format happened to want a colon anyway, so only the basic format case was ever visibly wrong.

The reason the existing tests didn't catch this: the shared `generateOffset` test helper in `_lib/test/index.ts` builds its expected offset with that same hardcoded colon, so the basic format tests were comparing the (wrong) actual output against an (equally wrong) expected value and always passed. I gave `generateOffset` an optional delimiter argument, defaulting to `:` so `formatRFC3339`'s tests (which always want the colon form) are unaffected, and pass an empty string from the basic format assertions in `formatISO`'s test file.

Verified the regression test actually catches the bug: reverting just the source fix while keeping the updated test makes `formats ISO-8601 basic format` fail with exactly the `+01:00` vs `+0100` mismatch; restoring the fix makes it and the rest of the suite pass again.

Ran `pnpm vitest run` in `pkgs/core`. The only failures are pre-existing and unrelated to this change (locale-dependent `intlFormatDistance` assertions that need an `en-US` ICU locale, and the `*.tp.ts` temporal test files) and reproduce identically on a clean checkout with none of these changes applied.

### DIFF
--- a/pkgs/core/src/_lib/test/index.ts
+++ b/pkgs/core/src/_lib/test/index.ts
@@ -18,7 +18,7 @@ export function resetDefaultOptions(): void {
 }
 
 // This makes sure we create the consistent offsets across timezones, no matter where these tests are ran.
-export function generateOffset(originalDate: Date) {
+export function generateOffset(originalDate: Date, delimiter: string = ":") {
   // Add the timezone.
   let offset = "";
   const tzOffset = originalDate.getTimezoneOffset();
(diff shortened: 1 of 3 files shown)


lbesecker195 (approved):
Verified: `formatISO` already computes `timeDelimiter` as `""` for basic format and uses it for the time part, so the offset was the one component still hardcoding `:`. ISO 8601 basic notation omits the separator in the offset too. LGTM.

PR #4312, opened 2026-09-01
fix(en-US): support narrow noon "n" in day period parsing

lx3133584:
## Problem
Formatting noon with the narrow token `bbbbb` produces `"n"`, but parsing `"n"` back fails to match any day period in the `en-US` locale parser. The parsed date silently falls through to hour 0 (midnight) instead of 12:00.

## Root Cause
In `pkgs/core/src/locale/en-US/_lib/match/index.ts`, `matchDayPeriodPatterns.narrow` matches `"n"`, but `parseDayPeriodPatterns.any.noon` only matches `/^no/i`, which fails to match the single-character string `"n"`.

## Fix
Update `parseDayPeriodPatterns.any.noon` to `/^(no|n$)/i` to support the narrow `"n"` day period token in addition to `"noon"`.

## Testing
Added a unit test for narrow noon parsing with `bbbbb` token in `pkgs/core/src/parse/test.ts` and verified with vitest.

### DIFF
--- a/pkgs/core/src/locale/en-US/_lib/match/index.ts
+++ b/pkgs/core/src/locale/en-US/_lib/match/index.ts
@@ -80,7 +80,7 @@ const parseDayPeriodPatterns = {
     am: /^a/i,
     pm: /^p/i,
     midnight: /^mi/i,
-    noon: /^no/i,
+    noon: /^(no|n$)/i,
     morning: /morning/i,
     afternoon: /afternoon/i,
     evening: /evening/i,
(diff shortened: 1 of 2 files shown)


lbesecker195 (approved):
Checked against date-fns 4 from npm with an `enUS` copy whose `dayPeriod` matcher uses this pattern. `format(noon, 'bbbbb')` gives `n`, and today `parse('n', 'bbbbb', ref)` lands on hour 0 because the narrow match pattern accepts `n` but none of the parse patterns do (`noon: /^no/i` needs two characters). With `/^(no|n$)/i` it parses to 12, so narrow noon round-trips like `mi` does. `mi`, `a`, `p`, `noon`, `midnight` and `a.m.` parse to the same hours as before, and since the parse pattern only sees the matched string, `n$` can't catch anything longer. LGTM.

PR #4314, opened 2026-09-02
Fix tzOffset returning the wrong sign for sub-hour western offsets

contactjawad:
I hit this while formatting some historical dates. `tzOffset` returns a positive value for zones that were on a negative offset smaller than an hour.

Monrovia was at UTC-00:44:30 until 1972:

```js
tzOffset('Africa/Monrovia', new Date('1970-01-01T00:00:00Z'))
// -> 44.5, should be -44.5
```

The offset string from `Intl` is `GMT-00:44:30`. In `calcOffset` the sign comes from `hours * 60 + minutes > 0`, but for `-00` the hours part is `-0`, so the only thing carrying the minus is the string, and it gets dropped — the positive branch wins.

I read the sign from the offset string instead of the magnitude, so `-00:44:30` stays negative. The other offsets (`+05:30`, `-08:00`, `-03:30`, …) are unchanged. Added a test for Monrovia.

### DIFF
--- a/pkgs/tz/src/tzOffset/index.ts
+++ b/pkgs/tz/src/tzOffset/index.ts
@@ -54,8 +54,7 @@ function calcOffset(cacheStr: string, values: string[]): number {
   const minutes = +(values[1] || 0);
   // Convert seconds to minutes by dividing by 60 to keep the function return in minutes.
   const seconds = +(values[2] || 0) / 60;
+  const sign = values[0]?.trim().startsWith("-") ? -1 : 1;
   return (offsetCache[cacheStr] =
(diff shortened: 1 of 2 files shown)


lbesecker195 (approved):
Verified: in both the Intl path (`"GMT-00:44:30".split(":")`) and the regex fallback (`([+-]\d\d)`), `values[0]` carries the sign, so `+"-00"` collapses to `-0` and the old `hours * 60 + minutes > 0` test picked the wrong branch for any offset between `-00:01` and `-00:59`. Reading the sign from the string is the right fix. LGTM.

PR #4320, opened 2026-09-12
fix(vi): parse Vietnamese quarters II and III as Q2 and Q3

NgoQuocViet2001:
## Problem

The Vietnamese locale is the only one that parses quarters as Roman numerals, and its patterns anchor only at the end of the matched string:

```ts
const parseQuarterPatterns = {
  any: [/(1|i)$/i, /(2|ii)$/i, /(3|iii)$/i, /(4|iv)$/i] as const,
};
```

`buildMatchFn` takes the **first** pattern that matches, and `"quý II"` does end with an `I`, so `/(1|i)$/` matches it before `/(2|ii)$/` is ever tried. The same happens for `III`.

Running the real matcher on every width:

```
input      width         parsed  expected
quý II     wide          1       2   <-- WRONG
quý III    wide          1       3   <-- WRONG
qII        abbreviated   1       2   <-- WRONG
qIII       abbreviated   1       3   <-- WRONG
II         narrow        1       2   <-- WRONG
III        narrow        1       3   <-- WRONG
```

`quý I`, `quý IV` and all the digit forms (`quý 2`, `q2`, `2`) are correct — `IV` ends in `V`, and a digit cannot satisfy an earlier pattern. So 6 of the 18 quarter spellings resolve to the wrong quarter.

The effect is a silent wrong answer: `parse` returns a valid date three or six months earlier than the string names, with no throw and no `Invalid Date`. Anything deriving a period from a Vietnamese quarter label reads Q2 and Q3 data as Q1.

## The snapshot already recorded it

`pkgs/core/src/locale/vi/snapshot.md` formats `2019-04-01` as `quý II` and parses it back to **`2019-01-01`**:

```
| | QQQQ | 2019-04-01T12:13:14.015Z | quý II | 2019-01-01T00:00:00.000Z |
```

The sibling rows on the same date — `QQQ` → `Q2`, `QQQQQ` → `2`, `qqqq` → `Quý 2` — all round-trip to `2019-04-01`, which isolates the defect to the Roman-numeral spelling.

## Fix

Anchor each numeral on a character that cannot be part of a Roman numeral, so a longer numeral can no longer satisfy a shorter pattern:

```ts
any: [
  /(?:^|[^iv])(1|i)$/i,
  /(?:^|[^iv])(2|ii)$/i,
  /(?:^|[^iv])(3|iii)$/i,
  /(?:^|[^iv])(4|iv)$/i,
] as const,
```

The `(?:^|[^iv])` alternative keeps the abbreviated and narrow widths working, where the numeral is preceded by `q` or by nothing at all.

## Testing

- Ran `vi.match.quarter` over all 18 spellings across the three widths: **18/18 correct** with the change, **6 wrong** without it (the table above).
- Recomputed every `format`/`parse` cell of the vi snapshot with the same calls `renderFormatParse` makes, before and after: **exactly one cell moves**, the `QQQQ` / `quý II` row, from `2019-01-01T00:00:00.000Z` to `2019-04-01T00:00:00.000Z`. That line is updated here.
- Not run: `mise //pkgs/core:gen/locales/snapshots` — `pnpm install` cannot complete on this machine because the JSR registry returns 403 for `@jsr/std__fmt`. The snapshot edit above is the computed result rather than generator output, so it is worth re-running before merge; per the diff above it should be a no-op.


lbesecker195 (approved):
`buildMatchFn` uses `findIndex` over `parsePatterns`, so order matters. Running the old and new `parseQuarterPatterns` against every string the vi `matchQuarterPatterns` can produce (`quý I..IV`, `qI..qIV`, `I..IV`, plus the digit forms for all three widths): old resolves II and III to 1 in every width, new gives 1/2/3/4 for all 24 inputs, and the digit forms and I/IV are unchanged. The snapshot row for `quý II` now round-trips to 2019-04-01. LGTM.

